PR #2, opened 2015-09-09
Use a package.json to fix versions of all the node modules we use

gagern:
Now the individual “npm install ‹package›” calls are replaced by a single
“npm update --dev” which uses the descriptions from package.json to not only
download all requested packages at once, but also at an appropriate version.


strobelm:
I don't feel qualified to judge all the implications of this pull request, Martin if you're sure it works feel free to merge it.

PR #4, opened 2015-09-11
Linearsolve warning

strobelm:
* Added a warning for singular matrices in LUDecomp
* Added LinearSolveQR (not working until a bug in RRQR is found)

### DIFF
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -2153,6 +2153,8 @@ List.LUdecomp = function(AA) {
                 Pk = j;
             }
         }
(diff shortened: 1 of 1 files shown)


gagern:
If you expect the code in `linearsolveQR` to work any day soon, I suggest you comment it out using `/* … */`. That way, enabling the code only modifies two lines, not all of them. Otherwise, it might be better to not add this code at all, and restrict this merge request to the warning.

strobelm:
Well the code of linearsolveQR works fine with the non-pivoted QR decomposition but not with it. I'm planned to accelerate the code anyway and within this process to fix the bug.

PR #5, opened 2015-09-15
isCircle property

strobelm:
Hi, i added code to GeoOps which indicated whether a conic is a circle or not. 

isCircle is written directly to the object properties. Should we rather do something like el.matrix = General.withUsage(el.matrix, "Circle")? The proposed changes at least mean no harm and won't break functions which expect the usage to be Conic.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -643,8 +643,8 @@ geoOps.ConicBy5.updatePosition = function(el) {
 
 
 geoOps._helper.isCircle = function(mat) {
-    var nI = CSNumber.abs2(General.mult(List.ii,(General.mult(mat,List.ii)))).value.real;
-    var nJ = CSNumber.abs2(General.mult(List.jj,(General.mult(mat,List.jj)))).value.real;
+    var nI = CSNumber.abs2(General.mult(List.ii, (General.mult(mat, List.ii)))).value.real;
+    var nJ = CSNumber.abs2(General.mult(List.jj, (General.mult(mat, List.jj)))).value.real;
 
     return ((nI < CSNumber.eps) && (nJ < CSNumber.eps));
 };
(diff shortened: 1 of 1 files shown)


gagern:
I'd say `withUsage(matrix, "Circle")` should be reserved for situations where due to the nature of the operation, the result will *always* be a circle. As done for `CircleBy3`.

What is the application for this numerically decided `isCircle` property? Perhaps that application should be included in the pull request to motivate the added functionality.

kortenkamp:
If a conic has been defined and was proven to be a (euclidean) circle, it should be usable as such. That's the reason for the ```withUsage``` construction.

gagern:
I agree that if we can *prove* such a property, we may use it for the usage. But there is no prover here yet, just some numeric evidence for the current parametrization. It might be coincidence, and I'd rather avoid the usage switching between conic and circle while the configuration gets manipulated.

strobelm:
Well if i remember correctly this feature was a wish by the BDFL. Despite exercises like in Cinderella one could use this for numerics since the intersection of e.g. circle/circle is more efficient than conic/conic.

Cinderella by the way seems to rely only on usage since circles defined by conicby5 are not detected as such (at least not by iscircle()).

gagern:
At least at the moment, the circle-circle intersection is a completely different operation than the conic-conic intersection. The former has two results while the latter has four, to name the most prominent distinction. Switching between these operations would thoroughly confuse tracing. So I don't think we can substitute algorithms before we have a randomized prover.

kortenkamp:
We could have  circle-circle+I+J-operation that uses circle-circle intersection and just adds the two imaginary intersections I and J. This wouldn't need a prover.

gagern:
I'd say we can add this code if we have some application. So if you indeed to circle-circle+I,J style intersections if both arguments are circles, then make that a separate commit in this same pull request. Otherwise I'm against merging what would essentially be dead code for now. We can of course still keep the branch around for when an application turns up.

strobelm:
I'll close the request since the topic needs more discussion.

PR #7, opened 2015-09-15
Adapt random number tests to reduced number of decimal digits

gagern:
Since f3b32d9577e085f6907efdf9b11680c73df79d42, numbers will be printed with only 4 decimal places.  But since trailing zeros are omitted, we might in fact get fewer than four decimals.  Requiring only two non-zero decimals will reduce the chance of a random test failure to 1% per test.

### DIFF
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -301,19 +301,19 @@ The following operators generate pseudo random numbers.
 
     > random()
     < 0.468
-    ~ 0\.\d+
+    ~ 0(\.\d+)?
     > 1000*random()
     < 278.4083
-    ~ \d+\.\d+
+    ~ \d+(\.\d+)?
 
 #### (0,1)-normally distributed random number: `randomnormal()`
 
     > randomnormal()
     < 0.1325
-    ~ -?\d+\.\d+
+    ~ -?\d+(\.\d+)?
     > 1000*randomnormal()
     < -1199.8909
-    ~ -?\d+\.\d+
+    ~ -?\d+(\.\d+)?
 
 #### Random boolean value `true` or `false`: `randombool()`
 
@@ -325,13 +325,13 @@ The following operators generate pseudo random numbers.
 
     > random(10)
     < 7.089
-    ~ \d\.\d+
+    ~ \d(\.\d+)?
     > random(1000)
     < 882.4784
-    ~ \d+\.\d+
+    ~ \d+(\.\d+)?
     > sum(1..1000, random(5))
     < 2464.8004
-    ~ 2\d\d\d\.\d+
+    ~ 2\d\d\d(\.\d+)?
 
 #### Uniformly distributed random integer number between 0 and `‹number›`: `randomint(‹number›)`
 
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and 1: `random()`
 
     > random()
-    < 0.4680764123124367
-    ~ 0\.\d{4,}
+    < 0.468
+    ~ 0\.\d{2,}
+    > 1000*random()
+    < 278.4083
+    ~ \d{2,3}\.\d{2,}
 
 #### (0,1)-normally distributed random number: `randomnormal()`
 
     > randomnormal()
-    < 0.1325114717517828
-    ~ -?\d+\.\d{4,}
+    < 0.1325
+    ~ -?\d+\.\d{2,}
+    > 1000*randomnormal()
+    < -1199.8909
+    ~ -?\d+\.\d{2,}
 
 #### Random boolean value `true` or `false`: `randombool()`
 
@@ -318,10 +324,13 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and `‹number›`: `random(‹number›)`
 
     > random(10)
-    < 7.089078226464412
-    ~ \d\.\d{3,}
+    < 7.089
+    ~ \d\.\d{2,}
+    > random(1000)
+    < 882.4784
+    ~ \d{2,3}\.\d{2,}
     > sum(1..1000, random(5))
-    < 2464.8003929607607
+    < 2464.8004
     ~ 2\d\d\d\.\d+
 
 #### Uniformly distributed random integer number between 0 and `‹number›`: `randomint(‹number›)`
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -301,19 +301,19 @@ The following operators generate pseudo random numbers.
 
     > random()
     < 0.468
-    ~ 0\.\d{2,}
+    ~ 0\.\d+
     > 1000*random()
     < 278.4083
-    ~ \d{2,3}\.\d{2,}
+    ~ \d+\.\d+
 
 #### (0,1)-normally distributed random number: `randomnormal()`
 
     > randomnormal()
     < 0.1325
-    ~ -?\d+\.\d{2,}
+    ~ -?\d+\.\d+
     > 1000*randomnormal()
     < -1199.8909
-    ~ -?\d+\.\d{2,}
+    ~ -?\d+\.\d+
 
 #### Random boolean value `true` or `false`: `randombool()`
 
@@ -325,10 +325,10 @@ The following operators generate pseudo random numbers.
 
     > random(10)
     < 7.089
-    ~ \d\.\d{2,}
+    ~ \d\.\d+
     > random(1000)
     < 882.4784
-    ~ \d{2,3}\.\d{2,}
+    ~ \d+\.\d+
     > sum(1..1000, random(5))
     < 2464.8004
     ~ 2\d\d\d\.\d+


kranich:
The tests fail, if second and third digits after the decimal point are 0 and the fourth digit is 4 or less. Hence, the probability of failure is 1/10 * 1/10 * 4/10 = 1/250, even smaller than 1%.

However, why should the tests fail at all? We *can* obtain a number from `random()` whose `niceprint()` has fewer than 2 places after the decimal point and we *can* obtain a number whose `niceprint()` is a natural number from `1000 * random()`, etc.

@gagern: I suggest to accept 1 place after the decimal point for `random()` (since 0 and 1 have probability 0), and to accept any number (without decimal point or at least 1 decimal place) for `1000 * random()`.

ref/Arithmetic_Operators.md
```diff
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and 1: `random()`
 
     > random()
-    < 0.4680764123124367
-    ~ 0\.\d{4,}
+    < 0.468
+    ~ 0\.\d{2,}
+    > 1000*random()
+    < 278.4083
+    ~ \d{2,3}\.\d{2,}
```
kranich: `1000*random()` may have only 1 digit before the decimal point.

ref/Arithmetic_Operators.md
```diff
@@ -318,10 +324,13 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and `‹number›`: `random(‹number›)`
 
     > random(10)
-    < 7.089078226464412
-    ~ \d\.\d{3,}
+    < 7.089
+    ~ \d\.\d{2,}
+    > random(1000)
+    < 882.4784
+    ~ \d{2,3}\.\d{2,}
```
kranich: `random(1000)` may have only 1 digit before the decimal point.

gagern:
I know that all these tests may still fail. After all, `0` is a valid (though unlikely) result of the random number generator. I guess my original aim was ensuring that we have numbers which “look. sufficiently random”. Which means not too many decimals, i.e. not `floor(random()*100)/100` or something like that as the internal implementation. I realize that this is hard to test in single examples.

We could generate a list of several random numbers, and check that at least one of them has all the random digits we expect from them. But that is only a quantitative change, because at least theoretically there is still some chance that all generated numbers are special. The larger the number of tests, the lower that probability, so we can certainly make this work in practice.

Otherwise, simply accepting *any* number is certainly the solution which is least likely to fail. It doesn't guarantee any kind of randomness, but the current regexps don't do that either. I guess if we want that, we should do some statictical analysis, investigate mean and variance and so on, and boil our assertions down to some boolean checks without actually printing the random numbers.

kranich:
>Otherwise, simply accepting *any* number is certainly the solution 
>which is least likely to fail. It doesn't guarantee any kind of 
>randomness, but the current regexps don't do that either. I guess if we 
>want that, we should do some statictical analysis, investigate mean and 
>variance and so on, and boil our assertions down to some boolean checks 
>without actually printing the random numbers.

It would be nice if you changed the regexps so that one leading decimal 
is also accepted. Then I think I would merge the request, since it 
certainly improves the current situation.

Statistical analysis might be an option in the future, if it does not 
increase the running time of the tests too much.

kranich:
Or you may want to change the regexps so that any number is accepted. That would also improve the current situation and might be the cleanest approach for the time being.

gagern:
I'm now accepting arbitrary numbers, as long as they contain at least one digit after the decimal point.

ref/Arithmetic_Operators.md
```diff
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and 1: `random()`
 
     > random()
-    < 0.4680764123124367
-    ~ 0\.\d{4,}
+    < 0.468
+    ~ 0\.\d+
+    > 1000*random()
+    < 278.4083
+    ~ \d+\.\d+
```
kranich: This should be `\d{1,3}(\.\d+)?`.

ref/Arithmetic_Operators.md
```diff
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
...
 
 #### (0,1)-normally distributed random number: `randomnormal()`
 
     > randomnormal()
-    < 0.1325114717517828
-    ~ -?\d+\.\d{4,}
+    < 0.1325
+    ~ -?\d+\.\d+
```
kranich: This should be `-?\d\.\d+`.

ref/Arithmetic_Operators.md
```diff
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
...
+    > 1000*randomnormal()
+    < -1199.8909
```
kranich: How can this be a number of absolute value > 1000?

ref/Arithmetic_Operators.md
```diff
@@ -300,14 +300,20 @@ The following operators generate pseudo random numbers.
...
+    ~ -?\d+\.\d+
```
kranich: This should be `-?\d{1,3}(\.\d+)?`.

ref/Arithmetic_Operators.md
```diff
@@ -318,10 +324,13 @@ The following operators generate pseudo random numbers.
 #### Uniformly distributed random real number between 0 and `‹number›`: `random(‹number›)`
 
     > random(10)
-    < 7.089078226464412
-    ~ \d\.\d{3,}
+    < 7.089
+    ~ \d\.\d+
```
kranich: This should be `\d(\.\d+)?`.

ref/Arithmetic_Operators.md
```diff
@@ -318,10 +324,13 @@ The following operators generate pseudo random numbers.
...
+    > random(1000)
+    < 882.4784
+    ~ \d+\.\d+
```
kranich: This should be `\d{1,3}(\.\d+)?`.

ref/Arithmetic_Operators.md
gagern: No. Although the standard normal distribution is *concentrated* around zero, there is in fact no bound on its values, large magnitudes are just pretty unlikely. Unless I made a mistake in [my computation](http://www.wolframalpha.com/input/?i=erfc%281%2Fsqrt%282%29%29%2F2), you have about 16% chance for a value less than -1 and again 16% for greater 1.

ref/Arithmetic_Operators.md
```diff
@@ -318,10 +324,13 @@ The following operators generate pseudo random numbers.
...
     > sum(1..1000, random(5))
-    < 2464.8003929607607
+    < 2464.8004
     ~ 2\d\d\d\.\d+
```
kranich: I do not see why the result should start with digit 2. I believe `sum(1..1000, random(5))` should be equivalent to `random (5000)`.

In this case, the regexp should be `\d{1,4}(\.\d+)?`. Otherwise, please explain.

ref/Arithmetic_Operators.md
gagern: Only if you care to cater for the case where the random number is exactly equal to zero, at least after rounding. If we were to do that here, we should do so for all those number patterns, since this situation is even more likely for the nullary `random()` call.

kranich:
@gagern: Sorry to bother you again once more. I have commented what I think the regexps should be. If you prefer, you may simply ignore my comments and merge. In any case, I appreciate your opinion on my comments.

ref/Arithmetic_Operators.md
kranich: OK, my mistake.

ref/Arithmetic_Operators.md
kranich: Same mistake in my reasoning as above.

ref/Arithmetic_Operators.md
gagern: This is already a weak statistical analysis. While `random(5000)` would be uniformly distributed, the sum of uniformly distributed variables is not uniform any more, but according to [this post](http://math.stackexchange.com/questions/370228/probability-distribution-of-a-sum-of-uniform-random-variables) distributed according to the [Irwin-Hall distributuon](https://en.wikipedia.org/wiki/Irwin%E2%80%93Hall_distribution) which is normal in the limit (and 1000 is pretty close to that limit, I dare say). So essentially we expect that value to be around 2500, and checking the first digit supports that claim while still leaving quite a lot of room for random deviations from the expected outcome.

ref/Arithmetic_Operators.md
kranich: No, I would like to take care of the case that `niceprint` of `random(10)` is one of 1, 2, 3, 4, 5, 6, 7, 8, 9.

gagern: Ah, I see. OK, I guess that evens out the probabilities: chances that all four output digits will get rounded to zero are always 1e-4, no matter what we have before the decimal point, as long as we have an integer-delimited uniform range. So I still maintain that we should either have it in all cases or none, for the sake of consistency. Probably cleanest to always include it in all cases.

ref/Arithmetic_Operators.md
kranich: OK. I still prefer if test cases do not produce failures that are false negatives.

By the way, does travis CI retry a build if such a failure occurs? Or might this cause problems?

gagern:
See comments in the outdated diff for my opinion. The one thing where you convinced me I've included in my latest commit.

kranich:
I still wonder what happens if a travis CI build fails due to a rare random condition. How much hassle is it to retry a build?

gagern:
I can trigger a rebuild with a single click. Not sure if others could do the same after registering with Travis.

kortenkamp:
One could increase the rarity of that condition by just doing it twice and asking for it to fail at most once. If it's a 1:1.000.000 chance that it fails it will be a 1:1.000.000.000.000 chance for that modified test. Repeat if necessary.

ref/Arithmetic_Operators.md
gagern: According to [my computations](http://www.wolframalpha.com/input/?i=1%2Berf%28%282000-2500%29%2Fsqrt%282*5*1000%2F12%29%29) we're talking of a probability of 1.7<sup>−132</sup> here. I'd say we can safely ignore that problem, since there are unavoidable technical reasons why Travis might fail which are far more likely.

PR #8, opened 2015-09-16
is operations

strobelm:
Hi,

i added the following operations to cindyscript:

iscircle(1)
isconic(1)
isline(1)
ispoint(1)
isgeometric(1)
issun(1)
ismass(1)
isspring(1)
isbouncer(1)
isundefined(1)

the following were already implemented:
isinteger(1)
isreal(1)
iscomplex(1)
iseven(1)
isodd(1)
islist(1)
ismatrix(1)
isnumbervector(1)
isnumbermatrix (1)
isstring(1)

Cinderella has the following according to http://doc.cinderella.de/tiki-index.php?page=Operator+Index

isinteger (1) test for integer
isreal (1) test for real number
iscomplex (1) test for complex number
iseven (1) test for even integer
isodd (1) test for odd integer
islist (1) test for list
ismatrix (1) test for nested list in matrix shape
isnumbervector (1) test for list which is a number vector
isnumbermatrix (1) test for list which is a number matrix
isstring (1) test for string
isgeometric (1) test for geometric object
isselected (1) test for being selected
ispoint (1) test for point
isline (1) test for line
iscircle (1) test for circle
isconic (1) test for conic
ismass (1) test for mass
issun (1) test for sun
isspring (1) test for spring
isundefined (1) test for being undefined

Therefor isspring and isbouncer are not backwards compatible.

### DIFF
--- a/examples/104_isOperations.html
+++ b/examples/104_isOperations.html
@@ -1,0 +1,120 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Konstruktion.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+
+<script id="csdraw" type="text/x-cindyscript">
+println("ispoint: point, line");
+println(ispoint(A));
+println(ispoint(a));
+
+println("isline: point, line");
+println(isline(A));
+println(isline(a));
+
+
+println("iscircle: circle, conic");
+println(iscircle(C0));
+println(iscircle(C1));
+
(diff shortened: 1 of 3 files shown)


gagern:
I'd like to see either an example file for these, or even better a ref test. For the latter, we'd need to modify the test runner, so that it [accepts lines](https://github.com/CindyJS/CindyJS/blob/a962adf700262ca5e681d3a1b7b23c4d17b8d148/ref/js/runtests.js#L58-L99) of geometry description (probably starting with `G `) and [creates a new `cjs` instance](https://github.com/CindyJS/CindyJS/blob/a962adf700262ca5e681d3a1b7b23c4d17b8d148/ref/js/runtests.js#L37-L41) with that geometry. Even if you only use an example and no automated test for now, you should add the new operators to [the reference manual](https://github.com/CindyJS/CindyJS/blob/master/ref/Boolean_Operators.md).

gagern:
I'd also ask you to squash the commits into one, using `rebase -i master`.

strobelm:
I added an example file and squashed the commits.

gagern:
Will have a look at this tomorrow.

I assume your example is exported from Cinderella. If so, you might want to use a more recent version, or run `tools/reformatCreateCindy.js` on the exported file. Either should condense the gslp to one line per geometric object.

PR #13, opened 2015-10-08
Drop unknown operations and their dependencies

gagern:
Sometimes Cinderella exports things we cannot handle. In those cases it is probably better to continue with the things we can display, instead of failing completely with mysterious errors printed to the console only.

This commit will remove geometric elements with unknown operations from the gslp. It will also drop things which depend on these. There is an example to verify that this works as advertised.


kortenkamp:
thanks a lot, very useful feature!!

PR #19, opened 2015-10-09
Replace multiple calls to global functions in examples

### LINKED ISSUE #18: csplay() is broken
In the current master the examples (like examples/40_Swarm.html) which use csplay are broken. I bisected the commits and found the faulty commit to be 6a94d0714b62f2783d53b08fcd4cf312a1f81a41.

gagern:
The tool was missing a global flag for one RegExp, causing only the first occurrence to be replaced.  This should be better now. Fixes #18.

### DIFF
--- a/examples/31_EventTest.html
+++ b/examples/31_EventTest.html
@@ -95,8 +95,8 @@
         
 
         <button onclick="cdy.csplay()" type="button" style="position:absolute; top:650px;left:60px" >Play</button>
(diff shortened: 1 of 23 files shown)


gagern:
No, not fixed yet. Probably a scoping issue.

gagern:
Better now. At least `40_Swarm.html` works. It could theoretically be that in some example the `cdy` variable is not defined at the global scope, but instead inside some function or other. I somehow doubt that, though, and didn't check all affected examples manually.

PR #21, opened 2015-10-13
Beispiel Pendel mit mkdv Gleichung

richter-gebert:
Ich lerne gerade Github und das ist mein erster Pull Request

### DIFF
--- a/examples/ForSFBGallery/20_Pedulum_Sm.html
+++ b/examples/ForSFBGallery/20_Pedulum_Sm.html
@@ -1,0 +1,143 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>WebGL testing</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../../build/js/Cindy.js"></script>
+<script type="text/javascript" src="../../build/js/Cindy3D.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+sec=seconds();
+use("Cindy3D");
+zdat=apply(1..64,1);
+cdat=apply(1..64,1);
+oldcdat=cdat;
+drawging=false;
+count=0;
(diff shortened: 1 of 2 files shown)


gagern:
Since these discussions are public, having them in English will allow others to profit from looking back at them. Of course, that's just a matter of personal taste, and I'd prefer seeing your pull requests written in German over not seeing them at all.

Unfortunately, this pull request mixes two unrelated issues. It would be better to have one pull request per issue, so that each issue could be discussed separately. What is worse: the commit https://github.com/richter-gebert/CindyJS/commit/8ce996d623a0c3b003491d42b155cdf6a4fe7d99 by you're building upon isn't ready for inclusion yet, as indicated by my comments on that commit. I'd not want to merge that into master. @strobelm, I'm not sure if you received the comments on that commit, since I wrote them for the version in this fork here.

The example itself works, looks nice and demonstrates interesting features. The directory `examples/ForSFBGallery` has been absent from the repository so far. Its name suggests that there might be more examples to come. If so, I wonder whether it would be better to have them in a separate repository, because not all of them might be of general interest, and some might contain fairly large amounts of data, adding considerably to the size of the repository. Furthermore, the Apache 2 license applied to the examples so far might not be suitable for all those examples. We could of course add more specific licensing instructions to such a subdirectory, but there is a good chance that this would simply be overlooked, because it might be unexpected. In the end, the choice is of course yours.

If you want me to, I can cherry-pick your example into the master branch, so you don't have to worry about it. If you want to, I can show you how to use `git rebase` to move your commit directly onto the official master, without Michael's commit getting in the way. And if you decide to set up a different repository instead, I can of course help with that, too. Let me know how you'd like to proceed.

strobelm:
Sorry mixed up the topics on my mobile phone. This pull request was never intended to be merged and only for demonstration purposes. Hence it can be closed.

PR #23, opened 2015-10-15
fixed drawing bug for parabolas

strobelm:
For parabolas the conic drawing routine had a bug that did not close the parabola at its vertex. This pull request fixes this which can be seen in examples/Parabel.html.

### DIFF
--- a/examples/Parabel.html
+++ b/examples/Parabel.html
@@ -1,0 +1,101 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Parabel.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+        html{overflow: hidden;}
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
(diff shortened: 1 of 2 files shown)


gagern:
I don't feel I'll be understanding all that drawing code any time soon. But if it improves the example, and offers a clean alternative to a division by zero, that sounds like a clear improvement.

Thinking about the code, I can't help but wonder whether one could still encounter the problem if the numerator is not exactly zero, but still small enough to result in infinities and nans in subsequent computations. But that's a general problem which I think we have in many places. No need to fix this one here now, but we might want to have a look at the big picture for this one day.

strobelm:
Well indeed you are right @gagern. I think i have to rework at least the transformation to canvas coordinates, which is at least likely to introduce numerical instability. 

To your questions regarding zero checks: there is an additional check at line 541 - 548 which sets sane defaults in doubt. 

The basic idea of the conic drawing code is the following: we iterate over the y-value and generate up to 2 solutions (which may be complex). To Speed things up we find the upper and lower bound of the drawing object (e.g. the vertex) and calculate only the necessary == visible parts.
The bounds are the solutions of line 534/535 despite the parabola case i forgot.

PR #25, opened 2015-10-19
Implement 4-ary form of mesh3d providing normals

gagern:
The example shows how these can be used to make a multi-colored surface appear smooth.  However, the colors are still per-face, not per-vertex, so the result doesn't look as smooth as could be hoped.

@richter-gebert: You can add `git@github.com:gagern/CindyJS.git` as a remote to your repository, and check out this branch from there to give it a try. If you agree with these changes, feel free to merge this request using the GitHub interface. Otherwise comment and/or assign the request to someone else.

### DIFF
--- a/src/js/cindy3d/Ops3D.js
+++ b/src/js/cindy3d/Ops3D.js
@@ -358,20 +358,61 @@ createCindy.registerPlugin(1, "Cindy3D", function(api) {
     }
   }
 
+  function meshWithNormalsAndColors(m, n, tcr, tcc, pos, normals, colors, appearance) {
+    for (let i = 1, k = 0; i < m; ++i) {
+      for (let j = 1; j < n; ++j) {
+        currentInstance.triangles.addPolygonWithNormalsAndColors(
+          [pos[k], pos[k + 1], pos[k + n + 1], pos[k + n]],
+          [normals[k], normals[k + 1], normals[k + n + 1], normals[k + n]],
+          [colors[k], colors[k + 1], colors[k + n + 1], colors[k + n]],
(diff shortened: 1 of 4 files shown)


src/js/cindy3d/Ops3D.js
```diff
@@ -323,8 +382,13 @@ createCindy.registerPlugin(1, "Cindy3D", function(api) {
       normal[2] += c[2];
       ++normalcnt;
     }
-    if (normaltype === "pervertex") {
-      normals = Array(m*n);
+    if (args.length === 4) {
+      let normals = coerce.toList(evaluate(args[3]))
+        .map(elt => coerce.toDirection(elt));
+      if (args.length !== m*n) return nada;
```
gagern: This is an error. Will fix it in the next iteration. Shold be `normals.length`.

gagern:
This should be better now: setting the color per vertex, with an automatically computed per-vertex normal, will allow both color and normal fo be interpolated accross each primitive, leading to a smooth look. We don't have per-vertex alpha control so far.

strobelm:
What is the current status of the code? The example files look fine to me.

gagern:
I know @richter-gebert has done some work building on this. I'm not sure whether he only added new functionality, or also had to modify my code. In the latter case, it might be worthwhile to have those modifications included in the same pull, but that's not a strong requirement.

gagern:
OK, @richter-gebert added a new feature (which doesn't exist in the Java version of Cindy3D either), but didn't modify this feature here, so we can merge this and build his feature on the merged version.

PR #26, opened 2015-10-20
arcPPP drawing implemented

strobelm:
Hi everyone,

i implemented the drawing routine for arcPPP. There are example files: examples/104_arcppp.html, examples/104b_arcfill.html.

### DIFF
--- a/ref/Elementary_Drawing_Functions.md
+++ b/ref/Elementary_Drawing_Functions.md
@@ -269,10 +269,11 @@ The following piece of code shows a combined usage of the `drawcircle` and the `
 
 ------
 
-#### Drawing circular arc: `drawarc(‹point› A,<point> B, <point> C)`
+#### Drawing circular arc: `drawarc(‹point1› ,<point2> , <point3> )`
 
 **Description:**
-Draws a circular arc from `‹point›` A to `‹point›` C via  `‹point›` B. 
+Draws a circular arc from `‹point1›` to `‹point3›` via `‹point2›`.
+
 The points may be given either in euclidean or in homogeneous coordinates.
 
 **Modifiers:**
@@ -280,10 +281,10 @@ This operator can handle the same modifiers as the `draw(‹expr›)` operator.
 
 ------
 
-#### Filling circular arcs: `fillarc(‹point› A,<point> B, <point> C)` 
+#### Filling circular arcs: `fillarc(‹point1›,<point2>, <point3>)` 
 
 **Description:**
-Draws the interior of a circular arc `‹point›` A to `‹point›` C via  `‹point›` B.
+Draws the interior of a circular arc `‹point1›` to `‹point3›` via `‹point2›`.
 The points may be given either in euclidean or in homogeneous coordinates.
 
 **Modifiers:**
--- a/ref/Elementary_Drawing_Functions.md
+++ b/ref/Elementary_Drawing_Functions.md
@@ -266,3 +266,32 @@ The following piece of code shows a combined usage of the `drawcircle` and the `
 | ![Image](img/FillCircleX.png) |
 | ----------------------------- |
 | ****                          |
+
+------
+
+#### Drawing circular arc: `drawarc(‹point› A,<point> B, <point> C)`
+
+**Description:**
+Draws a circular arc from `‹point›` A to `‹point›` C via  `‹point›` B. 
+The points may be given either in euclidean or in homogeneous coordinates.
+
+**Modifiers:**
+This operator can handle the same modifiers as the `draw(‹expr›)` operator.
+
+------
+
+#### Filling circular arcs: `fillarc(‹point› A,<point> B, <point> C)` 
+
+**Description:**
+Draws the interior of a circular arc `‹point›` A to `‹point›` C via  `‹point›` B.
+The points may be given either in euclidean or in homogeneous coordinates.
+
+**Modifiers:**
+This operator can handle the following modifiers:
+
+| Modifier | Parameter                   | Effect                                 |
+| -------- | --------------------------- | -------------------------------------- |
+| `color`  | `[‹real1›,‹real2›,‹real3›]` | sets the fill color to an RGB value    |
+| `alpha`  | `‹real›`                    | sets the opacity to the value `‹real›` |
+
+------
--- a/ref/Elementary_Drawing_Functions.md
+++ b/ref/Elementary_Drawing_Functions.md
@@ -269,7 +269,7 @@ The following piece of code shows a combined usage of the `drawcircle` and the `
 
 ------
 
-#### Drawing circular arc: `drawarc(‹point1› ,<point2> , <point3> )`
+#### Drawing circular arc: `drawarc(‹point1›,‹point2›,‹point3›)`
 
 **Description:**
 Draws a circular arc from `‹point1›` to `‹point3›` via `‹point2›`.
@@ -281,7 +281,7 @@ This operator can handle the same modifiers as the `draw(‹expr›)` operator.
 
 ------
 
-#### Filling circular arcs: `fillarc(‹point1›,<point2>, <point3>)` 
+#### Filling circular arcs: `fillarc(‹point1›,‹point2›,‹point3›)`
 
 **Description:**
 Draws the interior of a circular arc `‹point1›` to `‹point3›` via `‹point2›`.
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -121,11 +121,7 @@ eval_helper.drawarc = function(args, modifs, df) {
     var c = evaluateAndHomog(args[2]);
 
     // check for complex values
-    [a, b, c].forEach(function(v) {
-        v.value.forEach(function(vv) {
-            if (Math.abs(vv.value.imag) > CSNumber.eps) return nada;
-        })
-    });
+    if (!List._helper.isAlmostReal(List.turnIntoCSList([a,b,c]))) return nada;
 
     var abcdet = List.det3(a, b, c);
 
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -116,7 +116,6 @@ evaluator.fillarc$3 = function(args, modifs) {
 
 
 eval_helper.drawarc = function(args, modifs, df) {
-    // todo fill
     var v1 = evaluateAndVal(args[0]);
     var v2 = evaluateAndVal(args[1]);
     var v3 = evaluateAndVal(args[2]);
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -116,13 +116,17 @@ evaluator.fillarc$3 = function(args, modifs) {
 
 
 eval_helper.drawarc = function(args, modifs, df) {
-    var v1 = evaluateAndVal(args[0]);
-    var v2 = evaluateAndVal(args[1]);
-    var v3 = evaluateAndVal(args[2]);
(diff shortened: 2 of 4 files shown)


strobelm:
By the way: there is an ugly drawing bug in Chome which prevents the use of csctx.arc, hence i implemented the arc by segements.

gagern:
Nice to have this! It's not present in Cinderella, as far as I can tell. At least on the scripting level, although we do have the `Arc` as a geometric primitive there, so we can build on the scripted version once we get to porting that.

The name `fillarc` feels somewhat strange to me. I wonder whether we should make this `fillsegment` and `fillsector`. The former might be causing some confusion between a circle segment and a line segment, though, so I'm not totally convinced myself. `fillsector` would be nice to have in any case, I think. Since the collinear case there would be even uglier, adding that later on is all right, though.

The case of two rays towards infinity can happen with aprojectively transformed line as well, and isn't handled correctly there at the moment (as of ee4954cdd4ac278c12778119ca187c5df585e474). Perhaps we can share some code between these two use cases.

I'll have a closer look at the changes one of these days. A screenshot for the Chrome bug would be nice to have here on the bug report, but I guess once I start toying with the code I'll be able to produce one such easily enough.

kortenkamp:
I agree, nice to have.

`fillsegment` is a bad name.

Also, Cinderella does not support these as they are causing major headaches for non-Euclidean viewports…

While we are at it: We have several places where we have to "map" geometric primitives (like an arc) to renderings:
* drawing geometric elements
* drawing from CindyScript
* creating shapes
And each of these will create the same "strokes". It is an ongoing challenge to unify all this in Cinderella… and we could/should make sure that we won't have the same problem with CindyJS later on.

strobelm:
The function drawarc and fillarc were intended primary to write the drawing code, without the hassle implementing a geometric primitive (i won't insist on them).
For the latter topic: what is the best approach to handle arcs as geometric primitives? Are there enough use cases to introduce a new kind in geoOps? 

Here is a screenshot of the Chromium bug:
chrome-bug

Firefox (on the right) renders the scene correctly.

Is there a good way to debug the canvas drawing? Chromium recently removed the canvas inspector i used before.

gagern:
> For the latter topic: what is the best approach to handle arcs as geometric primitives? Are there enough use cases to introduce a new kind in geoOps?

We already do have an `Arc` primitive in Cinderella (via menu “Modes / Circle / Circular Arc”) and we should support that on export. That is use case enough, particularly once the rendering is in place.

> Is there a good way to debug the canvas drawing?

You can replace the functions on the canvas object with your own version which prints to console before invoking the original. i can thing of no more clever approach just now. If you can use this approach to isolate the problem, then filing a bug against Chromium might be worthwhile.

gagern:
> Are there enough use cases to introduce a new kind in geoOps?

I guess I misinterpreted that question. Kinds are things like point, line, conic, right? In Cinderella, an Arc is a Circle with some special handling. This is evidenced by e.g. the fact that an arc and a full circle will intersect in two points even if one of the points is outside the portion covered by the arc. And since a circle is just a special conic, the `kind` would be `C`.

Looking at the examples, the feature looks good. Looking at the code, I'll comment on that as I read through it.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+            List.turnIntoCSList([zer, zer, cen.value[2]])
+        ]);
+        var aa = List.normalizeZ(General.mult(mat, a));
+        var bb = List.normalizeZ(General.mult(mat, b));
+        var cc = List.normalizeZ(General.mult(mat, c));
+
+
+        // get angles of A and C 
+        var startAngle = -Math.atan2(aa.value[1].value.real, aa.value[0].value.real);
+        var endAngle = -Math.atan2(cc.value[1].value.real, cc.value[0].value.real);
```
gagern: I'm a bit unsure about how you switch from a complex-valued implementation to a real-valued one. Perhaps before you start computing `atan2` you could check whether there are any imaginary parts, and return early if there are?

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+
+        cen = List.normalizeZ(cen);
+        var arcDist = List.abs(List.sub(a, cen));
```
gagern: I wonder whether here could be any reason to compute the radius directly from the matrix, instead of via a distance as you did. Probably not, just thinking out loud. This *is* the radius, right? Because the term `arcDist` sounds so strange if there is such a simple term instead.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+
+eval_helper.drawarc = function(args, modifs, df) {
+    var v1 = evaluateAndVal(args[0]);
+    var v2 = evaluateAndVal(args[1]);
+    var v3 = evaluateAndVal(args[2]);
+
+    var a = List.normalizeZ(csgeo.csnames[args[0].name].homog);
+    var b = List.normalizeZ(csgeo.csnames[args[1].name].homog);
+    var c = List.normalizeZ(csgeo.csnames[args[2].name].homog);
+    var abcdet = List.det3(a, b, c);
```
gagern: I wonder what happens if one of the points is at infinity. If `b* is atinfinity, then you *should* get a pair of rays, i.e. what you get in the collinear case if `b` is outside the segment `ac`. The direction of `b` is irrelevant. If `a` is anywhere at infinity, then you'd get a single ray starting at `c` and passing through `b`, and conversely if `c` is at infinity. Long story short: it might be better to move that normalization to somewhere where these special cases have already been excluded. If you want to you can leave that for a later commit, though.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+        csctx.strokeStyle = Render2D.lineColor;
+
+        // canvas circle radius 
+        var rad = arcDist.value.real * m.sdet;
+
+        csctx.beginPath();
+        csctx.translate(xx, yy);
+
+        // use the canvas arc function -- buggy in Chrome at least in Okt 15
+        var useArc = false;
```
gagern: I can reproduce this on OS X.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+
+        if (df === "D") {
+            csctx.stroke();
+        }
+        csctx.restore();
+
+    } else { // segment case
+        var ptA = eval_helper.extractPoint(v1);
+        var ptB = eval_helper.extractPoint(v2);
+        var ptC = eval_helper.extractPoint(v3);
```
gagern: `extractPoint` should be followed by checks against `ptA.ok` and the likes. Just return early if one of them is not `ok`.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+        if (dAB < 1e-12 || dAB < 1e-12 || dAB < 1e-12) return nada;
+
+        // drawing points
+        var pt1, pt2, pt3;
+
+        // check by dets if B is in the middle
+        var randP = List.realVector([10 * Math.random(), 10 * Math.random(), 10 * Math.random()]);
+        var detABP = List.det3(a, b, randP).value.real;
+        var detACP = List.det3(b, c, randP).value.real;
+        var Bmiddle = detABP * detACP > 0;
```
gagern: I guess this works, too. For comparison, if you wanted to do this deterministically, you'd take `d` as the far point on the line, then compute the cross ratio `cr(a,c;b,d)` and check it's sign. You don't have to do it, I'm just offering this in case it's useful here or somewhere else.

src/js/libcs/OpDrawing.js
gagern: Why access `csgeo` here? After all, the arguments could already be some user-provided coordinate vector, and not a geometric element at all. Better use `evaluateAndHomog` instead of `evaluateAndVal`, and let that take care of turning the input into homogeneous coordinates in whatever way seems appropriate.

ref/Elementary_Drawing_Functions.md
```diff
@@ -266,3 +266,32 @@ The following piece of code shows a combined usage of the `drawcircle` and the `
 | ![Image](img/FillCircleX.png) |
 | ----------------------------- |
 | ****                          |
+
+------
+
+#### Drawing circular arc: `drawarc(‹point› A,<point> B, <point> C)`
```
gagern: A good thing you thought about adding documentation! I might have forgotten it. Please use the angled quotation marks seen in the first argument also for the other two. This convention might one day be important if we e.g. try to automatically compute arity of documented functions, in order to add stable anchor names. The same holds for `fillarc`.

gagern: Come to think of it, this distinction between type and name is unprecedented in the documentation, as far as I can tell. Other locations with multiple arguments of equal type tend to number their arguments. So you'd write `drawarc(‹point1›,‹point2›,‹point3›)`. The idea as I understand it is that `‹point1›` as a whole serves as a placeholder.

Personally I'm open to a change of that convention, but I think it should be discussed and perhaps even be applied consistently to existing documentation.

src/js/libcs/OpDrawing.js
```diff
@@ -106,6 +106,183 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
+
+            // first ray
+            // get direction and normalise
+            var dx = sflip * (ptA.x - ptB.x);
+            var dy = sflip * (ptA.y - ptB.y);
+            var norm = Math.sqrt(dx * dx + dy * dy);
+            dx = dx / norm;
+            dy = dy / norm;
+
+            var ptATmp = eval_helper.extractPoint(v1);
```
gagern: Why not just reuse `ptA` resp. `ptC` here? Am I missing something, or should these be exactly like this new invocation of `extractPoint`? If, for some reason I don't see, `extractPoint` is really required, you'll again have to check `ptATmp.ok` I guess.

strobelm:
Thank @gagern for the useful annotations, i'll update the branch today.

strobelm:
I added most of the suggestions, i was not sure about the documentation whether this right format. Correct me if a was wrong.

gagern:
The changes to the implementation look very good except for the reality check, as commented on the commit.

The documentation still has the wrong kind of angled brackets: `<…>` (U+003C and U+003E) is what you have for `point2` and `point3`, `‹…›` (U+2039 and U+203A) is what the rest of the documentation uses. And the rest of the doc doesn't use spaces in the argument list.

```diff
-#### Drawing circular arc: `drawarc(‹point1› ,<point2> , <point3> )`
+#### Drawing circular arc: `drawarc(‹point1›,‹point2›,‹point3›)`
-#### Filling circular arcs: `fillarc(‹point1›,<point2>, <point3>)`
+#### Filling circular arcs: `fillarc(‹point1›,‹point2›,‹point3›)`
```

strobelm:
Thanks for the patches @gagern ! Concerning the randomization part: i like the idea about the crossrations, it makes it more clear what the code should actually provide.

gagern:
https://github.com/CindyJS/CindyJS/pull/26#commitcomment-13950267 not addressed: the complex check may fail if the representatives are chosen in an unfortunate way. It's probably a rare case (thanks to normalizations all over the place), so I'm willing to leave it at this for now.

Thanks for making this work, and for implementing the review suggestions to make it work even better!

PR #35, opened 2015-10-23
Move default-setting functions out of csinit

gagern:
This is in preparation for the implementation of create, where these
functions might get called from different places.  There is no urgent need
to have them enclosed in csinit, so let's avoid that.  Since various other
planned modifications touch this code, having the move as a separate (and
hopefully non-controversial) commit will help building the rest on that.

The only change in this commit is position and whitespace.


strobelm:
The modifications look fine, thanks @gagern.

PR #36, opened 2015-10-24
fix typo in drawarc

strobelm:
This pull request fixes a typo which prevents an intended early return in drawarc.


gagern:
Whoops. Missed that in the review, too.

PR #37, opened 2015-10-26
Improve deployment

gagern:
* Drop files only neccessary for development
* Include some licensing information in header of minified sources
* Include commit id in minified file to quickly identify any given version
* Adjust source maps to refer to GitHub source repository

### DIFF
--- a/src/js/Cindy.js.wrapper
+++ b/src/js/Cindy.js.wrapper
@@ -1,5 +1,7 @@
-/* CindyJS - (C) 2014-2015  J. Richter-Gebert
- *
+/* CindyJS - (C) 2014-2015  The CindyJS Project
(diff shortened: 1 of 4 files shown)


src/js/Cindy.js.wrapper
```diff
@@ -1,5 +1,7 @@
 /* CindyJS - (C) 2014-2015  J. Richter-Gebert
```
kranich: Shouldn't this be "The CindyJS Project" instead of "J. Richter-Gebert"? Or at least "Jürgen Richter-Gebert"?

gagern: I didn't modify that line, but I assume you are right. Will augment my commit in this regard.

PR #42, opened 2015-10-27
Draw trace

gagern:
This is the the trace-drawing code by @wursterje which I polished somewhat. It's not feature-complete yet, since it lacks dimension adjustments and traces for lines, but it's a huge step forward and since I don't know when the other features will get implemented, I'd like to consider this for inclusion now.


strobelm:
Hi, the code looks good. Thanks for implementing this.
I was wondering whether it would be faster to maintain a list of points which have a trace instead of iterating over all points on every update and check if it has a trace or not. Am i missing something?

gagern:
@strobelm: I was wondering about the same thing. I even have a git stash on my hsdd with some preliminary work in that direction. But I'm no longer sure it's worth the effort.

We don't maintain a list of unpinned visible movable objects, [`getmover`](https://github.com/CindyJS/CindyJS/blob/b61f38e5179251317452248b269e0ef28a0509db/src/js/Events.js#L12-L15) just iterates over all potentially free objects and discards pinned and invisible ones. The same holds [rendering](https://github.com/CindyJS/CindyJS/blob/b61f38e5179251317452248b269e0ef28a0509db/src/js/libgeo/GeoBasics.js#L306). If we want to avoid unneccessary work, shouldn't we maintain lists of visible points resp. visible and unpinned objects? And if we do, what would that do for our code complexity.

I think if there are few points in the gslp, then iterating over all of them won't cost much performance since they are few. And if there are many, then rendering them all will be the real bottleneck, and one more iteration should be negigible by comparison. Only if there are many invisible points, then we have one more loop which might cause work comparable to that of skipping all the points during rendering. But I can think of no real world scenario where many invisible points would make any kind of sense.

On the other hand, the fact that we both thought of having a separate list might be seen as an indication that such a setup would be “the thing to do”. And it would be the only way to guarantee that the ability to draw traces has negligible performance impact on content which doesn't use it, namely one `=== null` per rendering phase. I guess I'll get out my stash, finish it up and push it either to this branch here if I like it well enough, or to a different branch if I'm less certain.

One thing we should keep in mind is that as soon as we can enable and disable tracing from CindyScript (via `inspect`), we'd have to update such a list whenever a trace gets enabled or disabled. Which shouldn't bother us in terms of performance since I assume that's rare enough, but it might be quite a bother in terms of code complexity, since it means more places where we have to deal with these data structures.

strobelm:
Agreement by merge. 
Probably it's the right way to do but given the other (large) tasks which are on the list perhaps we put this aside for a while.

PR #47, opened 2015-10-28
Numericjs drop

### LINKED ISSUE #45: dropping numeric.js
Hi! I had a look into our linear algebra routines and noticed that numeric.js is hardly used in our algorithms anymore. The only real usage is currently in the conic drawing routine where i rotate the conic for some reasons, which is only a 3x3 matrix multiplication. I could easily patch the code in order to use List.*. 

Furthermore there are List.inversereal, List.linearsolvereal and List.detreal which aren't used in productive code at all (as far i can see).

Therefor is there any reason to keep numeric.js? The benefit would be a 10% saving in the current Cindy.js filesize (minimized version).

strobelm:
Hi! This pull request will resolve #45. It drops all deprecated numeric.js routines from List.js, changes the conic routine to use List.js and finally removes numeric.js from the Makefile und the lib itself.

I could not find any usage of the routines at all (I == grep).


strobelm:
I think the code is fine, although i would feel better if someone was looking whether i missed some numeric.js code that is still used.

gagern:
I have verified that conics still get drawn. I have verified that the three functions you removed from `List` are really unused. I verified that there are no references to the `numeric` namespace. And I have tried to also check whether we have any references to the `T` type. Travis also looked over these changes. Everything looks fine, so I'll merge.

gagern:
Nice work, nicely structured commits, thanks a lot!

PR #57, opened 2015-10-29
Detect degenerate situations

### LINKED ISSUE #61: Tracing Problem
*(Manually migrated from GitLab issue 9 reported by @richter-gebert)*

Tracing breaks down on [this example](https://gist.githubusercontent.com/gagern/3291d83aa7e39b2b93a2/raw/27201e5f5087f06388c8c9f7c423b125073f07eb/Byrne1.html):

```js
createCindy({
  scripts: {},
  defaultAppearance: {},
  ports: [{
    id: "CSCanvas",
    width: 853,
    height: 588,
    background: "rgb(168,176,192)",
    transform: [{visibleRect: [-9.04, 9.32, 25.08, -14.2]}]
  }],
  geometry: [
    {name: "A", type: "Free", pos: [4, -0.4055299539170507, 0.4608294930875576], size: 0, color: [1, 0, 0]},
    {name: "B", type: "Free", pos: [1.1494252873563218, -4, 1.1494252873563218], size: 0, color: [1, 0, 0]},
    {name: "C", type: "Free", pos: [4, -1.897142857142857, 0.5714285714285714], size: 0, color: [1, 0, 0]},
    {name: "a", type: "Segment", args: ["B", "C"], pos: [-0.030418250950570432, 1.1406844106463878, 4], size: 5, color: [0, 0, 0]},
    {name: "b", type: "Segment", args: ["A", "C"], pos: [-0.43076053950992166, 0.296589223924864, 4], size: 5, color: [0, 0, 0]},
    {name: "C0", type: "CircleMP", args: ["A", "C"], size: 5, color: [0.14117648, 0.44705886, 0.54901963], printname: "$C_{0}$", visible: false},
    {name: "C1", type: "CircleMP", args: ["C", "A"], size: 5, color: [0.14117648, 0.44705886, 0.54901963], printname: "$C_{1}$", visible: false},
    {name: "Collection__1", type: "IntersectionCircleCircle", args: ["C0", "C1"]},
    {name: "D", type: "SelectP", args: ["Collection__1"], pos: [4, -0.45055967120693136, 0.6984583957469794], size: 0, color: [1, 0, 0]},
    {name: "", type: "SelectP", args: ["Collection__1"], pos: [4, -1.4286692464848771, 0.401884759739649], size: 0, color: [1, 1, 1]},
    {name: "c", type: "Segment", args: ["D", "C"], pos: [-0.738023708295076, -0.35125480664624464, 4], size: 5, color: [0.7568628, 0, 0]},
    {name: "d", type: "Segment", args: ["D", "A"], pos: [-0.31820462623032214, -4, -0.7579838443208037], size: 5, color: [0.7568628, 0, 0]},
    {name: "C2", type: "CircleMP", args: ["C", "B"], size: 5, color: [0.14117648, 0.44705886, 0.54901963], printname: "$C_{2}$"},
    {name: "e", type: "Join", args: ["D", "A"], pos: [-0.31820462623032214, -4, -0.7579838443208037], size: 5, color: [0.14117648, 0.44705886, 0.54901963], visible: false},
    {name: "f", type: "Join", args: ["D", "C"], pos: [-0.738023708295076, -0.35125480664624464, 4], size: 5, color: [0.14117648, 0.44705886, 0.54901963], visible: false},
    {name: "Collection__2", type: "IntersectionConicLine", args: ["C2", "f"]},
    {name: "F", type: "SelectP", args: ["Collection__2"], pos: [4, -3.649333562406507, 0.4175622195823888], size: 0, color: [1, 0, 0]},
    {name: "h", type: "Segment", args: ["C", "F"], pos: [-0.7380237082950761, -0.3512548066462446, 4], size: 5, color: [1, 0.49803925, 0]},
    {name: "C3", type: "CircleMP", args: ["D", "F"], size: 5, color: [0.7568628, 0, 0], printname: "$C_{3}$"},
    {name: "Collection__3", type: "IntersectionConicLine", args: ["C3", "e"]},
    {name: "E", type: "SelectP", args: ["Collection__3"], pos: [4, -0.36989745294140725, 0.2727911793814279], size: 0, color: [1, 0, 0]},
    {name: "g", type: "Segment", args: ["A", "E"], pos: [-0.31820462623032186, -4, -0.7579838443208057], size: 5, color: [0.14117648, 0.44705886, 0.54901963]}
  ]
});
```

-----

*(Comment originally by @kranich)*

Only in cindyjs, or also in Cinderella?

-----

*(Comment originally by @kranich)*

@wursterje In cases like this, it would be convenient to be able to (re)import from cindyjs to Cinderella.

-----

*(Comment originally by @gagern)*

Cinderella version is somewhat like [this](https://gist.githubusercontent.com/gagern/3291d83aa7e39b2b93a2/raw/27201e5f5087f06388c8c9f7c423b125073f07eb/Byrne1.cdy) (save as `*.cdy`). Doesn't break tracing.

The way I see it, we have a removable singularity here, and I'd like to understand expected behavior for this case a bit better before investigating actual behavior. How should tracing decide between passing through and removing a singularity on the one hand, and walking around to evade it on the other hand? Both kinds of behavior can be seen with Cinderella as well.

Without looking closer at the tracing decisions here, I just reduced `traceLimit` from `10000` to `100`. This kept the UI responsive. Occasionally (and luckily only rarely) I'd see a log message that that limit had been reached. Interestingly, those log messages tend to occur in bursts. So apparently when there is one bad trace step, others are likely to follow. On the other hand, I can't see any obvious loop around `traceMover`, and the rate (ca. 200 events per burst) looks too high to be completely mouse-driven, I think. Strange.

If we need a quick workaround, we could commit `traceLimit = 100`, but that's really just a workaround, and we should make an effort to understand what's really going on here.

-----

*(Comment originally by @kranich)*

For a removable singularity, going straight through it and taking a (small enough) complex detour should yield the same result.

So maybe we should just keep taking complex detours instead of trying to detect this special case.

If the singularity in the example is removable then maybe tracing catches a nearby complex singularity? This would at least explain the high number of subdivisions.

-----

*(Comment originally by @gagern)*

To be more precise: whether or not the singularity is removable depends on the path: moving A on a straight line through C gives a different result than *slowly* moving around C. If one moves quickly enough on a path that goes slightly past C, the singularity gets “incorrectly removed”. It's very much like the intersection of two circles of equal size where one center moves along a line that *almost* passes through the other center. You know that example, Stefan. In that case, the singularity is slightly off the real axis, and a quick movement may take the wrong path around it, different from what a purely real movement would do.

-----

*(Comment originally by @kortenkamp)*

@kranich In Cindy Classic we always go the detour – as you say, it should not matter. What can happen is that you catch a nearby singularity, but that's true anyway. And usually singularities do not lie in (non-real) complex space near to a removable singularity (ok, that's just experience, no proof given).

-----

*(Comment originally by @kranich)*

@kortenkamp It would be easy to construct an algebraic curve which has such a singularity distribution – maybe more difficult to find a corresponding geometric construction. That this "usually" does not happen is probably right though. :-)

@gagern I believe that the behaviour you describe is to be expected. It is always advisable to move slowly close to singularities. If you make steps which are much too large, you maybe need not even encircle a singularity to obtain jumps.

-----

*(Comment originally by @kranich)*

I just managed to make points swap in Cinderella, too (using a series of fast movements of A close to C).

It seems cindyjs is simply much more sensitive in this case.

-----

*(Comment originally by @gagern)*

I think the almost removable singularity and the swap it can cause is not the problem here. Both non-swapping and swapping behavior are justified, in a certain sense, due to the poor definition of the path at high speed. The real problem is the fact that CindyJS has a good chance of getting hung up in that situation for several seconds that feel like an eternity. The observations about the nature of the singularity are just becase that *might* have something to do with the hang-up, one way or another.

I wonder whether the following could be what's happening here: perhaps the mouse is moving *very* close to the singularity, to a situation that is *almost* singular. But not quite. Afterwards, any subsequent movement will use that as a the last good position, will have to refine a lot, hit the refinement limit and won't result in a new good position. Therefore the mouse position after that experiences the same thing. So the next 200 mouse events hit the refinement limit of 100 steps although the destination points are already far away from the singularity, simply because the source point is so close to the singularity and because the two different branches (swapping and non-swapping) are so close to one another. I wonder whether it would be a good idea to have *multiple* past good positions, so that the more often we hit the refinement limit, the further back in that history we can choose our starting point in order to take a larger detour around the singularity. Not sure how to implement this, just thinking.

-----

*(Comment originally by @gagern)*

    While investigating this issue, I'm documenting my debugging techniques. Adding `enableTraceLog: 500` to the `createCindy` call gave me access to the tracing data for the past 500 mouse events. Storing the log to file failed on Chrome since the stringified JSON is larger than `(1 << 28) - 16` characters, the internal limit of V8. Instead I used the JavaScript console and wrote this:

```js
    log = cdy.getTraceLog();
    d={};
    log.value.forEach(function(mouse){
      mouse.value[0].value.forEach(function(mover){
        var k=mover.value[0].value.length;
        d[k]=(d[k]||0)+1;
      });
    });
    d
```

The result indicated that there is a single call to `traceMover` spanning 41,060 steps, while all other such calls are done in fewer than 50 steps. So my assumption about multiple mouse positions starting from a bad position and taking too much time is wrong. And the count of 41,060 is considerably larger than the `traceLimit` of 10,000, so that limit is not enforced as strongly as I had expected.

Looking at the sources, I see that the only effect of a reached `traceLimit` is that further refinements are prohibited. But the stepsize, which may already be quite small by that time, is not enlarged again. I think that if the trace limit is reached, we want to either trace to the end in a single step or give up, i.e. do something like this:

```js
        var t = last + step;
        while (last !== t) {
            if (traceLimit === 0) {
                console.log("tracing limit reached");
                t = 1;
            }
```

This guarantees that the message about a reached tracing limit is only printed once, but the number of step in a subsequent problematic run was still in excess of 40,000. Perhaps because the `traceLimit` is only increased when refining, so successful steps at fine step sizes don't count. Might want to change that, too, but this whole `traceLimit` is more of a hack to mitigate worst-case behavior.

My next step would be trying to understand the same situation in Cinderella, and in particular find out whether it manages to trace correctly in a shorter time, or to detect its inability to trace correctly somewhat earlier, or whether it, too, has some heuristic to detect and abort a trace that is taking too long.

One more piece of information: the distinction between `1e-5` and `1e-6` in the `tracing2` implementation in Cinderella doesn't seem to be related to this issue here. I changed the two relevant constants in the CindyJS `tracing2core` to `1e-6` just as Cinderella has them, and could still reproduce the problem with that choice.

-----

*(Comment originally by @gagern)*

@kranich, since you are our expert on visualizing singularities, do you have a good idea of how to perform a visualization of the parameter space of this configuration where we could see the tracing path in relation to the singularity, and gain some intuition as to whether the extemely small steps are actually necessary or should be avoided in the first place?

-----

*(Comment originally by @gagern)*

Some more debug info:

```js
    log = cdy.getTraceLog();d={};
    log.value.forEach(function(mouse){
      mouse.value[0].value.forEach(function(mover){
        mover.value[0].value.forEach(function(step){
          step.value[0].value.forEach(function(elt){
            var k = elt.value[1].value + ": " + elt.value[8].value;
            d[k]=(d[k]||0)+1;
    });});});});d
```

yields the following (in a different run):

```
    Collection__1: Need to refine.: 326
    Collection__1: Normal case, everything all right.: 84278
    Collection__1: Reached refinement limit, giving up.: 2
    Collection__2: Need to refine.: 9342
    Collection__2: Normal case, everything all right.: 11969
    Collection__2: Reached refinement limit, giving up.: 20829
    Collection__3: Need to refine.: 711
    Collection__3: Normal case, everything all right.: 22154
    Collection__3: Reached refinement limit, giving up.: 42020
```

So most of the refinements are requested by the tracing2 call in the first of the two `IntersectionConicLine` calls. Indeed, one can remove the elements `C3`, `E` and `g` from the gslp and still reproduce this problem. Also notice that there are no outcomes titled “Moved into singularity.” or similar, so the situation is never considered singular by our code.

If I change the gslp so that `f` is the line joining `A` and `C` instead of `D` and `C`, then `Collection__2` no longer depends on `Collection__1`. In that case, I can get the construction to hang for some time as well, but never long enough to trigger the `traceLimit`. And it does detect singular situations in that setup.

I think I have an idea as to what's going on here. As points `C` and `D` are coming closer together, the line joining these will become poorely defined. But due to our call to `normalizeMax` the resulting vector gives no indication to this effect. So we have a line which is mostly numeric noise, but the algorithm assumes that it depends on the mover in an algebraic way.

Perhaps we should have some kind of singularity detection for single-valued operations like `Join`, to detect and avoid such problems. Will have to experiment. Does Cinderella do something like this? I don't think I have read such a thing in its sources.

-----

*(Comment originally by @gagern)*

Perhaps `normalizeMaxValid` in Cinderella has something to do with this. Although I don't see `Tracing2Algorithm` using that `Vec.invalid` information anywhere. But perhaps it's not the flag which matters, perhaps it's because in the invalid situation it will return vectors which are exactly zero, which presumably would be detected by the `Tracing2Algorithm` case distinctions.

-----

*(Comment originally by @gagern)*

I just modified `normalizeMax` like this (as an experiment):

```js
    List.normalizeMax = function(a) {
        var s = CSNumber.inv(List.maxval(a));
        var r = s.value.real, i = s.value.imag;
        if (r * r + i * i > 1e17)
            s = CSNumber.zero;
        return List.scalmult(s, a);
    };
```

But I still get `tracing limit reached` situations, this time with a large bunch of `DIVISION BY ZERO` messages before that point. The latest statistics read like this:

```
    Collection__1: Moved into singularity.: 2
    Collection__1: Moved out of singularity.: 70
    Collection__1: Need to refine.: 2335
    Collection__1: Normal case, everything all right.: 99314
    Collection__1: Reached refinement limit, giving up.: 2
    Collection__1: Staying inside singularity.: 32
    Collection__2: Need to refine.: 10598
    Collection__2: Normal case, everything all right.: 21988
    Collection__2: Reached refinement limit, giving up.: 17107
    Collection__2: Tracing failed due to NaNs.: 17
```

Not sure yet how to interpret this.

gagern:
_(Manually migrated from GitLab merge request 12 reported by @gagern)_

This here is an attempt to detect degenerate situations while tracing, in order to abort tracing and not end up with excessive amounts of refinements. It should fix #61.

This solution still feels slightly hackish, so I'll welcome any input on how to make it cleaner while still achieving its designated goal. @kranich, @strobelm: any ideas?

---

_(Comment originally by @gagern)_

I should have pointed out that @kanich already did contribute to this approach, so the idea of detecting degenerate situations right where they occur and throwing an exception was his. Still, I hope for some ideas on how to make this cleaner. If we don't come up with anything soon, we might still merge this as is, as long as we are convinced that it doesn't hurt. Are we?

---

_(Comment originally by @strobelm)_

If we follow this approach: Shouldn't we also check in `normalizeMax` whether we normalize by a maximum which is almost zero?

---

_(Comment by @kortenkamp removed since it might contain copyright-protected material)_

---

_(Comment originally by @strobelm)_

@kortenkamp Martin and i had a look on tracing4 and we were not quite sure how an object which was marked invalid is handled in the context of tracing. Could you give us a hint where the magic happens? ;-)

### DIFF
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -535,6 +535,11 @@ List.maxIndex = function(lst, fun, startIdx) {
 List.normalizeMax = function(a) {
     var s = CSNumber.inv(List.maxval(a));
     if (!CSNumber._helper.isFinite(s)) return a;
+    var r = s.value.real, i = s.value.imag;
+    if (r * r + i * i > 1e17) {
+        degenerateSituationDetected();
+        s = CSNumber.zero;
+    }
     return List.scalmult(s, a);
 };
 
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -452,6 +452,20 @@ List._helper.isNaN = function(a1) {
     return erg;
 };
 
+List._helper.isFinite = function(a1) {
+    var erg = true;
+    for (var i = 0; i < a1.value.length; i++) {
+        var av1 = a1.value[i];
+
+        if (av1.ctype === 'list') {
+            erg = erg && List._helper.isFinite(av1);
+        } else {
+            erg = erg && CSNumber._helper.isFinite(av1);
+        }
+    }
+    return erg;
+};
+
 
 List.set = function(a1) {
     var erg = [];
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -535,7 +535,8 @@ List.maxIndex = function(lst, fun, startIdx) {
 List.normalizeMax = function(a) {
     var s = CSNumber.inv(List.maxval(a));
     if (!CSNumber._helper.isFinite(s)) return a;
-    var r = s.value.real, i = s.value.imag;
+    var r = s.value.real;
+    var i = s.value.imag;
     if (r * r + i * i > 1e17) {
         degenerateSituationDetected();
         s = CSNumber.zero;
--- a/src/js/libgeo/Tracing.js
+++ b/src/js/libgeo/Tracing.js
@@ -77,6 +77,19 @@ var RefineException = {
     }
 };
 
+var DegenerateException = {
+    toString: function() {
+        return "DegenerateException";
+    }
+};
+
+var mayThrowDegenerateException = false;
+
+function degenerateSituationDetected() {
+    if (mayThrowDegenerateException)
+        throw DegenerateException;
+}
+
 function requestRefinement() {
     // Call this whenever you would need exra refinement.
     // Possible outcomes: either an exception will be thrown to
@@ -159,13 +172,16 @@ function traceMover(mover, pos, type) {
     //console.log("Tracing from " + niceprint(originParam) + " to " + niceprint(targetParam));
     var t = last + step;
     while (last !== t) {
+        if (traceLimit === 0)
+            t = 1;
+
         // Rational parametrization of semicircle,
         // see http://jsperf.com/half-circle-parametrization
         var t2 = t * t;
         var dt = 0.5 / (1 + t2);
         var tc = CSNumber.complex((2 * t) * dt + 0.5, (1 - t2) * dt);
         noMoreRefinements = (last + 0.5 * step <= last || traceLimit === 0);
-        if (traceLimit === 0) console.log("tracing limit Reached");
+        mayThrowDegenerateException = (t < 1);
         var refining = false;
 
         if (traceLog && traceLog.currentMouseAndScripts) {
@@ -174,11 +190,20 @@ function traceMover(mover, pos, type) {
         try {
             traceOneStep();
         } catch (e) {
+            if (e === DegenerateException) {
+                console.log(e.toString());
+                tracingFailed = true;
+                traceLimit = 0; // triggers noMoreRefinements
+                t = 1; // no more steps, go straight to the end
+                continue; // doesn't add to log! TODO!
+            }
             if (e !== RefineException)
                 throw e;
             step *= 0.5; // reduce step size
             t = last + step;
             --traceLimit;
+            if (traceLimit === 0)
+                console.log("tracing limit reached");
             refining = true;
         }
         if (traceLog && traceLog.currentMouseAndScripts) {
@@ -193,6 +218,7 @@ function traceMover(mover, pos, type) {
             traceLog.currentParam = null;
         }
     }
+    mayThrowDegenerateException = false;
     tracingStateReport(tracingFailed);
     for (i = 0; i < deps.length; ++i) {
         el = deps[i];
--- a/src/js/libgeo/Tracing.js
+++ b/src/js/libgeo/Tracing.js
@@ -399,11 +399,11 @@ function tracing2core(n1, n2, o1, o2) {
             // Evil: modify can break copy on write! But it's safe here.
         };
     }
-    if (List._helper.isNaN(n1) || List._helper.isNaN(n2)) {
+    if (!(List._helper.isFinite(n1) && List._helper.isFinite(n2))) {
         // Something went very wrong, numerically speaking. We have no
         // clue whether refining will make things any better, so we
         // assume it won't and give up.
-        debug("Tracing failed due to NaNs.");
+        debug("Tracing failed due to NaNs or Infinity.");
         tracingFailed = true;
     } else if (do1o2 > cost * safety && dn1n2 > cost * safety) {
         // Distance within matching considerably smaller than distance
@@ -522,11 +522,11 @@ function tracing4core(n1, n2, n3, n4, o1, o2, o3, o4) {
 
     for (i = 0; i < 4; i++) {
         if (need_refine) break;
-        if (List._helper.isNaN(new_el[i])) {
+        if (!List._helper.isFinite(new_el[i])) {
             // Something went very wrong, numerically speaking. We have no
             // clue whether refining will make things any better, so we
             // assume it won't and give up.
-            debug("Tracing failed due to NaNs.");
+            debug("Tracing failed due to NaNs or Infinity.");
             tracingFailed = true;
             return res;
         }


strobelm:
Hey, i created a rebased and beautified version of the branch:
https://github.com/strobelm/CindyJS/tree/DegenerateException

gagern:
@strobelm: Thanks, reset my branch to your version of it.

strobelm:

PR #58, opened 2015-10-31
Quickhull3d on demand

### LINKED ISSUE #51: Clean up status of convexhull3d
Currently, the [`convexhull3d` operator](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/Operators.js#L3595) only works if it can call an external function `convexhull` [defined](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/java/cindyjs/quickhull3d/JavaScriptBindings.java#L20-L21) by the GWT-compiled `quickhull3d` code. So that code has to be loaded for the operator to work; otherwise an exception will be thrown.

Possible solutions include

1. Load the code of `quickhull3d` unconditionally before starting the first CindyJS instance
2. Make `quickhull3d` into a regular plugin, and detect the need for it in Cinderella upon export
3. Load code on demand if a given CindyJS instance seems to need it, before starting that instance
4. Port code to JavaScript, assuming that the result will be easier to handle than the GWT-compiled code

gagern:
This turns quickhull3d into a proper plugin, using the plugin API. It detects upon instance initialization whether said plugin is needed (i.e. if `convexhull3d` is called as a function from CindyScript) and loads it on demand. As an extra benefit, we're getting rid of Ant, using the command line interface to GWT (which I recently learned to use) instead.

This is one way to fix #51.


strobelm:
Hi, i tested the code by looking over almost all examples which worked nicely (despite the problem with physics which wasn't merged to the branch).

Issue #31, opened 2015-10-21
(0,0) is not a pixel, but between pixels


gagern:
*(Manually migrated from GitLab issue 5 reported by @kortenkamp)*

We are running into an issue that is well known from the Java 1.0 -> Java 1.2 transition, wenn non-integer pixel coordinates were introduced.

Background: The location of (0,0) in a pixel image is a well defined pixel. The location of (0.0,0.0) is in the middle of the pixels (-1,0), (1,0), (0,-1) and (0,1). Drawing a pixel at (0,0) will either fill the pixel exactly, or distribute the color of the pixel to the four surrounding pixels.

In Cinderella Classic, the (0,0) coordinate is therefor shifted by 0.5 pixels in both x and y direction, resulting in crisp lines:

![Cinderella Screenshot](https://cloud.githubusercontent.com/assets/190759/10651798/d4a3af68-7851-11e5-8701-4f23f7179da5.png)

In CindyJS, this is not the case, resulting in a blurred line:

![CindyJS Screenshot](https://cloud.githubusercontent.com/assets/190759/10651806/dc085222-7851-11e5-9489-f1b3675fde4b.png)

I prefer the crisp lines and suggest to adapt Render2D accordingly. What do you think?

-----

*(Comment originally by @gagern)*

I'm not sure I fully understand the situation. Having the shifted grid results in smooth horizontal lines *if* the *y* coordinate of said lines is an integer. How likely is that to be, in particular after the user used the zoom at least once? After all, user coordinates are a lot different from pixel coordinates. I assume there is some mechanism which increases the probability of integer user space coordinates ending up at integer (or after the 0.5 shift exactly between integer) pixel coordinates. But what mechanism is that? My most likely guess would be the [stroke control rendering hint](http://docs.oracle.com/javase/8/docs/api/java/awt/RenderingHints.html#KEY_STROKE_CONTROL). I can see no mention of `STROKE_CONTROL` in the driftin sources, and I believe that it is enabled by default.

Or is this really only about the origin, with no regard for other integer positions? Is there some mechanism rounding the position of the origin to integer coordinates even when the user moves the canvas around?

If we just want a 0.5px shift, so that the origin ends up in the center of a pixel (not a pixel corner), then I'd not do so in Render2D but instead modify the initial transformation matrix. If we want to emulate something like stroke control, things will become way more difficult. Unless tehre is some way to let the browser do this, which I haven't found yet.

-----

*(Comment originally by @kortenkamp)*

[...]

PR #59, opened 2015-11-02
Shift default origin by half a pixel

gagern:
By shifting the default position of the origin, integer user coordinates have a higher chance of ending up at the center of a pixel, as opposed to the corner of a pixel.

This will affect constructions which use the default coordinate system, as well as those which apply relative transformations to that default, like `scale` and `translate`.  Absolute transformations like `scaleAndOrigin` and `visibleRect` are not affected.

This addresses #31.

Since this is an *incompatible* change, I'd like an OK from Jürgen @richter-gebert on this. Can we just change the default, adjusting any existing content which relies on pixel-exact positioning using the old defaults?

### DIFF
--- a/ref/Control_Operators.md
+++ b/ref/Control_Operators.md
@@ -201,76 +201,76 @@ This code fragment will draw a 3 × 3 array of points.
     > );
     D lineWidth = 1.2
     D beginPath()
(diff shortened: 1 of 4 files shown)


strobelm:
I tested all examples visually and found no problems.

gagern:
Thanks for the testing!

PR #60, opened 2015-11-02
Shift mouse position by half a pixel

gagern:
Compared to the drawing coordinate system, where integer coordinates represent corners of pixels, the input coordinate system is shifted by half a pixel so that integer coordinates represent centers of pixels.  Here we make this shift explicit, transforming input coordinates to the drawing coordinate system.

The need for this modification was first discussed in #31. This complements the shift in origin position introduced in #59.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -114,8 +114,8 @@ function setuplisteners(canvas, data) {
 
     function updatePostition(event) {
         var rect = canvas.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


strobelm:
Since we merged #59 this could be merged too, right?

gagern:
Merging this one was the correct thing to do even before #59 landed, but if you merge it, too, we can close #31.

PR #63, opened 2015-11-02
Format angles as degrees by default

### LINKED ISSUE #17: angles don't respect that they are angles when printed
*(Manually migrated from GitLab issue 4 reported by @kortenkamp)*

In Cindy Classic, each complex value has a "usage", for example to mark that it "is" an angle. This is respected when the complex value is formatted for printing or drawing on screen. An angle is usually drawn using the 0°-360° notation. In example [Timos Winkel](https://github.com/CindyJS/CindyJS/blob/1f34b34b7d5305bbd44168b8df3cb189f532c5c4/examples/100_ScreenResolution.html) you can see that this is currently not the case in Cindy JS, leading to ugly output.

Of course it is possible to work around this, but it would be better to show the default behavior here.

Cinderella screenshot
CindyJS screenshot

-----

*(Comment originally by @gagern)*

Agreed. We do have a usage in CindyJS as well. We are even [using it](https://github.com/CindyJS/CindyJS/blob/560eead3d8739cc08d0ac60c17a552ed16d4b1e0/src/js/libcs/CSNumber.js#L326) for some inverse trigonometric functions. So it's mostly the printing which we are lacking, and that mostly because we don't have anything corresponding to the angle format setting Cinderella offers the user.

By the way, is that setting only stored with the user settings, or also within the file? Would a file created by a user using radians and opened by a user using degrees show degrees where it used to show radians? I assume so.

-----

*(Comment originally by @kortenkamp)*

In my opinion the setting should be stored in the file, but it is not, as it is not a per-file (or per-Dispatcher) setting, but a global user preference, like language.

So we should find a good way to solve it. One idea would be to make it a per-file setting (not too difficult), but then again a user might want to have a different setting for each angle he displays... Let's also ask @richter – what do you think? Where should be the setting for the default angle display? Is it a user setting (stored in `cinderella2-user.properties`), a file setting (stored in `*.cdy`/`*.js`), or an individual one (which is difficult to do)?

gagern:
Now we're making use of the `usage: "Angle"` setting which might have been applied to some numbers, and format the resulting numbers using degrees instead of radians by default.  There is an instance argument called `angleUnit` which might be used to override that default.  Various units are already defined, using the turn as the unit of reference.

For future extensions, it's already possible to enter a special formatting function into that list.  Such a function could conceivably be used to e.g. format a number as degrees, minutes and seconds.  That isn't supported yet, though, at the time of this writing. Depending on how long the review takes on this, I might add additional commits to take care of that, but if I do I'd also like to extend the testing framework so we can actually unit-test such things.

This fixes #17.

### DIFF
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -121,9 +121,18 @@ For noninteger values of `b` only one principal value of `a^b` will be returned.
 
 This operator multiplies any number by the constant `pi/180` .
 This makes possible angle conversion from degrees to radians.
+Note that angles will be printed in degrees by default.
 
     > 180°
+    < 180°
+    > 180° + 0
     < 3.1416
+    > 180° / 3
+    < 60°
+    > 0.5 * 180°
+    < 90°
+    > 20° + 30°
+    < 50°
     > cos(180°)
     < -1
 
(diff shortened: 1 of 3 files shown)


strobelm:
Works nicely, thanks a lot!

PR #68, opened 2015-11-04
Figure out script base directory in a broader range of use cases

gagern:
Now we no longer rely on the script being named `Cindy.js`.  Instead we use [`document.currentScript`](https://html.spec.whatwg.org/multipage/dom.html#dom-document-currentscript) if available, and fall back to the last script added to the DOM if that isn't available and we can't get a match by script name either.  Since content exported from Cinderella currently refers to a file called `Cindy-dev.js`, there is some need for flexibility here.

Using the last script was suggested in [http://stackoverflow.com/a/2161748](http://stackoverflow.com/a/2161748/1468366) by StackOverflow user [bobince](http://stackoverflow.com/users/18936/bobince).

Currently the [`katex1.html` example](https://github.com/CindyJS/CindyJS/blob/b99d31a78ead7837f0c0365cf9bd39c803d6ca54/examples/katex1.html) relies on this machinery to locate the current script. It would be particularly interesting to test whether that example works on Internet Explorer, both if the script file is called `Cindy.js` and something else.

### DIFF
--- a/src/js/Head.js
+++ b/src/js/Head.js
@@ -52,26 +52,52 @@ var createCindy = (function() {
 
             var baseDir = null;
             var cindyJsScriptElement = null;
-            var waitingForLoad = {};
+            var lastScriptAtLoadTime = null;
+
+            // The following code must be executed immediately
+            // when the script gets loaded, not from some callback.
+            if (typeof document !== "object") {
+                cindyJsScriptElement = false;
+            } else if (document.currentScript) {
+                cindyJsScriptElement = document.currentScript;
+            } else {
+                // Assuming synchroneous loading, our script should be
+                // the last one in the DOM.
+                var scriptsAtLoadTime = document.getElementsByTagName("script");
+                lastScriptAtLoadTime =
+                    scriptsAtLoadTime[scriptsAtLoadTime.length - 1];
+            }
 
             createCindy.getBaseDir = function() {
                 if (baseDir !== null)
                     return baseDir;
-                var scripts = document.getElementsByTagName("script");
-                for (var i = 0; i < scripts.length; ++i) {
-                    var script = scripts[i];
-                    var src = script.src;
-                    if (!src) continue;
-                    var match = /Cindy\.js$/.exec(src);
-                    if (match) {
-                        baseDir = src.substr(0, match.index);
-                        console.log("Will load extensions from " + baseDir);
-                        cindyJsScriptElement = script;
-                        return baseDir;
+                if (cindyJsScriptElement === null) {
+                    var scripts = document.getElementsByTagName("script");
+                    for (var i = 0; i < scripts.length; ++i) {
+                        var script = scripts[i];
+                        var src = script.src;
+                        if (!src) continue;
+                        var match = /Cindy\.js$/.exec(src);
+                        if (match) {
+                            baseDir = src.substr(0, match.index);
+                            console.log("Found Cindy.js script at " + src);
+                            console.log("Will load extensions from " + baseDir);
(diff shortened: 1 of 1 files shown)


gagern:
Tested successfully on Internet Explorer by @wursterje.

Now that I think about it, I'm not so sure that this change is a good thing for the master branch. Before this change, there was exactly one way to figure out the current script. So if a given page worked in one browser, it would work in any browser. Now we have three different ways to figure out the script, namely using `document.currentScript`, by name matching and using the last loaded script in the DOM at initialization. Different browsers will choose different approaches for different pages. So I as a content developer can no longer rely that something which works for me will work for others as well.

It *might* be better to not merge this change to master, but instead take the stance that any deployment of CindyJS should please name the file so that it ends in `Cindy.js`. If different versions should be distinguished, that can be achieved using different paths, e.g. `…/dev/Cindy.js`. If a hierarchical path isn't possible either, one can still use a prefix, like `…/dev-Cindy.js`, in which case related resources would share that prefix, e.g. `…/dev-katex-plugin.js`. The format `Cindy-dev.js` currently used by Cinderella would be officially deprecated and would not support automatic loading of additional resources.

Since I am unsure, please share your thoughts on this.

kortenkamp:
But you could not to be sure that something that works for you works for others in other browsers before that, could you?

strobelm:
The more i think about it my opinion is rather not to merge. We are planing to deploy on many platforms (Win, Linux, *BSD, iOS, Android, Atari TOS) and the benefit of the code does probably not counterbalance the cost of testing every platform.

gagern:
@kortenkamp: The old code had no browser-dependent case distinctions as far as I'm aware. All browsers inspect the set of all script tags, looking for a match against the same RegExp.

The problem was more along the lines that some content which worked in one setup (like the katex example on my local system) might fail in another context due to different naming conventions. By not merging this we'd declare that in such situations, the one who set up the context is to blame. We could trigger the `getBaseDir` call unconditionally upon initialization, so that these users would see an appropriate warning in the web console even if the example they were using for testing didn't require any external content.

kortenkamp:
@gagern Even if the code has no browser-dependent case distinctions it can be browser-dependent... 

I think that, even if we are talking about a library, we should take away the responsibility for such things from the users of the library. Following the general design idea of CindyScript: don't require that users can remember exact syntax ;-)

gagern:
@kortenkamp: I'm not sure I can follow you. What kind of browser-dependence do you have in mind here? Where do you see the analogy to remembering syntax?

In my opinion the only thing CindyJS needs is clear install instructions. Something like

> 1. Download https://github.com/CindyJS/deploy/archive/snapshot.zip
> 2. Unzip it to a directory of your choice in your web document tree
> 3. If you don't like the name of the extracted top level directory (`deploy-snapshot` in this case), rename it or merge its content into some existing directory
> 4. Reference the `Cindy.js` file in that directory from the HTML pages containing CindyJS content
>
> As an alternative, don't download anything but reference https://raw.githubusercontent.com/CindyJS/deploy/snapshot/Cindy.js directly

Of course, once we have releases the documentation should be updated to not refer to snapshots any more. once we have such documentation, a normal user would automatically use the name `Cindy.js` and be done with it. And anyone else who decides to tinker with names will be on their own.

gagern:
Since Cinderella merge request 34 the URL used for exports is http://cinderella.de/CindyJS/latest/Cindy.js. Since we have a proper deployment tree at http://cinderella.de/CindyJS/latest/ we don't need this hack here in CindyJS master.

PR #69, opened 2015-11-04
Clean up examples

gagern:
* Change script type from `cindyscript` or `text/cindyscript` to `text/x-cindyscript`
* Rename firstDrawing to csmove since that's what it usually is
* Add some test in Makefile to detect if the above appear again
* Remove trailing whitespace from all lines in the HTML files
* Remove examples which look dead and refer to firstDrawing


strobelm:
Hi i looked over the examples and all of them seemed to work (at least at first glance). So thanks a lot @gagern (and sed, awk i think ;-) )

gagern:
No awk, but quite a lot of sed. Thanks for testing!

PR #72, opened 2015-11-10
Make line widths more consistent

### LINKED ISSUE #50: Inconsistent default line widths
We have two different default line widths.

One is defined using `csport.drawingstate.linesize` and [defaults to 1.0](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libgeo/GeoState.js#L13). It ends up in the [`Render2D.lsize`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/Render2D.js#L46) setting. This is used for segment drawing. Other pieces of code set `csctx.lineWidth` directly, namely

* [drawarc defaults to `4/2.5`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L170-L175),
* [drawcircle defaults to `4/2.5`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L318-L321),
* [drawconic defaults to `4/2.5`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L405-L410) ([reused](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L594) in nested function),
* [polygon defaults to `4/2.5`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L819-L822),
* [plot uses `Render2D.lsize`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L1058) so it's consistent with segments,
* [plotX hardcoded to 1](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#L1265).

I'm not sure about `plotX` (perhaps this should even be removed?), but the difference between `1` and `4/2.5=1.6` is confusing, as is the fact that the latter isn't affected by the [`defaultAppearance`](https://github.com/CindyJS/CindyJS/blob/57a564d89124df6d86f605272bc91740fb2c2348/src/js/libcs/OpDrawing.js#user-content-defaultappearance) option to the `createCindy` call. This should be unified.

gagern:
Now the default line width is 1 for geometric elements, for straight lines drawn from scripts, and for conics and polygons drawn from scripts. The width for geometric elements can be set by passing `defaultAppearance: { lineWidth: 2 }` or similar to `createCindy`, while the width for scripted elements is affected by a call to `linesize(2)` or similar from within the script. Individual geometric elements still can have their own `size: 2` property, and draw instructions their `size->2` modifier.

The newly introduced `Render2.preDrawCurve` serves as a common place from where the drawing context can be prepared for a subsequent drawing operation.  `plot` doesn't call this since it uses a different line join style.  Fill operations have to set the `fillStyle` independently, too.

This fixes #50. Here is a comparison of the new example before and after the fix.

before

after

As you can see, the alpha value of the arc looks wrong, but that's due to how we have to draw arcs due to a bug in Chrome. On the whole, the line widths in each of the figures are now more consistent, and the two left figures are consistent with one another as well.


strobelm:
Hi! Look good, thanks a lot especially for tidying up my code!

Issue #24, opened 2015-10-19
Support all algorithms Cinderella supports


gagern:
The following is an automatically generated (but manually edited) list of all the algorithms Cinderella (as of build 1786) appears to understand when loading files. There might be false positives, i.e. things which are not really intended as stand-alone algorithms. There might be false negatives, i.e. things which do not appear in the list despite the fact that they are in fact different algorithms, like the different versions of `PointOnLine`. Checked algorithms are those for which we have a matching entry in `geoOps` or `geoMacros`, so that presumably these are supported. Please tick additional algorithms once their implementation gets merged into master.

* [ ] Angle
* [ ] AngleMark
* [ ] AngularBisector
* [ ] AnimationAlg
* [ ] Arc
* [ ] Area
* [ ] Basis
* [ ] Calculation
* [ ] Center
* [ ] CindyScriptElement
* [X] CircleBy3
* [ ] CircleByFixedRadius
* [X] CircleByRadius
* [ ] CircleHelperForPolygons
* [X] CircleMP
* [X] Compass
* [ ] CompassByRadius
* [ ] ComplexAdd
* [ ] ComplexMult
* [ ] Conic1P4L
* [ ] Conic4P1L
* [ ] Conic5L
* [X] ConicBy5
* [ ] ConicFoci
* [ ] ConicFociH
* [ ] ConicInSquare
* [ ] ConicParabolaPL
* [ ] ConicPrincipleDirs
* [ ] Dist
* [ ] EditableText
* [ ] EuclideanMid
* [ ] Evolute
* [ ] ExtractTransform
* [X] FreeLine
* [X] FreePoint (mapped to `Free` on export)
* [ ] HorizontalLine
* [ ] IFS
* [X] IntersectionCircleCircle
* [ ] IntersectionConicConic
* [X] IntersectionConicLine
* [X] Join
* [ ] LabelText
* [ ] LineByFixedAngle
* [ ] Locus
* [ ] LocusConic
* [ ] LocusFlat
* [ ] LocusInterval
* [ ] LocusLine
* [ ] MTransform
* [X] Meet
* [X] Mid
* [X] Orthogonal (mapped to `Perp` on export)
* [X] OtherIntersectionCC
* [X] OtherIntersectionCL
* [X] Parallel (mapped to `Para` on export)
* [ ] PointOnArc
* [X] PointOnCircle
* [ ] PointOnCircleBasis
* [X] PointOnLine with `type == NEAREST`
[...]

PR #73, opened 2015-11-11
implemented arcPPP in gslp

strobelm:
Hi! I quickly implemented arcPPP in gslp. To be consistent with naming i called it ArcBy3, any objections?

There is an example: in examples/104c_arc_gslp.html.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1012,6 +1012,10 @@ geoOps.CircleBy3.updatePosition = function(el) {
 
 };
 
+geoOps.ArcBy3 = {};
+geoOps.ArcBy3.kind = "C";
+geoOps.ArcBy3.updatePosition = geoOps.CircleBy3.updatePosition;
+
 geoOps.Polar = {};
 geoOps.Polar.kind = "L";
 geoOps.Polar.updatePosition = function(el) {
--- a/src/js/libgeo/GeoRender.js
+++ b/src/js/libgeo/GeoRender.js
@@ -40,6 +40,25 @@ function render() {
         }
     }
 
+    function drawgeoarc(el) {
+        if (!el.isshowing || el.visible === false)
+            return;
+
+        var modifs = {};
+        modifs.color = el.color;
+        modifs.alpha = el.alpha;
+        modifs.size = el.size;
+
+        // check if we have filled: true
+        var df = el.filled ? "F" : "D";
+
+        var newargs = [csgeo.csnames[el.args[0]].homog,
+            csgeo.csnames[el.args[1]].homog,
+            csgeo.csnames[el.args[2]].homog
+        ];
+        eval_helper.drawarc(newargs, modifs, df);
+    }
+
 
     function drawgeoconic(el) {
         if (!el.isshowing || el.visible === false)
@@ -138,7 +157,8 @@ function render() {
     var i;
 
     for (i = 0; i < csgeo.conics.length; i++) {
-        drawgeoconic(csgeo.conics[i]);
+        if (csgeo.conics[i].type === "ArcBy3") drawgeoarc(csgeo.conics[i]);
+        else drawgeoconic(csgeo.conics[i]);
     }
 
 
--- a/examples/104c_arc_gslp.html
+++ b/examples/104c_arc_gslp.html
@@ -14,23 +14,25 @@
 
 // Drawing code, executed whenever the canvas gets redrawn.
 
-drawarc(A,B.homog,C.xy);
 </script>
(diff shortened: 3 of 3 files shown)


gagern:
As per #24, the primitive there is just called `Arc`. Although I do concede that in this case the name you picked is the more descriptive one, so I wouldn't mind using that. After all, we might one day have an arc defined by a pair of lines and a point, using the point of intersection as the center and the given point to define the radius and the sector to choose. Or something like that. So I'd say keep the name you chose and adjust Cinderella export.

Make sure that after name adjustment, exporting arcs from Cinderella works. (There is no Toolbar button for arcs, but it's in the menu.) I'll have a closer look at your code now.

src/js/libcs/OpDrawing.js
```diff
@@ -107,18 +107,26 @@ evaluator.fillcircle$2 = function(args, modifs) {
...
 };
 
 evaluator.fillarc$3 = function(args, modifs) {
-    return eval_helper.drawarc(args, modifs, "F");
+    var newargs = [csgeo.csnames[args[0].name].homog,
+        csgeo.csnames[args[1].name].homog,
+        csgeo.csnames[args[2].name].homog
+    ];
+    return eval_helper.drawarc(newargs, modifs, "F");
 };
```
gagern: No good! Here you're assuming that the arguments will be geometric points. Just like it was before 3c550cd2b25f78d28714a5de9b46c152a4932ee4. Here you are at the CindyJS level, and `evaluateAndHomog` is all you need.

src/js/libgeo/GeoRender.js
```diff
@@ -40,6 +40,25 @@ function render() {
...
+
+        // check if we have filled: true
+        var df = el.filled ? "F" : "D";
+
+        var newargs = [csgeo.csnames[el.args[0]].homog,
+            csgeo.csnames[el.args[1]].homog,
+            csgeo.csnames[el.args[2]].homog
+        ];
+        eval_helper.drawarc(newargs, modifs, df);
+    }
```
gagern: Here the conversion from argument names to their coordinates is right. This conversion from `el.args` to `newargs` should be enough to make the other similar changes unneccessary.

src/js/libgeo/GeoOps.js
```diff
@@ -1012,6 +1012,12 @@ geoOps.CircleBy3.updatePosition = function(el) {
 
 };
 
+geoOps.ArcBy3 = {};
+geoOps.ArcBy3.kind = "C";
+geoOps.ArcBy3.updatePosition = function(el) {
+    return geoOps.CircleBy3.updatePosition(el);
+};
+
```
gagern: Just an idea: you could even shorten this to `geoOps.ArcBy3.updatePosition = geoOps.CircleBy3.updatePosition`. Would avoid one level of indirection, which shouldn't be noticeable in my opinion. So the main question is, which form is easier to understand at first glance? Yours fits in better with the pattern of the implementations around, while my suggestion makes it clear that the implementation is *exactly* the same. Pick that if you think the benefits outweight the drawbacks. Given that you came up with this here first, I doubt they do.

src/js/libcs/OpDrawing.js
gagern: To avoid this problem from manifesting itself again, could you please change the existing arc examples to (also?) include some arc drawn using vectors? E.g. make it `drawarc(A, B.homog, C.xy)` or something like that.

gagern:
After renaming, the export from Cinderella works. Haven't tried most attributes, but I guess the'd work if they work for conics.

strobelm:
Hi! Was disconnected for the last few hours, will fix it asap.

Am 11. November 2015 22:52:25 MEZ, schrieb Martin von Gagern <notifications@github.com>:
>After renaming, the export from Cinderella works. Haven't tried most
>attributes, but I guess the'd work if they work for conics.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/CindyJS/CindyJS/pull/73#issuecomment-155921166

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

strobelm:
Hi, i pushed the changes to my branch. Thank @gagern for the comments! 

Concerning geoOps.ArcBy3.updatePosition of course its matter of taste, i'm fine with both. For the sake of shorter code and less function calls i switched to direct assignment.

strobelm:
There is none, i simply overlooked that. Any further remarks?

gagern:
Now the diff looks really nice and clean. I'll have to give it one last try before merging, but I can't imagine anything going wrong.

PR #74, opened 2015-11-13
Make drawing of traces more consistent with Cinderella

gagern:
Now we implement the cubic alpha dimming and the dimension change in the same way `CubicAlphaAgingRenderer` does things in Cinderella.

The property in the geometry list was renamed from `trace` to `drawtrace`. That's what Cinderella exports already, and it's more descriptive and less likely to be confused with the topic of complex tracing around singularities. This name change earned the pull request an “incompatible” lable, though. I hope there is not too much content relying on the old name.

The storage for past trace positions was changed from a list with endpoint modifications to a list of fixed length with cyclical index.  [Tests](http://jsperf.com/shift-register/2) suggests that this should be orders of magnitude better than what we had, and it also avoids trouble due to a varying list length.

### DIFF
--- a/examples/106_DrawTrace1.html
+++ b/examples/106_DrawTrace1.html
@@ -22,7 +22,7 @@ var cdy = createCindy({
     transform: [{visibleRect: [-9.04, 9.32, 18.16, -4.12]}]
   }],
   geometry: [
-    {name: "A", type: "Free", pos: [0, 0], trace: true, tracedim: 1, tracelength: 200, traceskip: 1},
+    {name: "A", type: "Free", pos: [0, 0], drawtrace: true, tracedim: 1, tracelength: 200, traceskip: 1},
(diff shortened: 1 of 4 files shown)


strobelm:
Hi! After beautifying the branch tests are passed. 

The second example won't work until the new version of Cindy.js is pushed to cinderella.de (direct inclusion from export) which works fine if one uses  Cindy.js from the branch.

Therefor merged.

gagern:
Beautify has a strange taste of beauty here. But I should have run that locally. Should try to remember next time, but I'm glad we have Travis to remind me.

And I probably should update the URL from the example next time I'm editing that. Was pure luck that I had tested the example in a local setup which replaces remote versions of Cindy.js with a local development copy via proxies, so I didn't even notice the possible problem.

PR #75, opened 2015-11-16
Restore drawing of clipped lines

gagern:
While we use startpos and endpos to draw segments, we stick to using the defining points when drawing a line clipped to its defining points.

I broke this in ee4954cdd4ac278c12778119ca187c5df585e474 but since 47f4826ca74d38eab7fa404b1ee6c6f2cd338a25 moved code between files since then, there is little point in basing the fix on the former.

### DIFF
--- a/src/js/libgeo/GeoRender.js
+++ b/src/js/libgeo/GeoRender.js
@@ -80,20 +80,30 @@ function render() {
         if (!el.isshowing || el.visible === false || !List._helper.isAlmostReal(el.homog))
             return;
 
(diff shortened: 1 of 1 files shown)


gagern:
Segment drawing doesn't seem to work as intended yet. Investigating…

gagern:
False alarm: the segment drawing issues were related to overhang handling, which isn't part of this pull request yet.

strobelm:
Code looks fine to me and the examples i tested are working.

PR #76, opened 2015-11-16
Support more geometric attributes

gagern:
Support default font family, overhang and dashtype. This is intended to make Cinderella exports more consistent with the appearance in Cinderella itself. A matching pull request againt Cinderella will be filed as well.

### DIFF
--- a/examples/106_DrawTrace2.html
+++ b/examples/106_DrawTrace2.html
@@ -19,7 +19,7 @@
             width: 100%;
         }
     </style>
-    <script type="text/javascript" src="http://cinderella.de/CJS/build/js/Cindy-dev.js"></script>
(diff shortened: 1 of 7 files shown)


strobelm:
Interesting that clicking on the canvas makes the dashes move, while Events.js claims that there is nothing to move. @gagern is this intended?

gagern:
@strobelm No, definitely not intended behavior. I guess the code is modifying the pattern array instead of creating a new one. Will investigate. Thanks for finding this!

strobelm:
Good old arr.slice(), thanks @gagern for implementing this, the code looks fine now! 

Just one proposal: if you use a exported example next time it would be great to adjust the Cindy.js path. I was looking around for some minutes to find that the path is linked to cinderella.de and hence the dashing wasn't working at all ;-).

PR #77, opened 2015-11-19
Implemented x and y Accessors 

strobelm:
Hi, i implemented the x,y accessors: it's now possible to set A.x = 5 and A.y = 3. 

I wanted to quickly debate the following: do we want list assignment for A.x and A.y? E.g. A.x = [1] and A.y = [2]? I'm not sure if there is a usecase (and Cinderella ignores the assignment), if not i'll remove that from the code.

Furthermore how do we handle infinity, what should be A.x of A.homog = [1,2,0]? Currently we return NaN.

Last but not least i fixed an jshint error in Operators.js.

### DIFF
--- a/src/js/libcs/Accessors.js
+++ b/src/js/libcs/Accessors.js
@@ -168,6 +168,15 @@ Accessor.setField = function(geo, field, value) {
         movepointscr(geo, List.turnIntoCSList([value.value[0], value.value[1], CSNumber.real(1)]), "homog");
     }
 
+    if (field === "x" && geo.kind === "P" && geo.movable && value.ctype === "number") {
+        movepointscr(geo, List.turnIntoCSList([CSNumber.mult(value, geo.homog.value[2]), geo.homog.value[1], geo.homog.value[2]]), "homog");
+    }
+
+    if (field === "y" && geo.kind === "P" && geo.movable && value.ctype === "number") {
+        movepointscr(geo, List.turnIntoCSList([geo.homog.value[0], CSNumber.mult(value, geo.homog.value[2]), geo.homog.value[2]]), "homog");
+    }
+
+
     if (field === "homog" && geo.kind === "P" && geo.movable && List._helper.isNumberVecN(value, 3)) {
         movepointscr(geo, value, "homog");
     }
--- a/examples/109_xyAccessors.html
+++ b/examples/109_xyAccessors.html
@@ -1,0 +1,71 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+// test for far points
+D.homog = [2,2,0];
+D.x = 2;
+D.y = 1;
+println(D.homog);
+</script>
+<script id="csdraw" type="text/x-cindyscript">
+// test homogs
+C.homog = [2,2,2];
+C.x = 2;
+println(C.homog);
+
+D.homog = [2,2,2];
+D.y = 2;
+println(D.homog);
+
+// test gslp points
+A.x = 2;
+B.y = -5;
+println(A.xy);
+println(B.xy);
+
+// invalid assignments
+Ax = A.x;
+A.x = "aa";
+println(Ax == A.x);
+By = B.y;
+B.y = [2,3];
+println(B.y == By);
+
+
+
+</script>
+<script type="text/javascript">
+
+var cdy = createCindy({ // See ref/createCindy documentation for details.
+  ports: [{id: "CSCanvas"}],
+  scripts: "cs*",
+  language: "en",
+  defaultAppearance: {
+    // See GeoBasics.js for possible attributes.
+  },
+  geometry: [
+    {name:"A", type:"Free", pos:[0,0]},
+    {name:"B", type:"Free", pos:[0,0]},
+    {name:"C", type:"Free", pos:[0,0]},
+    {name:"D", type:"Free", pos:[0,0]}
+    // For allowed types, see GeoOps.js.
+    // For allowed properties, see csinit in GeoBasics.js.
+  ] // End of geometry array.
+});
+
+// Remove all comments after adjusting this template for your use case.
+
+</script>
+</head>
+
+<body style="font-family:Arial;">
+  <canvas id="CSCanvas" width="500" height="500"
+          style="border:2px solid black"></canvas>
+</body>
+
+</html>
--- a/src/js/libcs/CSNumber.js
+++ b/src/js/libcs/CSNumber.js
@@ -589,6 +589,10 @@ CSNumber._helper.isNaN = function(a) {
     return (isNaN(a.value.real)) || (isNaN(a.value.imag));
 };
 
+CSNumber._helper.isFinite = function(z) {
+    return isFinite(z.value.real) && isFinite(z.value.imag);
+};
+
 
 CSNumber._helper.isAlmostImag = function(a) {
     var r = a.value.real;
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -543,6 +543,7 @@ List.maxIndex = function(lst, fun, startIdx) {
 
 List.normalizeMax = function(a) {
     var s = CSNumber.inv(List.maxval(a));
+    if (!CSNumber._helper.isFinite(s)) return a;
     return List.scalmult(s, a);
 };
 


src/js/libcs/Accessors.js
```diff
@@ -168,6 +168,24 @@ Accessor.setField = function(geo, field, value) {
         movepointscr(geo, List.turnIntoCSList([value.value[0], value.value[1], CSNumber.real(1)]), "homog");
     }
 
+    if (field === "x" && geo.kind === "P" && geo.movable && (List._helper.isNumberVecN(value, 1) || value.ctype === "number")) {
+        if (CSNumber._helper.isAlmostZero(geo.homog.value[2])) {
+            console.log("cannot set x component without y component of far points");
```
kranich: Why should we not be allowed to set x to a non-zero value if y and z are (almost) zero?

src/js/libcs/Accessors.js
```diff
@@ -168,6 +168,24 @@ Accessor.setField = function(geo, field, value) {
...
+        } else {
+            var xval = (value.ctype === "list") ? value.value[0] : value;
+            movepointscr(geo, List.turnIntoCSList([xval, CSNumber.div(geo.homog.value[1], geo.homog.value[2]), CSNumber.real(1)]), "homog");
+        }
+    }
+
+    if (field === "y" && geo.kind === "P" && geo.movable && (List._helper.isNumberVecN(value, 1) || value.ctype === "number")) {
+        if (CSNumber._helper.isAlmostZero(geo.homog.value[2])) {
+            console.log("cannot set y component without x component of far points");
```
kranich: Same as above.

kranich:
> I wanted to quickly debate the following: do we want list assignment for A.x and A.y? E.g. A.x = [1] and A.y = [2]? I'm not sure if there is a usecase (and Cinderella ignores the assignment), if not i'll remove that from the code.

I do not see a use case.

> Furthermore how do we handle infinity, what should be A.x of A.homog = [1,2,0]? Currently we return NaN.

This should be `Infinity`, imho.

strobelm:
Well since A.x is a very euclidean concept (at least a i understand it) applied to something what happens on the line at infity. Cinderella returns Infinity or NaN (interestingly).

kranich:
>Furthermore how do we handle infinity, what should be A.x of A.homog = [1,2,0]? Currently we return NaN.

More precisely, just dehomogenize.

`1/0` yields `Infinity`, as does `2/0`. So `A.x` and `A.y` should be `Infinity`.

For `A.homog = [1,0,0]`, `A.y` is `0/0` and thus `NaN`.

src/js/libcs/Accessors.js
kranich: By the way, do we `normalizeMax` on assignment? This would make change of representant impossible. (I only see pathological use cases for change of representant, anyway.)

kranich: Cinderella treats assignment of `A.x` in an interesting way:

```
A.homog=[1,2,0];
A.x = 2;
err (A.homog);
```

yields

`A.homog ==> [4,0,0]`

Maybe @richter or @kortenkamp can comment on the issue?

strobelm:
I talked to @richter-gebert and he thinks it's the best way to not assign x and y values separately to points at infinity. The behavior of Cinderella is odd indeed.

The code for list assignment was removed with the last commit. 
When we talk about the return value of A.x we should perhaps rethink the behavior of CSNumber.div(a,b) if b is zero and change it to Infinity.

kranich:
>When we talk about the return value of A.x we should perhaps rethink the behavior of CSNumber.div(a,b) if b is zero and change it to Infinity.

Only if `b` is exactly zero and unless `a` is also zero. If both `a` and `b` are exactly zero, we should return `NaN`.

strobelm:
+1 (the social media thing, not the arithmetic operation)

gagern:
> we should perhaps rethink the behavior of CSNumber.div(a,b) if b is zero and change it to Infinity

Apart from mathematical correctness when manually looking at a printed value, what would be the main benefit of returning `Infinity + i*Infinity` instead of `Infinity + i*NaN`? Right now the main difference that comes to my mind would be that one could use 1/∞=0. In other words, we could actually use the result to continue the computation *if* we also had a special case for `1/(± Infinity ± i*Infinity)=0`. It might make sense to add these two special cases together, along with a test case demonstrating that we can indeed continue with the result as described.

I'll have a look at the core code change tomorrow.

strobelm:
On the other hand we should keep an eye on performance, when we clutter CSNumber.div with a lot of ifs and elses we will most likely see a performance breakdown in the linear algebra routines. I'm not sure whether it is worth the price.

gagern:
@strobelm Agreed. Do we have some good test case to check the performance impact of such a modification? If not it might be worthwhile to write one.

kortenkamp:
I don't think that `Infinity` is better (or even more correct) than `NaN`.

strobelm:
just hacked a quick benchmark: https://paste.debian.net/hidden/ce31d4dc/

it looks like that the checked version is around 15-30% slower (for sure my code is not optimized)

src/js/libcs/Accessors.js
```diff
@@ -168,6 +168,22 @@ Accessor.setField = function(geo, field, value) {
         movepointscr(geo, List.turnIntoCSList([value.value[0], value.value[1], CSNumber.real(1)]), "homog");
     }
 
+    if (field === "x" && geo.kind === "P" && geo.movable && value.ctype === "number") {
+        if (CSNumber._helper.isAlmostZero(geo.homog.value[2])) {
+            console.log("cannot set x component without y component of far points");
+        } else {
+            movepointscr(geo, List.turnIntoCSList([value, CSNumber.div(geo.homog.value[1], geo.homog.value[2]), CSNumber.real(1)]), "homog");
+        }
+    }
```
gagern: I've got an alternative suggestion: instead of `[value, y/z, 1]` you might use `[value*z, y, z]` as the new coordinate vector. That would allow you to drop the case distinction since you wouldn't have a division by zero and you would have well-defined (if not exactly geometrically meaningful) behavior even for points at infinity. And you would avoid the costly division, using the cheaper multiplication instead.

kortenkamp: And it is homogenous thinking!

src/js/libcs/Accessors.js
```diff
@@ -168,6 +168,22 @@ Accessor.setField = function(geo, field, value) {
         movepointscr(geo, List.turnIntoCSList([value.value[0], value.value[1], CSNumber.real(1)]), "homog");
     }
 
+    if (field === "x" && geo.kind === "P" && geo.movable && value.ctype === "number") {
+        if (CSNumber._helper.isAlmostZero(geo.homog.value[2])) {
+            console.log("cannot set x component without y component of far points");
```
kortenkamp: It is CindyScript philosophy (at least I thought so) to silently discard such things. Should you really log this?

src/js/libcs/Accessors.js
kranich: Beware that if `y` and `z` are zero, we assign `[0, 0, 0]` as homogeneous coordinates. Is this OK?

kortenkamp: But in the original code you'd assign `[x, NaN, 1]` which is far from being better. `[0,0,0]` is `NaHC` (not the chemical formula, but short for `Not a Homogeneous Coordinate`)

kranich: @kortenkamp In the original code, the case distinction precludes assignment of `[x, NaN, 1]`.

If we drop the case distinction, we allow the user to assign `NaHC` as a homogeneous coordinate of a point. Otherwise, such an assignment is not possible (at least using this method).

I think simply ignoring a requested assignment may have at least as unexpected consequences as assigning `NaHC`. In this case, the warning may be justified. (We also still warn about division by zero in `CSNumber.inv`.)

An assignment of `NaHC` can be detected by users without a log message. This seems much cleaner. (If users *want* to shoot themselves in the foot by assigning xy-coordinates of far points, why should we not let them?)

I am in favour of dropping the case distinction as @gagern suggested.

gagern:
I think assigning `[0,0,0]` is the right thing to do in those cases where it happens. A null vector is the common result wherever we encounter some geometrically undefined operation, and this should be no exception. So I'd still drop the case distinction completely.

I also thought about the following situation: take `[3,1,0]` as an example. Thats a point at infinity and not on one of the coordinate axes. So conceptually both its *x* and its *y* coordinate would be infinite in magnitude, even if “*x* is three times larger than *y*”. Now if you assign a finite value to e.g. *x*, what you get is a point with still infinite *y* value and an *x* coordinate which is infinitely smaller than that. `[0,1,0]` would be the correct result here, and that's *exactly* what you get from the multiplicative approach.

strobelm:
Okay, If we are sure that there are no unwanted side effects i'll remove the check.

strobelm:
Okay now we get a lot of more NaNs since geoOps.Free.getParamForInput will do a normalizeMax the moved points. 
What about stopping NormalizeMax to normalize by 0? Probably a good idea in many situations.

gagern:
Thinking about all the possible special cases which standard floating point arithmetic can throw at us, I think it might make sense to introduce some new functions in the `CSNumber` namespace:

```js
CSNumber.isFinite = function(z) { return isFinite(z.value.real) && isFinite(z.value.imag); };
CSNumber.isNaN = function(z) { return isNaN(z.value.real) || isNaN(z.value.imag); };
CSNumber.isStrongInfinity = function(z) { // real and imaginary part are both infinite
  var r = z.value.real, i = z.value.imag;
  return !isFinite(r) && !isFinite(i) && !isNaN(r) && !isNaN(i);
};
CSNumber.isWeakInfinity = function(z) { // one of real and imaginary part is infinite
  var r = z.value.real, i = z.value.imag;
  return (!isFinite(r) && !isNaN(r)) || (!isFinite(i) && !isNaN(i));
};
```

The `NaN` test we already have in [`CSNumber._helper.isNaN`](https://github.com/CindyJS/CindyJS/blob/051f80c46ce61aef084de0ef255a493bdf428efe/src/js/libcs/CSNumber.js#L588-L590), so it would be enough to rename that. And #57 currently proposes to add [`CSNumber._helper.isFinite`](https://github.com/gagern/CindyJS/blob/3ec0a9d1f5514f76a9bad1509c234499238ef8b9/src/js/libcs/CSNumber.js#L557-L559).

By the way, is there any clear distinction between `CSNumber` itself and `CSNumber._helper`? Theoretically I'd assume that the helpers should only be called from within `CSNumber`, and consitute some API which is considered private to that file. But that's not the way it's being used; we have calls to `CSNumber._helper.isAlmostReal` all over the place. Should I write a pull request which simply drops that intermediate `_helper` element?

Back to your question:

> What about stopping NormalizeMax to normalize by 0? Probably a good idea in many situations.

I agree. How about we change [`List.normalizeMax`](https://github.com/CindyJS/CindyJS/blob/051f80c46ce61aef084de0ef255a493bdf428efe/src/js/libcs/List.js#L544-L547) like this:

```diff
 List.normalizeMax = function(a) {
     var s = CSNumber.inv(List.maxval(a));
+    if (!CSNumber.isFinite(s)) return a;
     return List.scalmult(s, a);
 };
```

That would take care of the division by zero. It would still leave us with a null vector if one of the input coordinates happens to be infinite. I haven't decided yet whether that's desirable or not. Do you have an opinion on this?

Other normalization functions could be handled in the same fashion. We might want to be careful, though, since there exists code which relies on certain properties of the normalized result, and won't check for that. For this reason I'd rather not change more than `normalizeMax` just now.

gagern:
One more thing regarding the original core of this request: what's the indended behavior for semi-free elements? In other words, what should happen if you have a point on a line with some certain slope, and set the *x* coordinate of that point? With your suggestion here, the new position of the point would be off the line, and the actual position after the update would be on the line but not with the given *x* coordinate. That's what Cinderella does, too.

However, I'd prefer it if setting a single coordinate of a point on a line would set that coordinate *exactly*, and cause the other coordinate to be chosen to match. Which essentially means taking your line, intersecting it with a vertical line at the given *x* position, and taking the point of intersection as the new homogeneous coordinates. Things would be more difficult for other semi-free elements like point on circle, since there are two alternatives to choose from. Do we want to have something like this, or is compatibility with Cinderella too important to even consider this? If we do that, we probably should do so in a separate issue or pull request, so that we can get this one here in before dealing with all the ugly details of semi-free elements.

strobelm:
@gagern semi free elements are a good point and definitely needs further discussion. I would open an separate issue on that since the fields we're dealing with here are already very vast.

The same holds for CSNumber, which could really use a some refactoring at some time.

I like the idea of isFinite in normalizeMax. So can we agree on altering normalizeMax and the current code for a merge or are there further wishes? I would boil down the merge to the essence and discuss the further changes in separate places (especially since the benevolent dictator for life would like to have the x/y operators).

gagern:
I think yes, current code plus a change to normalizeMax, probably minus the jshint stuff so we can keep that separate. Squashed to a single commit preferably, and as usual include some test case as well. It might be good to have both a HTML example and a unit test in `ref/Interaction_with_Geometry.md`. The latter requires unit tests with gslp content, or the `create` functionality. I'll open a ticket for the former, and am still busy reviewing and polishing the latter.

strobelm:
Ok here we go again. 

I named the isFinite function CSNumber._helper.isFinite to be consistent with the current scheme, if we decide to refactor CSNumber this should not be a huge burden. We should check for Infinite, NaNs and almost 0 in List.normalizeMax and therefor there is an additional call of CSNumber._helper.isAlmostZero.

gagern:
Looks pretty clean now.

I'm not totally convinced by the `isAlmostZero` case. We are talking about the inverse of the maximal value here. So if that is *almost* zero, it means that some entry of the vector is *huge*. What's wrong with normalizing that to `1`? We know that even though the scale factor is almost zero, the final result is not.

strobelm:
Hi! I changed the behavior of the operations a little bit (for
performance reasons):

+    var s = List.maxval(a);
-    return List.scalmult(s, a);
+    if (!CSNumber._helper.isFinite(s) ||
CSNumber._helper.isAlmostZero(s)) return a;
+    return List.scalmult(CSNumber.inv(s), a);

First we get the max value of a, if this one is NaN, Infinite or (almost
0) we return a, this happens before the inversion which is now in the
last line. This means if the number is huge (but not Infinite) we're
fine and the inversion will take place.

But i think i got your point (indirectly): we actually want division by
Infinity, don't we?

Am 25.11.2015 um 17:09 schrieb Martin von Gagern:
>
> Looks pretty clean now.
>
> I'm not totally convinced by the |isAlmostZero| case. We are talking
> about the inverse of the maximal value here. So if that is /almost/
> zero, it means that some entry of the vector is /huge/. What's wrong
> with normalizing that to |1|? We know that even though the scale
> factor is almost zero, the final result is not.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/77#issuecomment-159656938>.
>

-- 
Michael Strobel, M.Sc.

strobelm:
I did some rudimentary performance tests and since direct inversion + isFinite slightly faster (and more readable) i altered rearranged the code again (hopefully for the last time).

gagern:
I like it.

Regarding your previous attempt: I always think it's better to *first* perform the inversion and *then* check whether the result is finite, instead of *first* checking whether the input is close enough to zero that it *might* overflow, and *then* perform the inversion and hope for the best.

PR #81, opened 2015-11-19
Remind developer to adjust URL after export from Cinderella

gagern:
Since I've now for the second time committed an example exported from Cinderella which was still referring to the Cindy.js file on the Cinderella webserver, I feel that I might need some technical help to avoid this in the future.


strobelm:
It's always good to have a helping hand ... or makefile. Thanks Martin!

PR #83, opened 2015-11-20
Better handling of different node versions

gagern:
This should fix the problem of ancient npm versions installing prerelease versions of some modules, as discussed in https://github.com/strobelm/CindyJS/commit/118f4e30216f1bcd16b72d55e21ab423aae1e6ec#commitcomment-14525889. It also makes switching node versions easier which may be useful if at some point in the future 0.12 isn't good enough any more.

@strobelm, please test this. Make sure to clean the `build/node_modules.stamp` file to trigger a re-installation of all node modules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,9 +68,9 @@ download/node/bin/npm: download/arch/$(NODE_TAR)
 	mv download/node-v$(NODE_VERSION)-* download/node
 	touch $@
 
-build/node_modules.stamp: package.json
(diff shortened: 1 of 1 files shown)


strobelm:
Hi! Works great, thanks! I'll investigate whether Fedora will update node/npm to something more modern in the near future.

gagern:
You'll deprive us of these great testing opportunities!?! :stuck_out_tongue_winking_eye:

Issue #78, opened 2015-11-19
Line jumps when the pointer is clicked near it


elkins0:
When you click and release the pointer near the line thru a point the line jumps back and forth across the pointer each time you do this. Doing this near the point A does not do this.

The proposed fix shown below eliminates this behavior.

In https://github.com/CindyJS/CindyJS/tree/master/src/js/Events.js:
```javascript
function getmover(mouse) { ...
        } else if (el.kind === "L") { ...
```
Change lines 50 & 51 to negate the values of dx and dy:

```javascript
            dx = ln.value[0].value.real * dist;
            dy = ln.value[1].value.real * dist;
```

to:

```javascript
            dx = -ln.value[0].value.real * dist;
            dy = -ln.value[1].value.real * dist;
```

gagern:
I can reproduce both your observation and your fix. Looks good to me. Can you create a pull request for this? Since the fix is so simple, I'd suggest you try using the GitHub web interface for this: simply click on the pen icon in the top right corner of the file view, and you should be able to add those minus signs and create a pull request for me to include them.

elkins0:
This issue has been resolved by #87.

PR #87, opened 2015-11-20
Update to Events.js to fix issue #78

elkins0:
This is the proposed fix for issue #78 to avoid lines jumping when the pointer is clicked near to them.


gagern:
Thanks a lot for this fix!

Issue #79, opened 2015-11-19
Behavior of FreeLine, Horizontal and Vertical movable lines


elkins0:
"Export to CindyJS..." from Cinderella can produce references to FreeLine, HorizontalLine and VerticalLine. The HTML produced has to be changed to match CindyJS so that the types "HorizontalLine" and "VerticalLine" became "Horizontal" and "Vertical". Even with those changes "Horizontal" and "Vertical" do no work as they reference a non-existant point argument. And while FreeLine displays, it displays randomly and is not movable.

In https://github.com/CindyJS/CindyJS/tree/master/src/js/libgeo/GeoOps.js,
based on `geoOps.Through` and `geoOps.Free` I propose the following replacements for
`geoOps.FreeLine`, `geoOps.Horizontal` and `geoOps.Vertical`. To avoid having to change Cinderella's "Export to CindyJS..."  `geoOps.Horizontal` and `geoOps.Vertical`
could be renamed `geoOps.HorizontalLine` and `geoOps.VerticalLine`. That decision is up to you.

Also, something strange I noticed during my perusal of the code in GeoOps.js. This is the only geoOps.*.putParamToState that has a return in it:
```javascript
geoOps.PointOnLine.putParamToState = function(el, param) {
    return putStateComplexVector(param);
};
```

### geoOps.FreeLine

```javascript
geoOps.FreeLine = {};
geoOps.FreeLine.kind = "L";
geoOps.FreeLine.isMovable = true;
geoOps.FreeLine.initialize = function (el) {
    var pos = geoOps._helper.initializePoint(el);
    putStateComplexVector(pos);
};
geoOps.FreeLine.getParamForInput = function (el, pos, type) {
    var homog = List.cross(pos, List.ez);
    homog = List.cross(homog, pos);
    return List.normalizeMax(homog);
};
geoOps.FreeLine.getParamFromState = function (el) {
    return getStateComplexVector(3);
};
geoOps.FreeLine.putParamToState = function (el, param) {
    putStateComplexVector(param);
};
geoOps.FreeLine.updatePosition = function (el) {
    var param = getStateComplexVector(3);
    putStateComplexVector(param); // copy param
    el.homog = General.withUsage(param, "Line");
};
geoOps.FreeLine.stateSize = 6;
```

### geoOps.Horizontal

```javascript
geoOps.Horizontal = {};
geoOps.Horizontal.kind = "L";
geoOps.Horizontal.isMovable = true;
[...]

PR #89, opened 2015-11-21
geoOps.FreeLine bug fix for issue #79

elkins0:
This change makes FreeLine behave in response to pointer movement like it does in Cinderella.


elkins0:
Having a space after the function keyword before the open parenthesis causes a mismatch with beautified output

PR #90, opened 2015-11-21
Add freely movable HorizontalLine & VerticalLine

elkins0:
Adds geoOps.HorizontalLine and geoOps.VerticalLine to support Cinderella's freely movable HorizontalLine and VerticalLine types. (Cinderella semantics)


gagern:
I think a single file to test both constructs should be sufficient. They are similar enough that any bug affecting one will likely affect the other as well. You can use colors to distinguish the elements and make sure that they don't get mixed.

In general our geoOps are written in such a way that the geometric constraints have precedence over the specified position. If you e.g. for `Through` use the initial `pos` to describe a line which doesn't pass through the given point, it will be adjusted (via parallel transport in this case) in such a way that it does pass through it. I think we should do the same here, i.e. zero out one coordinate even if it happens to be non-zero on input.

If one coordinate is always zero, we might of course use this fact to truncate the state vector to two complex numbers. On the other hand, the benefit is probably so small that it's not worth the extra code complexity, so leave it as is unless you can think of some particularly elegant way to juggle the coordinates.

The comments are nice this way, should help to make things clear.

elkins0:
OK, you now have the file [ConstructionHorizontalAndVerticalLines.html](@2cf4d7bf3579b4b897f6202810e8e89ef61dac71).

Yes, if there is a possibility that the initial pos value is off in the direction it can be ignored and replaced with the correct one. In that case I will code it to ignore the pos direction entirely and replace it. Convert pos to the homogeneous normal line form [-sin, cos, distance] keeping only the distance from the origin at the normal point to the line.
 * Normalized pos[0.6, -0.8, 5] becomes [0, -1, 4] as HorizontalLine.
 * Normalized pos[0.6, -0.8, 5] becomes [-1, 0, 3] as VerticalLine.

movablefixedangleline

Here is another thought I had initialy considered. As proposed the movable HorizontalLine and VerticalLine and the existing depenpendent Horizontal and Vertical types coding differ only by the directions (List.ex and List.ey). Would it make sense for these to be implemented as `geoMacros.*` and setting the angle or direction before calling more general code?

For example something like this (just off the cuff -- not tried yet):
```javascript
geoMacros.HorizontalLine = function(el) {
    el.type = "MovableFixedAngleLine";
    el.angle = 0; // or el.dir = List.ex;
    return [el];
};

geoMacros.VerticalLine = function(el) {
    el.type = "MovableFixedAngleLine";
    el.angle = Math.PI / 2; // or el.dir = List.ex;
    return [el];
};

geoOps.MovableFixedAngleLine.initialize = function(el) {
...
    if (el.angle) {
        var a = CSNumber._helper.input(el.angle);
        el.dir = List.turnIntoCSList([
            CSNumber.cos(a),
            CSNumber.sin(a),
            CSNumber.zero
        ]);
    }
...
};
...
geoOps.MovableFixedAngleLine.getParamForInput = function(el, pos, type) {
    var homog = List.cross(pos, el.dir);
    return List.normalizeMax(homog);
};
...
```

gagern:
I don't think we need to get overly complicated. If the user entered invalid coordinates, then there is no correct way if interpreting them. The only thing we should care about is that the constraints of the algorithm are satisfied. The reason this is important is because otherwise expectations might get violated. For example, the complex tracing we do expects that if its change in parameter is sufficiently small, then the next position will come arbitrary close to the initial one. This can't happen if the initial position is in violation of the constraints.

To make it short, I'd just explicitely zero one coordinate, i.e. do something like this:

```js
homog = List.turnIntoCSList([homog.value[0], CSNumber.zero, homog.value[2]]);
```

Everything else is up to the user.

The idea of a `MovableFixedAngleLine` is intriguing. I don't think it's worth the effort just to avoid code duplication for `HorizontalLine` and `VerticalLine`. But it might be an interesting object in its own right. I'll discuss this with the other developers.

I just realize that `Horizontal` and `Vertical` in the existing CindyJS semantics might as well be implemented as a pinned version of `Through`. If `MovableFixedAngleLine` gets implemented, it might make sense to keep its direction specification in line with that of `Through`, i.e. allow `dir` or `pos`. And if we want to allow `angle` as well, add that to `Through`, too.

elkins0:
OK, fine by me. Do note however how all this works. The line, whatever constraints it may have, is just the sum of two things, namely a point and a direction. If we are given its homogeneous coordinates el.homog then we can extract a direction using dir = List.cross(el.homog, List.linfty) and a normal point using pn = List.cross(el.homog, dir). If we are given any point on the line, call it p, and a direction, call it dir, we can determine its homogeneous coordinates el.homog = List.cross(p, dir). So there are several ways to supply the point on the line, as dependency using el.args, the pointer position or as the normal point extracted from the pos value homogeneous coordinates. The direction of the line too can be supplied in several ways, as an absolute angle parameter, as a relative angle with respect to the direction/angle of a line dependency using el.args, as the difference of two points and as the direction/angle extracted from the pos value homogeneous coordinates.

I say that if you supply an angle value you should supply pos as the homogeneous coordinates of a point on the line rather than the homogeneous coordinates of the line.

If a line is fully determined by its dependencies why supply any pos or angle at all? For example a parallel line is but a line thru a point with a relative fixed angle of 0&deg; to a line. And by same token, a perpendicular line is but a line thru a point with a relative fixed angle of 90&deg; to a line. One could make intriguing changes by using 180&deg; and 270&deg;, respectively, instead.

The difference here is that in the case of a movable line the dependency is on the pointer rather than another element. This is complicates the code a bit more because we now must support getParamForInput, getParamFromState and putParamToState as well as making putStateComplexVector and putStateComplexVector in a few places. So it is best that movable line (and point) code not be used to implement fully dependent elements. The only rationale for using pinning is if you want be fully general and allow the user to pin and unpin everything. Say be able to unpin a parallel line make it's angle 90&deg; re-pin it and have yourself a perpendicular line? An intriguing thought for  CindyScript perhaps...

gagern:
> The line, whatever constraints it may have, is just the sum of two things, namely a point and a direction.

I'd disagree. A line is less than the sum of a point and a direction, because designating the point is an arbitrary choice, and therefore not part of the line.

> I say that if you supply an angle value you should supply pos as the homogeneous coordinates of a point on the line rather than the homogeneous coordinates of the line.

I disagree again. For the sake of consistency, it should always be possible to give the position of a point or line in its `pos` argument using homogeneous coordinates for the element, and be certain that (semi-)free elements get placed accordingly. The export functionality depends a lot on this.

If you are writing code manually, and you want to rather place the point than the line, then we might want to introduce some other property name for this, but I'd strongly discourage adding additional alternatives at this point unless there is some good reason why we need them. For horizontal and vertical lines at least, converting from point position to line coordinates is simple enough, so I don't see a need for this for those operations.

> If a line is fully determined by its dependencies why supply any pos or angle at all?

Right now because the export in Cinderella is dumb in this respect: It doesn't check which elements are fully determined by its dependencies, but exports coordinates for all points and lines. Might get fixed one day if I can work out (or someone else tells me) how to distinguish these cases.

 > For example a parallel line is but a line thru a point with a relative fixed angle of 0° to a line. And by same token, a perpendicular line is but a line thru a point with a relative fixed angle of 90° to a line.

We don't want to sacrifice our high performance `Parallel` and `Orthogonal` operations for the sake of the more generic `LineByFixedAngle`, but in theory you are right. Perhaps we shouldn't sacrifice the existing `Horizontal` and `Vertical` for that very same reason. Is that the point you wanted to make?

> One could make intriguing changes by using 180° and 270°, respectively, instead.

Lines in CindyJS are conceptually unoriented. Which makes me wonder how Cinderella decides the direction when animating a point on a line, but we don't have such animations now, so I can wonder some more before reproducing that.

To get back to the issue at hand, can you please add explicit zeroing of coordinates to the initialization procedure? I think then we can get this merged and continue other ideas later on.

gagern:
Oh, and since we are currently in conflict, I'd suggest you rebase this branch onto master if you can. If not, because you're restricted to the GitHub web interface, then I'll have to do so before merging.

gagern:
The merge conflict is simply because both this and #96 used the same code position to add their operation. We should make sure to have the code here *before* `LineByFixedAngle` so that it comes right after `Vertical`.

elkins0:
@gagern:OK. I deleted both blank lines ahead of the `geoOps.VerticalLine` code:
```javascript
-
-
// Cinderella's freely movable VerticalLine (Cinderella semantics)
geoOps.VerticalLine = {};
```
This successfully changed the insertion point for the merge. And sorry, I'm just using the GitHub web interface.

gagern:
Merged with some tweaks. You forgot one of the `stateSize` settings. And in general we should treat CindyScript data types (i.e. those with a `ctype` property) as immutable after construction, since we employ copy-on-write semantics.

PR #92, opened 2015-11-22
Implements CSNumber._helper.input

elkins0:
CSNumber._helper.input(a) can be used to convert a plain real number or an object of the form `{ r: number, i: number }` to a CSNumber. Intended for use in allowing pos and other values to be input as complex numbers as well as real numbers. For example:
` pos: [ -3.0, -4.0, { r: -7.654042494670958E-17, i: 1.25 } ]`

### DIFF
--- a/examples/ConstructionRadicalAxisApart-complex.html
+++ b/examples/ConstructionRadicalAxisApart-complex.html
@@ -1,0 +1,44 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>ConstructionRadicalAxisApart.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+    <script type="text/javascript">
+createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: { dimDependent: 0.7 },
+	geometry: [ 
+		{ name: "A", type: "Free", pos: [ 4.0, -3.0, -1.0 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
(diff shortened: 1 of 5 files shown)


strobelm:
Hi, this is a feature we definitely want to have. It is probably not a good idea to assign 0+i*0 as default of parsing goes wrong.

gagern:
I agree with @strobelm: invalid input should be represented by `NaN+i*NaN` in my opinion.

I wonder whether we should perform conversion to number here as well. Something like

```js
if (typeof a === "object")
  CSNumber.complex(+a.r, +a.i);
else
  return CSNumber.real(+a);
```

Note that unary + converts arbitrary types to numbers. In a sensible way where that conversion makes sense, or to NaN where it does not. Of course, this would mean we'd start allowing strings as input, which might cause some people to expect that we'd also accept a complex number in a single string literal. I'm not certain yet, open to comments.

strobelm:
I think the "+" conversion is the way to go, it's very fast and sufficiently safe. If someone wants to pass a string i won't hinder him. We should simply not advertise the possibility.

kortenkamp:
Can somebody explain to me what 'NaN +i*NaN' is? I mean, conceptually? If something is not a number, how can I add or multiply it?

gagern:
`NaN + i*NaN` is an object in the CindyJS internal representation for a complex number, where both the real and the imaginary part are not a number according to the IEEE whatever floating point specs. In my own understanding, this is the canonical representation of a somewhat broader class of not-a-complex-numbers. So for example, `3 + i*NaN` would not really be a complex number, so it could be considered NaN itself. But it's not the floating-point NaN value, so we shouldn't call it NaN.

kranich:
And the '+ i *' is basically just notation to indicate that the tuple (a, b) represents the complex number 'a + i * b'.

gagern:
Cross reference: see #93 for details on the three test cases. I think `ConstructionRadicalAxisApart-complex.html` is the only one we need, since it's the only one demonstrating the new functionality. The `index` thing was already used by other stuff, and since its assignment is kind of arbitrary, I'd prefer to deprecate its use, not add another example demonstrating it.

If noone objects (within the next two hours or so), I'll cherry-pick the relevant commits (i.e. all except for the other two test cases), adapt the helper as discussed, and merge the whole thing into master.

gagern:
And I'll also write a matching pull request on the Cinderella side of things, so we can actually export coordinates like these.

gagern:
Merged in 4bbf1e0348c47ba5347ad3bdd8b48579456aba6f with modifications as discussed. Thanks for the contribution! Will tackle the Cinderella side of this next.

PR #96, opened 2015-11-24
Implementation of LineByFixedAngle type

elkins0:
This proposal implements the LineByFixedAngle type referenced by the "Export to CindyJS..." html file from Cinderella.

In Cinderella, to draw a line at an angle with respect to another line:

1. Draw line `a` through point `A`.linethrupoint
2. Then draw line `b` on line `a` through point `B` at 10&deg;.
 * Click the "Draw Line With Fixed Angle" tile.linebyfixedangle
 * Change the Line angle to 10&deg; in the Info (i) dialog which appears.
 * Click and release near line `a` creating point`B` and line `b`.
 * Change line `b` color to yellow.
2. Now draw line `c` on line `a` through point `A`.
 * Change the Line angle to -20&deg; in the Info (i) dialog which appears.
 * Click near line `a`, drag the pointer over to `A` and release creating line `c`.
 * Change line `c` color to cyan.
3. "Export to CindyJS..." as [ConstructionLineByFixedAngle.html](@162c20edf797994bcd0b7faf92844891e3b3c096).

Note that there appears to be no way to change the angle after creating line `b` and `c` from the Cinderella UI.

Also the proposal expects that the angle will exported as radians not degrees as shown in the "Construction Text" window.

The [ConstructionLineByFixedAngle.html](@162c20edf797994bcd0b7faf92844891e3b3c096)  test file has had the values `angle: -20 * Math.PI / 180` and `angle: 10 * Math.PI / 180` added.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -176,17 +176,18 @@ geoOps.LineByFixedAngle.initialize = function(el) {
     var a = CSNumber._helper.input(el.angle);
     var c = CSNumber.cos(a);
     var s = CSNumber.sin(a);
+    // Setup matrix for applying the angle rotation
     el.rot = List.turnIntoCSList([
-        List.turnIntoCSList([s, c]),
-        List.turnIntoCSList([CSNumber.neg(c), s])
+        List.turnIntoCSList([c, CSNumber.neg(s), CSNumber.zero]),
+        List.turnIntoCSList([s, c, CSNumber.zero]),
+        List.turnIntoCSList([CSNumber.zero, CSNumber.zero, CSNumber.zero])
     ]);
 };
 geoOps.LineByFixedAngle.updatePosition = function(el) {
     var l = csgeo.csnames[(el.args[0])];
     var p = csgeo.csnames[(el.args[1])];
-    var dir = List.turnIntoCSList([l.homog.value[0], l.homog.value[1]]);
-    dir = List.append(List.productMV(el.rot, dir), CSNumber.zero);
-    dir = List.normalizeMax(dir);
+    var dir = List.cross(l.homog, List.linfty);
+    dir = List.productMV(el.rot, dir);
     el.homog = List.cross(p.homog, dir);
     el.homog = List.normalizeMax(el.homog);
     el.homog = General.withUsage(el.homog, "Line");
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -170,6 +170,29 @@ geoOps.Vertical.updatePosition = function(el) {
 };
 
 
+geoOps.LineByFixedAngle = {};
+geoOps.LineByFixedAngle.kind = "L";
+geoOps.LineByFixedAngle.initialize = function(el) {
+    var a = CSNumber._helper.input(el.angle);
+    var c = CSNumber.cos(a);
+    var s = CSNumber.sin(a);
+    el.rot = List.turnIntoCSList([
+        List.turnIntoCSList([s, c]),
+        List.turnIntoCSList([CSNumber.neg(c), s])
+    ]);
+};
+geoOps.LineByFixedAngle.updatePosition = function(el) {
+    var l = csgeo.csnames[(el.args[0])];
+    var p = csgeo.csnames[(el.args[1])];
+    var dir = List.turnIntoCSList([l.homog.value[0], l.homog.value[1]]);
+    dir = List.append(List.productMV(el.rot, dir), CSNumber.zero);
+    dir = List.normalizeMax(dir);
+    el.homog = List.cross(p.homog, dir);
+    el.homog = List.normalizeMax(el.homog);
+    el.homog = General.withUsage(el.homog, "Line");
+};
+
+
 geoOps.Through = {};
 geoOps.Through.kind = "L";
 geoOps.Through.isMovable = true;
--- a/examples/ConstructionLineByFixedAngle.html
+++ b/examples/ConstructionLineByFixedAngle.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>ConstructionLineByFixedAngle.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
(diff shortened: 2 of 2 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -170,6 +170,29 @@ geoOps.Vertical.updatePosition = function(el) {
...
+geoOps.LineByFixedAngle.kind = "L";
+geoOps.LineByFixedAngle.initialize = function(el) {
+    var a = CSNumber._helper.input(el.angle);
+    var c = CSNumber.cos(a);
+    var s = CSNumber.sin(a);
+    el.rot = List.turnIntoCSList([
+        List.turnIntoCSList([s, c]),
+        List.turnIntoCSList([CSNumber.neg(c), s])
+    ]);
+};
```
gagern: Storing the rotation as a matrix makes sense, I think. However, the matrix is not the one I expected (which would be `[[c, -s], [s, c]]`) so it took me a moment to understand why this works. The reason it does is that you start with the coordinates of a line, apply the transformation and treat the result as a point. This reinterpretation adds another 90° so to speak. A comment might help future readers of this code.

Seeing how you do a lot of coordinate-wise manipulation in `updatePosition`, I'd say make this a 3×3 matrix to start with:

```
      ⎡ s  c  0⎤
dir = ⎢-c  s  0⎥ ∙ l.homog
      ⎣ 0  0  0⎦
```

This will make the code below simpler by two lines.

gagern:
I think I'll want this implemented on the Cinderella side first, at least locally, so I can give the combination some testing. Should take just a moment, though.

gagern:
OK, the export from Cinderella works well with this code. I'll wait for a change to the matrix handling before committing.

src/js/libgeo/GeoOps.js
```diff
@@ -170,6 +170,30 @@ geoOps.Vertical.updatePosition = function(el) {
...
+geoOps.LineByFixedAngle.kind = "L";
+geoOps.LineByFixedAngle.initialize = function(el) {
+    var a = CSNumber._helper.input(el.angle);
+    var c = CSNumber.cos(a);
+    var s = CSNumber.sin(a);
+    // Setup matrix for applying the angle rotation
+    el.rot = List.turnIntoCSList([
+        List.turnIntoCSList([c, CSNumber.neg(s), CSNumber.zero]),
+        List.turnIntoCSList([s, c, CSNumber.zero]),
+        List.turnIntoCSList([CSNumber.zero, CSNumber.zero, CSNumber.zero])
```
strobelm: Hi just a quick remark: i was a bit confused why one would use CSNumber.zero in el.rot(3,3). Of course we only deal with far points here, but mathematically this isn't even a projective transformation. 

The rest looks fine to me.

gagern:
@elkins0: Note that GitHub unfortunately doesn't send a notification when you push a new commit for an existing pull request. So I hadn't noticed your update until @strobelm commented on it. It would be nice if you could post a short comment if you made an update which you think addresses all raised concerns, so that I can review that in a timely manner. And I think I should write the GitHub support guys about automatically posting notifications.

elkins0:
It is more elegant than, say:
```javascript
    var cc = CSNumber.cos(el.angle);
    var ss = CSNumber.sin(el.angle);
    var dx = l.homog.value[0];
    var dy = l.homog.value[1];
    var dir = List.turnIntoCSList([
        CSNumber.add(CSNumber.mult(ss, dx), CSNumber.mult(cc, dy)), 
        CSNumber.sub(CSNumber.mult(ss, dy), CSNumber.mult(cc, dx)),
        CSNumber.zero]);
```

kortenkamp:
Just curious: Why don't you use the cross product of `[sin(el.angle), cos(el.angle), 0]` with the point?

elkins0:
@gagern: OK, I can do that. So I had thought this update addresses all raised concerns. But I'm no rush. I'll wait for a consensus and change the code to meet your needs.

elkins0:
@kortenkamp: There are three steps required here:
```javascript
    var dir = List.cross(l.homog, List.linfty); // gets the direction of the reference line
    dir = List.productMV(el.rot, dir); // rotates that direction by the relative angle
    el.homog = List.cross(p.homog, dir); // produces the line's homogeneous coordinates
```
Using the cross product of the point's homogeneous coordinates and the direction `[sin(el.angle), cos(el.angle), 0]` would give me  the homogeneous coordinates of a line with an angle relative to the y-axis rather than the reference line. That is not what is desired here.

kortenkamp:
Ah, sure. I did not remember that we are relative to some other line. 

Anyway: there remains an important question: what about other Geometries?

elkins0:
@kortenkamp: Do you mean Hyperbolic View, Spherical View, Polar Euclidean View and Polar Spherical View as shown in Cinderella's View menu? I have not come across the support code yet. Is there a kernel of it in there already? Got go now...

gagern:
I've merged your code with a slight tweak, to avoid one cross product and instead use the matrix suggested in https://github.com/CindyJS/CindyJS/pull/96#discussion_r45793662, and also to perhaps mitigate the confusion @strobelm expressed.

I was confused because the matrix in your latest commit was not the one I had suggested. I had not noticed the cross product at first, and assumed that the difference should result in a change in angle. So I wanted to test this before commenting, and only just now found the time to do so.

@kortenkamp: We definitely are not supporting non-Euclidean geometries at this point. If we do one day, we might want to do so in a completely different set of geometric operations, both for the sake of performance and to clearly distinguish which operations are implemented (or even make sense) in a non-Euclidean setup. We'd probably change a number of operator signatures as well, e.g. turn midpoint and parallel into multi-valued operations, which doesn't agree well with the rule of having the signatures clean and independent of input types.

elkins0:
@gagern:Sorry about that. I was just about to make that change myself. I had seen this:

```
      ⎡ s  c  0 ⎤
dir = ⎢-c  s  0 ⎥ ∙ l.homog
      ⎣ 0  0  0 ⎦
```
and thought:
```
dir =  l.homog × [0 0 1]
      ⎡ c  -s  0 ⎤
dir = ⎢ s  c  0 ⎥ ∙ dir
      ⎣ 0  0  0 ⎦
```
and thought, yes that would work, no sense it making it harder to read. And your saying this will make the code simpler by two lines threw me for a loop. The `dir = List.normalizeMax(dir);` was not needed in any case. So I went for simplicity rather than optimization. This morning I thought about @strobelm's comment about all those zeros and thought wait, I could at least use them to better effect then realizing what you were trying to say. But note that in either case `List.productMV` with a `3x3 ∙ 3` is more than twice the CSNumber.ops of a `2x2 ∙ 2`, i.e. 18 versus 8.

PR #99, opened 2015-11-26
Rename IntersectConicConic IntersectionConicConic

elkins0:
I simply renamed IntersectConicConic to IntersectionConicConic in six places in GeoOps.js so that it matches up to the Cinderella html exported for CindyJS. The existing code works fine with the included test file.


strobelm:
Hi! I'm not sure whether it would be easier to adapt Cinderella here, since the renaming would break naming conventions in geoOps: geoOps.IntersectLC, geoOps.IntersectCirCir, geoOps.IntersectConicConic and additionally many examples which rely on the name IntersectConicConic: examples/88_TracingVis2.html examples/69_tracing4_many.html examples/69_Intersection_Conic_Conic.html examples/69_tracing4.html.

gagern:
Renaming existing types is problematic with respect to backwards compatibility. You could instead follow the things we did with `IntersectionConicLine` and `IntersectionCircleCircle`, namely [establish aliases](https://github.com/elkins0/CindyJS/blob/19e0e9472e5a683afa2957ad0df18af35dcd9474/src/js/libgeo/GeoOps.js#L1516-L1525).

On the developer mailing list we decided that at some point we'd unify the names so that both in the code exported from Cinderella and in the implementation in geoOps, the clearer name for each operation would be chosen. In most cases that's the name Cinderella uses internally (with `Center` vs. `CenterOfConic` the only exception that I know of). So one day we'll likely swap things, use the Cinderella name as default name and the other as an alias. We might even want to introduce some `geoAliases` table, since `geoMacros` is too powerful (and too verbose) for the simple renaming we do most of the time. But I'd rather do all that renaming in one step. If you want to, you can make that part of your pull request, but I think adding a single alias would be preferable just now.

elkins0:
So then:
```javascript
geoMacros.IntersectionConicConic = function(el) {
    el.type = "IntersectConicConic";
    return [el];
};
```
like the existing
```javascript
geoMacros.IntersectionConicConic = function(el) {
    el.type = "IntersectConicConic";
    return [el];
};
```
What about ConicBy4p1l and it ilk? Shall I roll them all into this pull request or would you rather have each separately?

gagern:
I think I'd prefer a single pull request with all the simple renaming issues you can find in #24. But with separate commits and separate test cases.

I'm not sure `ConicBy4p1l` is such a simple case, though. It's a multi-valued operation, and while Cinderella already should export that as two distinct operations, I doubt it's exporting sufficient data to actually choose the right conic in all situations. Will give it a try later on (too busy just now), but that might require more work on the Cinderella side. If in doubt, better leave those more complicated cases out of this pull request for now.

elkins0:
@gagern: I think you mean  "ConicFociH" which just uses the second result of "ConicBy2Foci1P"...

OK. Here's where I'm headed with this (add two lines here):
```javascript
function csinit(gslp) {
...
    // Das ist für alle gleich || This is the same for all
    for (k = 0; k < gslp.length; k++) {
        el = gslp[k];
+       var typeAlias = geoAliases[el.type];
+       if (typeAlias) el.type = typeAlias;
        var macro = geoMacros[el.type];
```
and an geoAliases object:
```javascript
var geoAliases = {
    "AngularBisector": "angleBisector", // change SelectP -> SelectL + index: 1..2
    "Arc": "ArcBy3",
    "CircleByFixedRadius": "CircleMr", // + radius: value + pinned: value
    "CircleByRadius": "CircleMr",
    "Conic1P4L": "ConicBy1p4l",  // change SelectP -> SelectConic + index: 1..4
    "Conic2P3L": "ConicBy2p3l",  // use SelectConic + index: 1..2
    "Conic3P2L": "ConicBy3p2l",  // use SelectConic + index: 1..2
    "Conic4P1L": "ConicBy4p1l",  // change SelectP -> SelectConic + index: 1..4
    "ConicFoci": "ConicBy2Foci1P", // change SelectP -> SelectConic + index: 1..2
    "ConicFociH": "ConicBy2Foci1P", // change SelectP -> SelectConic + index: 2
    "IntersectionCircleCircle": "IntersectCirCir",
    "IntersectionConicConic": "IntersectConicConic",
    "Parallel": "Para"
};
// Nota bene:
// erg ist Ergebnis || res is result
```
The comments indicate the additional changes I found necessary in the lines which follow in the geometry of the html Cinderella creates to get them to work. Also the examples 23_Mirrors.html & 52_Rays.html make use of geoMacros.CircleMFixedr without setting "pinned: true".

gagern:
> I think you mean "ConicFociH"…

No. I meant ConicBy4p1l. If you look at the Cinderella internals (e.g. by copying such an object to a text editor), you will see that there is one operation which results in two objects. Exporting this to CindyJS, we get two operations, namely `ConicBy4p1l` and some `Select*`.

We shouldn't use `Select*` + `index` since we don't make any guarantees about which result is which. All you get is that if an operation has *n* results then the indices 0 through *n*−1 will describe them all. We really need Cinderella to export the matrices of its conics.

Introducing the alias table now would conflict with [the create branch](https://github.com/gagern/CindyJS/commits/create) which I'm preparing for review. So the two lines in `csinit` would need to get moved manually. No big deal, though. `CircleByFixedRadius` should not be handled with an alias, since the pinning requires more than that.

elkins0:
OK I'm done here. So I won't mess with geoAliases at all. I have added a geoMacros function for renaming "IntersectionConicConic" to "IntersectConicConic". The only other simple rename would be "Arc" to "ArcBy3". The rest of the others all would have to have index values added to the examples which would later have to be removed and replaced with a new name value pair. I don't understand how to handle a "separate commit" for "Arc" within the same pull request.

gagern:
I think the GitHub interface will make a separate commit for each of your edits. This pull request here currently has three distinct [commits](https://github.com/CindyJS/CindyJS/pull/99/commits).

I think `ArcBy3` is the more descriptive title, so I'd prefer changing this on the Cinderella export side. I'll give your pull request as it stands some testing when I find the time; will be traveling tomorrow and I don't know whether I'll have sufficient connectivity on the trip.

Looking at the code, I wonder why the example is as long as it is. Haven't looked at it yet, so perhaps it's sufficiently interesting to warrant that size. Otherwise one might consider truncating it after all four intersections have been constructed, and perhaps use different colors for those to verify correct assignment.

strobelm:
Hi, i merged the macro directly into the master. Anything else looked redundant, any further wished or can i close this?

gagern:
I think we're done here. Thanks @strobelm for taking this on.

PR #100, opened 2015-11-27
Initialize SelectConic from coordinates

gagern:
The SelectConic element now accepts a description of a conic using 6 variables, corresponding to the equation

> xx⋅x² + yy⋅y² + zz⋅z² + xy⋅x⋅y + xz⋅x⋅z + yz⋅y⋅z = 0

Currently the distance measurement is based on projectiveDistMinScal.  It might make sense to try out something along the lines of tracingSesq for this as well.

### DIFF
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -1034,6 +1034,28 @@ List.projectiveDistMinScal = function(a, b) {
 
 };
 
+function conicMat2Vec(m) {
+    var v = m.value;
+    var r0 = v[0].value;
+    var r1 = v[1].value;
+    var r2 = v[2].value;
+    return List.turnIntoCSList([
+        r0[0],
+        CSNumber.add(r0[1], r1[0]),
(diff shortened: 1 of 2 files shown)


strobelm:
@gagern, the commit breaks 65a_Conic_3P_2l.html and 65_Conic_4P_1l.html. Is this somehow related to issue #101?

gagern:
It seems I deleted too much out of the SelectcConic code. I had assumed that whatever operation created this set of conics would already have normalized and tagged the matrices, but apparently that's not the case. Thanks for catching this, and also thanks for verbally pointing out the debug code in there.

If we one day decide that we should move normalization and/or tagging to the set-creating operations, that would be a different issue. I'm undecided whether the benefits of having “cleaner” data in all our representations outweights the drawback of having to reproduce (or call) the same functionality in more places.

strobelm:
Much better, thanks!

Issue #56, opened 2015-10-29
Clip lines through infinity to canvas


gagern:
Currently, [lines](https://github.com/CindyJS/CindyJS/blob/e7a5496832a21db46741c9e7926b2f8478595e92/src/js/libcs/Render2D.js#L451-L479) and [rays](https://github.com/CindyJS/CindyJS/blob/e7a5496832a21db46741c9e7926b2f8478595e92/src/js/libcs/OpDrawing.js#L253-L275) through infinity have some pretty arbitrary extent when rendering. The former has its size determined in user coordinates, while the latter uses pixel coordinates. Both may fail for exotic zoom levels or canvas sizes. It would be more appropriate to clip things to the size of the canvas, extended by width of the line.

Issue #53, opened 2015-10-29
Render projectively transformed segments 


gagern:
*(Manually migrated from GitLab issue 20 reported by @gagern)*

While ee4954cdd4ac278c12778119ca187c5df585e474 provided a `TransformS` operation to apply a projective transformation to line segments, the rendering engine doesn't handle that right. In particular

1. if the segment passes through infinity, then the wrong part is rendered
2. contrary to Cinderella, overhang isn't computed with regard to the projective basis

gagern:
The code from #26 for drawing a pair of rays might be useful here, too.

PR #103, opened 2015-11-27
Fix Render2D.drawline to correctly clip

elkins0:
The test file demonstrates the problems. The clipping is not correct as currently implemented by Render2D.drawline. Depending on the angle of a line and where it is drawn it may be drawn outside the clipping rectangle and not within even though it clearly should be.

### DIFF
--- a/examples/LineClippingTest.html
+++ b/examples/LineClippingTest.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>LineClippingTest.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+    <script type="text/javascript">
+createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0 }, 
+	angleUnit: "°", grid: 10,
+	geometry: [ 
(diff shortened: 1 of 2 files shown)


gagern:
The clipping of lines to the ±30 units square is a bug (namely #56), not a feature. Ideally, we'd use the screen size (e.g. as returned by [screenbounds](https://github.com/CindyJS/CindyJS/blob/6d865efd8418caebcb1e1e88797d6a7814ef0650/src/js/libcs/OpDrawing.js#L1355-L1361)) (ideally adjusted for line width) and clip to that. And if we do, it doesn't matter whether the clipping is clean or not; any clipping which leaves all the visible portions of the conceptually infinite line visible will do.

So I see no reason to merge your code because clipping becomes cleaner that way. But your code also is *way* shorter than what we currently have, and that *is* a concern, in terms of performance and maintainability. It took me a bit to understand what you do. Essentially you're checking for each segment whether the two endpoints lie on different sides of the line. I'd like to think a bit more about whether there can be any corner cases where this does not work as intended, but due to your use of booleans I think it should work as long as the line width is taken into account for clipping.

It would be interesting to do a performance comparison between your code and what we had before. You or one of us could come up with some example which draws *many* lines, and we could add [Performance](https://developer.mozilla.org/en-US/docs/Web/API/Performance) timers to measure how long each draw takes in general. Not sure wheether and when I will find the time for this.

Note that JSHint is unhappy with your code as it uses `!=` instead of `!==`. The latter is semantically  cleaner since it doesn't do any funny type conversions, and it's also better in terms of performance. So when you edit your code, please change those.

elkins0:
Finally! So, well, use it as you best see fit. For now I'll just have to keep the view in side those bounds. There's a whole cottage industry grown up around optimizing JavaScript. Other than for things like this I keep it turned off or highly restricted using a combination of Firefox and NoScript and turn off images most of the time too.

elkins0:
Bloody buttons...

kortenkamp:
I just would like to add comment - be careful, segments can go through infinity.

gagern:
@kortenkamp These are complete infinite lines here, not segments, so we don't have to worry yet. Segments *should* go through infinity, but that's something #53 has to take care of. Perhaps building on the code here.

gagern:
OK, I did some performance checks, and couldn't find a significant difference one way or the other, using Firefox and Chrome. After beautification the difference in code size is almost negligible as well. I'll leave this open for a bit so we can discuss whether there are any other benefits or drawbacks to consider.

kranich:
Superseded by #292. Thank you, @elkins0 for your many helpful contributions to #292.

PR #104, opened 2015-11-27
Clean up Pole and Polar, breaking compatibility

### LINKED ISSUE #86: Naming inconsistency for polars
Currently, `PolarPoint` and `Polar` (with its alias `PolarLine`) are treated inconsistently. The `kind` property suggests that `PolarPoint` is a point while `PolarLine` is a line. The `usage` of the resulting vector is exactly reversed, as is the implemented computation. Cinderella internally follows that latter interpretation as well. My first intuition regarding the terms would have assumed the former to be correct (which is the reason why I wrote the `kind` properties that way in the first place).

I'm unsure how to handle this. Switching names might in my opinion avoid confusion in the long run, but would cause problems for existing content and could easily cause confusion when reading Cinderella internals, since then Cinderella would have to translate on output. Not switching names but adjusting the `kinds` would mean we'd have to stick to the current interpretation in the long run.

A third solution could be renaming the operations to `Pole` and `Polar`, make things consistent with that and avoid reusing terms which are used in a confusing way in Cinderella.

gagern:
With `PolarPoint` and `PolarLine`, there was a horrible mess about which is which, mostly because Cinderella uses them in an intuitive way: `PolarPoint` is a line, `PolarLine` a point.  To avoid any such confusions in the future, this commit introduces a new pair of nouns, namely `Pole` (which is a point
defined by a line and a conic) and `Polar` (which is a line defined by a point and a conic).  The order was unifies so that the conic always comes last.

Since the state of affairs so far was such a mess, the commit deliberately does not try to preserve backwards compatibility.

This pull request fixes #86. For those interested in the history of the issue: bf59f449655737739b30723e6098ae0d28d16c04 implemented `Polar` as we are now using it. 2664a18c8f3f50e34c38f72f2829c7d9900220f8 added `PolarPoint` and `PolarLine`, modified the existing code and thus created the current inconsistent situation.

### DIFF
--- a/examples/101_PolarOperations.html
+++ b/examples/101_PolarOperations.html
@@ -39,11 +39,11 @@ var cdy = createCindy({
     {name: "A", type: "Free", pos: [3.714285714285714, -4, -1.7857142857142856], color: [1, 0, 0], labeled: true},
     {name: "C0", type: "CircleByRadius", args: ["A"], radius: 4.216918306061904, size: 1, color: [0, 0, 1], printname: "$C_{0}$"},
     {name: "B", type: "Free", pos: [0.6333333333333333, -4, -0.8333333333333334], color: [1, 0, 0], labeled: true},
-    {name: "a", type: "Polar", args: ["B", "C0"], pos: [-0.25419812047450324, -0.4929902942535819, 4], size: 1, color: [1, 0, 0], labeled: true},
+    {name: "a", type: "PolarOfPoint", args: ["B", "C0"], pos: [-0.25419812047450324, -0.4929902942535819, 4], size: 1, color: [1, 0, 0], labeled: true},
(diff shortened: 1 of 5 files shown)


gagern:
This should be ready for review now. The CenterOfConic example now works, too.

strobelm:
i just started to rework the code, but reviewing is even better ;-).

Am 27.11.2015 um 22:27 schrieb Martin von Gagern:
>
> This should be ready for review now. The CenterOfConic example now
> works, too.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/104#issuecomment-160206999>.
>

-- 
Michael Strobel, M.Sc.

strobelm:
Looks fine, even the old examples. Thanks!

PR #111, opened 2015-11-28
Implementation of ConicPrincipleDirs

elkins0:
Support for the algorithm "ConicPrincipleDirs" type referenced by html produced by "Export to CindyJS" from Cinderella. This is one of the items on the #24 list.


elkins0:
See [ConstructionConicPrincipleDirs-PerpendicularM-P5-alt.html](https://github.com/elkins0/CindyJS/blob/patch-2/examples/ConstructionConicPrincipleDirs-PerpendicularM-P5-alt.html) for an example construction of the right angle case.

gagern:
I'd rather not do the special case handling you suggested. Instead, I'd prefer to go through `geoOps._helper.conicFromTwoDegenerates`.

So once you have `M`, `P1` and `P2`, compute the point `P3` and the line `P1M` as you did. Then compute lines orthogonal to `P1M` passing through `P1` and `P3`. These two lines form one degenerate conic, and `P1M` itself taken twice is another degenerate conic. These are conics which pass through `P1` and have center `M` and the desired symmetry properties. The requested conic must lie in that pencil of conics. `conicFromTwoDegenerates` can find it.

Let me know whether you want to give this a try, or whether you want me to take care of it when I find the time.

gagern:
I also think the term should be `ConicFromPrincipalDirections`, since `Principle` seems misspelled, and since omitting `From` made me think at first that this were an operation which takes a conic as an argument and returns its axes of symmetry or something like that. If you agree I'll promote a suitable renaming upon export from Cinderella.

elkins0:
I would definitely like to give your recipe a try. I did not want to have to add special case handling either.  I agree with using ConicFromPrincipalDirections instead of the ConicPrincipleDirs.

elkins0:
@gagern: OK, code is ready to go. It is much cleaner now. Thank you so much for your help!

gagern:
Squashed to [three commits](https://github.com/CindyJS/CindyJS/compare/35608b4fcbef95c1d3c039aaf3f8312a19984679...9bbf58d3fd104f2cdb66983c581374f062ce8be1) and merged in 720c61f312c37a0d3f3e72602724a0f798b4c52b. Thank you very much for this contribution!

PR #114, opened 2015-11-30
Avoid using MathJax in our examples

gagern:
Using the KaTeX plugin for in-canvas rendering is preferred.


strobelm:
Hi the examples look fine, thanks for the cleanup @gagern

PR #116, opened 2015-12-01
Implementation of ConicInSquare

elkins0:
Adds support for the algorithm "ConicInSquare" type referenced by html produced by "Export to CindyJS" from Cinderella. This is one of the items on the #24 list.

The code makes use geoOps._helper.conicFromTwoDegenerates and just like geoOps.ConicFromPrincipalDirections has one of the degenerate conics comprised of a doubled up line. Also the "ConicInSquare" name may be subject to renaming, the simplest being ConicBy4points. The conic is certainly not always inside nor is it always square. In German "Viereck" may have been loosely translated in the past to "square" instead of quadrilateral or quadrangle.

### DIFF
--- a/examples/ConstructionConicInSquare.html
+++ b/examples/ConstructionConicInSquare.html
@@ -1,0 +1,47 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>ConstructionConicInSquareWithDottedLines.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+    <script type="text/javascript">
+createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0 }, 
+	angleUnit: "°", 
+	geometry: [ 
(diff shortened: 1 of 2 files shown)


strobelm:
Thanks @elkins0 for implementing this! I rebased the commits, renamed the example and pushed it to master.

gagern:
I [evaluated this implementation](https://gist.github.com/gagern/9fdd5175933ce6a5e557) and could confirm that it is correct (which is great since I only have a vague idea as to what's going on here), but also find that it has unneccessary degenerate situations. This time, my approach is easy to explain: take the (±1,±1) square and the inscribed unit circle, and map the square corners projectively to the given points. One could probably reuse [`evaluator.map$8`](https://github.com/CindyJS/CindyJS/blob/927e87cd702d2f48693a1f5acded7ff70d413774/src/js/libcs/Operators.js#L3267-L3284) or [`geoOps.TrProjection.updatePosition`](https://github.com/CindyJS/CindyJS/blob/927e87cd702d2f48693a1f5acded7ff70d413774/src/js/libgeo/GeoOps.js#L1591-L1614) for this task. One might even consider writing this as a macro, once we can apply projective transformations to conics.

@elkins0 do you want to pick this up again? It's not urgent, and you don't have to unless you want to.

elkins0:
Sure, I would certainly like to continue on with this. It was only a proposal after all. I believe I understand what you are saying. Now it make sense why it should be called "ConicInSquare". I had not looked at it that way before. My old approach was to use a geometric construction to define a five-point-conic. The new approach cut-out the intermediate five-point-conic step and used the conicFromTwoDegenerates method directly.

In this image the labels shown match the code. The purple lines b & d are the first degenerate conic which is tangent to the conic at the intersection point with the cyan line g which are the double lines of the the second degenerate conic. The point H is located at the tangent point of line c with the desired conic:

conicinsquare-2

This shows the old approach using a geometric construction to define a five-point-conic. The first four tangent points (H,K, L & M) to the red lines a, b, c & d are easily defined. The final point (O) is a bit more of a challenge:

conicinsquare-1

elkins0:
The follow on continues at #129.

PR #118, opened 2015-12-01
add normalization to conic drawing

### LINKED ISSUE #102: Conics incorrectly rendered as degenerate
I've tried this example with the code from #100. I found that moving the points, the conic very often gets rendered as a degenerate conic, i.e. a pair of lines, even if it is pretty far from a degenerate situation. Sometimes one of the lines doesn't even pass through any of the defining points. This needs investigating.

```js
createCindy({ 
	scripts: "cs*", 
	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0 }, 
	angleUnit: "°", 
	geometry: [ 
		{ name: "A", type: "Free", pos: [ 1.4414414414414414, -4.0, -0.9009009009009008 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "B", type: "Free", pos: [ 4.0, 2.2335025380710665, 0.5076142131979695 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "a", type: "Join", color: [ 0.0, 0.0, 1.0 ], args: [ "A", "B" ], labeled: true }, 
		{ name: "C", type: "Free", pos: [ -2.08, 3.68, 4.0 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "D", type: "Free", pos: [ -1.161290322580645, -4.0, 3.225806451612903 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "E", type: "Free", pos: [ 4.0, -2.7555555555555555, 2.2222222222222223 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "F", type: "Free", pos: [ 4.0, -0.9491525423728815, 1.6949152542372883 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "Collection__1", type: "ConicBy4p1l", args: [ "C", "D", "F", "E", "a" ] }, 
		{ name: "C0", type: "SelectConic", pos: { xx: { r: 0.07387367172659677, i: 6.938893903907228E-18 }, yy: 1.0, zz: { r: -0.9072629140306034, i: 5.551115123125783E-17 }, xy: { r: -0.46083968841032463, i: 5.551115123125783E-17 }, xz: 0.4650631263425033, yz: { r: -0.5469413106840574, i: 1.1102230246251565E-16 } }, color: [ 1.0, 1.0, 0.0 ], args: [ "Collection__1" ], printname: "$C_{0}$" }, 
		{ name: "C1", type: "SelectConic", pos: { xx: { r: -0.6078235732312288, i: -5.551115123125783E-17 }, yy: { r: -0.24110554809642848, i: 1.3877787807814457E-17 }, zz: { r: 1.0, i: -5.551115123125783E-17 }, xy: -0.3125808979848061, xz: { r: 1.2628662589541286, i: 1.1102230246251565E-16 }, yz: { r: -0.18984389822023837, i: 2.7755575615628914E-17 } }, color: [ 1.0, 1.0, 1.0 ], args: [ "Collection__1" ], printname: "$C_{1}$" } ], 
	ports: [ 
		{ id: "CSCanvas", width: 680, height: 336, transform: [ { visibleRect: [ -9.06, 9.34, 18.14, -4.1 ] } ], background: "rgb(168,176,192)" } ], 
	cinderella: { build: 1798, version: [ 2, 9, 1798 ] } });
```

strobelm:
I investigated bug #102. The conic drawing routines runs in to trouble in some examples due to a lack of normalization. After normalizing the matrix in the routine the degenerate drawings vanish.

This should resolve #102, perhaps @gagern could verify.

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -377,11 +377,13 @@ evaluator.drawconic$1 = function(args, modifs) {
             Conic.matrix = arr;
         }
 
+
     }
-    return eval_helper.drawconic(Conic, modifs);
+    Conic.matrix = List.normalizeMax(Conic.matrix);
+    return eval_helper.drawconic(Conic.matrix, modifs);
 };
 
-eval_helper.drawconic = function(aConic, modifs) {
+eval_helper.drawconic = function(conicMatrix, modifs) {
 
     Render2D.handleModifs(modifs, Render2D.conicModifs);
     if (Render2D.lsize === 0)
@@ -389,8 +391,8 @@ eval_helper.drawconic = function(aConic, modifs) {
     Render2D.preDrawCurve();
 
     var eps = 1e-14; //JRG Hab ih von 1e-16 runtergesetzt
-    var mat = aConic.matrix;
-    var origmat = aConic.matrix;
+    var mat = List.normalizeMax(conicMatrix);
+    var origmat = mat;
 
     // check for complex values
     for (var i = 0; i < 2; i++)
@@ -598,7 +600,7 @@ eval_helper.drawconic = function(aConic, modifs) {
 
 
         var step;
-        var perc = 0.05;
+        var perc = 0.1;
         var diff = ymax - ymin;
         var ssmall = perc * diff + ymin;
         var slarge = ymax - perc * diff;
@@ -611,7 +613,7 @@ eval_helper.drawconic = function(aConic, modifs) {
                 step = 3;
             }
 
-            var inner = -a * c * Math.pow(y, 2) - 2 * a * e * y - a * f + Math.pow(b, 2) * Math.pow(y, 2) + 2 * b * d * y + Math.pow(d, 2);
+            var inner = -a * c * y * y - 2 * a * e * y - a * f + b * b * y * y + 2 * b * d * y + d * d;
             inner = Math.sqrt(inner);
 
 
--- a/src/js/libgeo/GeoRender.js
+++ b/src/js/libgeo/GeoRender.js
@@ -69,8 +69,7 @@ function render() {
         modifs.alpha = el.alpha;
         modifs.size = el.size;
 
-        var args = el;
-        eval_helper.drawconic(args, modifs);
+        eval_helper.drawconic(el.matrix, modifs);
 
 
     }


gagern:
Yes, this does resolve #102.

I'm a bit unhappy about the location where you fixed this: in my opinion, the renderer should treat the geometry as read-only. Even though a change in representative should be meaningless in most situations, the fact that you added this to fix a bug indicates that reality is something different. And I could imagine that in a few years we might be spending days debugging code which behaves different depending on whether you draw things in between.

Points get normalized after essentially any geometric operation. Perhaps we should do the same for conics. Or if we normalize things in the renderer, then we should not write the normalized representation back to the element. I fear we might want both of these: the former will ensure that operations other than rendering won't suffer from overflow or underflow due to lack of normalization. And the latter will ensure that the fix also applies to conics drawn by the user using a script command instead of a gslp element. What do you think?

gagern:
I fear this issue was my fault. As I just noticed in https://github.com/CindyJS/CindyJS/pull/100#issuecomment-161142656, I incorrectly removed some normalization from `SelectConic` in the code I was using during testing. Changing the code from select-by-position to select-by-index, things look smooth against the current master.

So the original point of this ticket here seems to be moot. Nevertheless, there is this question of whether we should normalize input from scripts. Should I file a new report for this, or do you want to re-purpose this one here to that effect?

strobelm:
I'm fine reusing this one and agree that altering the element from the drawing probably isn't the best idea, i was a bit hasty. 

Normalizing the conic handed over by scripts is generally a good idea. The best thing would be that the drawing ops always get a normalized version. Nevertheless i'm currently in favor of a fallback, non element altering, in the drawing routine, just to be sure.

strobelm:
I updated the branch and added normalization to evaluator.drawconic, removed the modification of the element and changed some lines for performance in the drawing routine. 

the variable perc handles how many points are generated at the apex of the parabola, increasing this from 5 to 10% has not a real performance impact, but the conics look a bit better.

src/js/libcs/OpDrawing.js
```diff
@@ -389,8 +391,8 @@ eval_helper.drawconic = function(aConic, modifs) {
     Render2D.preDrawCurve();
 
     var eps = 1e-14; //JRG Hab ih von 1e-16 runtergesetzt
-    var mat = aConic.matrix;
-    var origmat = aConic.matrix;
+    var mat = List.normalizeMax(aConic.matrix);
+    var origmat = JSON.parse(JSON.stringify(mat));
```
gagern: Why would you want to do that JSON detour? All our CindyScript-typed data structures are treated as immutable in any case, so there should be no reason to make a copy. The [conic transformation](https://github.com/strobelm/CindyJS/blob/927e87cd702d2f48693a1f5acded7ff70d413774/src/js/libcs/OpDrawing.js#L427-L429) creates a new matrix and causes `mat` to refer to this, without affecting the original matrix.

src/js/libcs/OpDrawing.js
```diff
@@ -377,7 +377,9 @@ evaluator.drawconic$1 = function(args, modifs) {
             Conic.matrix = arr;
         }
 
+
     }
+    Conic.matrix = List.normalizeMax(Conic.matrix);
```
gagern: For a moment I was afraid that this, too, would modify an existing element. But that's just a local object. Which made me think: why do we have that object? Why not simply pass the matrix directly to `eval_helper.drawconic`, without enclosing it into such an object? Not really related to the issue at hand here, just got me wondering.

strobelm: Currently eval_helper.drawconic relies on the fact that aConic is handed over as an object with property matrix, we could of course change that which would entail some reworking of calls of drawconic.

src/js/libcs/OpDrawing.js
strobelm: You are right, the unnecessary detour was due to safety considerations. I'll change that.

strobelm:
Here is the update. Let me know if there are further wishes (especially the re-working of the eval_helper.drawconic argument handling). If not i would squash the commits for merging this.

gagern:
Sorry, I forgot about this request. Yes, I'd say change the argument handling to just pass the matrix, then squash this.

strobelm:
Here is the squashed update with direct matrix handling.

Issue #107, opened 2015-11-27
Welcome deverlopers on Windows


gagern:
It would be nice if we could come up with a way how developers using Windows could build CindyJS from source. Since the code itself is JavaScript, being cross-platform seems natural, and it would be a shame if we lost that just because of a too *nix-oriented build system.

The main problem is the central Makefile and the utilities it relies on. We could ask users to install [Cygwin](http://cygwin.com/), but that's a pretty big install and not everyone might be comfortable with this. We could try [MSYS](http://mingw.org/wiki/MSYS) or [MSYS2](http://sourceforge.net/projects/msys2/), see whether that's good enough for our needs. Theoretically we could also try reproducing the functionality of GNU make using JavaScript. Or we could move away from Make ourselves, and use some JavaScript build tool (like [Gulp](https://www.npmjs.com/package/gulp) or [Grunt](https://www.npmjs.com/package/grunt) or something else) to control our builds.

The dependency on Java is probably something we won't get rid of easily. We need it to run the closure compiler and to build the GWT-based quickhull plugin. We can try to ensure that `make plain=1` (or its analogon in a new build system) works without Java. If we can get the tests to work as well, that would be great. And we could consider downloading Java automatically. I think it should be possible to set this up using environment variable tweaking, without permanent modifications to the system outside the CindyJS working directory.

gagern:
Short status report: I've rewritten the build system using JavaScript. I wasn't happy with several of the concepts used by Gulp or Grunt. Grunt felt too inflexible for my taste, since we have some pretty exotic steps in there, and Gulp has the tendency to keep everything in memory, while the closure compiler reads its input from disk. Having the gulp task write them back to disk to a temporary dir doesn't make too much sense in my opinion. So I essentially duplicated the core logic of make: look at the inputs, look at the output(s), compare timestamps, and run tasks where the targets are outdated. Got some minor issue to work out yet, then a local test run on an emulated Windows, and then it should be ready for some public testing.

kranich:
@gagern Can we close this?

gagern:
I was waiting for feedback from @elkins0 on whether he can build on Windows. So far it's mostly theory, since the system on which I tested this has Cygwin installed, so it's not exactly your standard run of the line Windows with respect to what commands are available.

elkins0:
[...]

PR #119, opened 2015-12-01
Js make

gagern:
This switches the build system from Makefile to JavaScript, in an attempt to address #107.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js: node
-script: make alltests
+script: node make alltests
 sudo: false
 addons:
   apt:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-all: build/js/Cindy.js cindy3d
+all: js_make
+	$(JS_MAKE)
 
 clean:
 	$(RM) -r build
@@ -6,26 +7,6 @@ clean:
 .PHONY: all clean
 
 .DELETE_ON_ERROR:
-
-######################################################################
-## List all our sources
-######################################################################
-
-libcs := src/js/libcs/Namespace.js src/js/libcs/Accessors.js src/js/libcs/CSNumber.js src/js/libcs/List.js src/js/libcs/Essentials.js src/js/libcs/General.js src/js/libcs/Operators.js src/js/libcs/OpDrawing.js src/js/libcs/OpImageDrawing.js src/js/libcs/Parser.js src/js/libcs/OpSound.js src/js/libcs/CSad.js src/js/libcs/Render2D.js
-
-libgeo := src/js/libgeo/GeoState.js src/js/libgeo/GeoBasics.js src/js/libgeo/GeoRender.js src/js/libgeo/Tracing.js src/js/libgeo/GeoOps.js src/js/libgeo/GeoScripts.js
-
-liblab := src/js/liblab/LabBasics.js src/js/liblab/LabObjects.js
-
-lib := lib/clipper/clipper.js
-
-inclosure = src/js/Setup.js src/js/Events.js src/js/Timer.js $(libcs) $(libgeo) $(liblab) $(extra_inclosure)
-
-ours = src/js/Head.js $(inclosure) src/js/Tail.js
-
-srcs = $(lib) $(ours)
-
-js_src = $(filter-out tools/%,$(filter %.js,$^))
 
 ######################################################################
 ## Download stuff either using curl or wget
@@ -58,6 +39,7 @@ NPM_DEP:=$(if $(NODE_NEEDED)$(NPM_NEEDED),download/$(NODE_BASENAME)/bin/npm,)
 NODE_PATH:=PATH=node_modules/.bin:$(if $(NPM_DEP),$(dir $(NPM_DEP)):,)$$PATH
 NPM_CMD:=$(if $(NPM_DEP),$(NODE_PATH) npm,$(NPM))
 NODE_CMD:=$(if $(NPM_DEP),$(NODE_PATH) node,$(NODE))
+JS_MAKE=$(NODE_CMD) make/index.js js_compiler=$(js_compiler)
 
 download/arch/$(NODE_TAR):
 	mkdir -p $(@D)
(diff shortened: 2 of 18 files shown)


strobelm:
I have mixed feelings about this.

On the one hand a really appreciate the work you put into this and the flexibility concerning other platforms. On the other hand 'make' is tested and very well documented, while a custom script is not (by nature). If you get hit a train tomorrow (perish the thought!) nobody would really know how our build system works.

If we merge this it would be really a good idea to have at least a documentation how basic work flow with the script looks like.

(by the way travis reported a build error)

gagern:
I'd view this under a bit of a different angle: both Make and Node.js are well tested and documented. Everything else built on top of that. Or in other words, I think I'm doing some pretty strange and non-standard things with the Makefile already, and would guess that were I hit by a train, you might have a tough time with that, too.

But I agree that this might be made a bit more severe by some of the added code, and documenting this is certainly a good idea. I think I'll drop a `README.md` into that `make` directory, explaining things. Then I'll let you look over this once more, and see whether you can follow the logic. If so, you can be my backup in case of a train accident. If not, I should continue to clarify until we get there.

gagern:
@strobelm Please have a look at [the new README](https://github.com/gagern/CindyJS/blob/jsMake/make/README.md), see whether this explains the tool well enough to allow others to maintain it.

kranich:
@gagern I also have some concerns regarding this approach.

If I understand you correctly, the main goal is to make building CindyJS easier for Windows developers. Windows developers do not have Make but could install [Cygwin](http://www.cygwin.com/).

On the other hand, Mac OS and Linux developers usually have Make but not necessarily NodeJS installed. So we are making life easier for Windows developers but potentially harder for Mac OS and Linux developers. Under Linux, NodeJS can of course easily be installed via the package manager. Under Mac OS, installing NodeJS is potentially more difficult (unless the developer uses MacPorts, etc.).

Moreover, I find it more likely that NodeJS changes substantially in the future than Make. Then the build system would have to be adapted to such changes.

Please do not get me wrong! I do appreciate that you want to provide a more welcoming build system for Windows developers. I am also not saying that we should not use the approach you suggest. I just want to make sure that we are aware of its potential drawbacks.

gagern:
@kranich Your concerns are duly noted. Note that in this pull request here, the Makefile has been modified to forward almost all its targets to the JavaScript build system, but the bootstrapping by downloading Node remains in place.

The most serious problem with this forwarding that I can see is that if you run e.g. `make nodetests tests` then up till now, make would know that `tests` depends on `nodetests` and run the latter only once, while now each target gets forwarded independently, leading to multiple execution. `node make nodetests tests` would help there, but it's no big issue either way.

Here are some things which I think the JavaScript based build system does better than what we have with make.

1. **Running multiple targets in a single process.** Things like e.g. the rendering of the markdown from the reference manuale gained a lot of speed by [my most recent commit](https://github.com/gagern/CindyJS/commit/7d956a36c81f21de265db19018ec14b9dcbada24). The Makefile on the other hand had to start up a new node process for each of these. It would be possible to combine the processing of multiple files into a single tool, but the core problem here is that make has some trouble representing rules with more than one output file.

2. **Saving previous settings.** Make has no distinction between variables used to abbreviate things and variables intended to be overridden on the command line. This jsMake however does distinguish these, and can tell when some override is in effect. That's important because it allows us to rebuild exactly those dependencies which are affected by the setting in question.

3. **Data exchange.** If we write multiple tools which can be combined to form more complex workflows, doing so at the JavaScript require level is more natural than at the stdio pipe level. Sure, we could always add yet another file to do the combining, but that might lead to some confusing names.

4. **Break chain of outdated commands.** I haven't implemented this yet, but this is something that has been bugging me for some time. Suppose a change to some general config file requires some target to update, and its update then requires some depending target to update as well. If the second of these updates takes a long time, that is annoying. Often the change in config file will not actually change the output from the first step. So what we could do is compare the output we are about to write with what's already on disk, and if these agree then cut the dependency chain here, allowing depending targets to skip execution unless one of their own inputs is newer than its outputs. I've been wanting something like this in Make pretty often in the past, but I've never found a way to make it work.

Please look at the documentation, look at the code (both old and new) and spare a thought for the above. If after due consideration you still think this approach unwise, I'll instead see what it takes to get builds running from Cygwin or the Git Bash. I'm not fixed on this idea, and if I were the only one comfortable with this change, that would indeed be a bad situation.

kranich:
I am OK with the change. What do @richter-gebert and @kortenkamp think?

@gagern Have you considered using the node.js bindings for ClosureCompiler (see https://github.com/dcodeIO/ClosureCompiler.js)? Then the developer need only install node.js (JRE and ClosureCompiler are installed via the npm package closurecompiler). And you could use the node.js ClosureCompiler API in your build scripts.

gagern:
I have tested neither this nor the one at https://github.com/tim-smart/node-closure. I hadn't expected either of them to get Java on its own. And while a JRE is enough for the closure compiler, I fear GWT will require a full JDK, for which Oracle wants users to manually accept license agreements, thereby discouraging all efforts to automatically install those. Might be worth investigating, but I wouldn't get my hopes too high.

kortenkamp:
It's too late, anyway, but for me the change is ok.

gagern:
@kortenkamp Why would it be too late?

kortenkamp:
ah, misunderstood github – it's not too late. So, I'm ok with this.

strobelm:
Hi, i thought about this again and since the advantages of js_make (better documentation, faster building, platform independence) outreach the drawbacks (less testing/stability) i discussed this again with @kranich and @gagern and we came to terms merging this.

If there are no major concerns (especially from @richter-gebert) i'll merge the code this evening.

gagern:
This conflicts with #142, will rebase shortly.

strobelm:
Here we go, building works for me. Please report any problems in case they occur.

kranich:
>Here we go, building works for me. Please report any problems in case 
>they occur.

No problems, works like a charm! Thank you, @gagern, for implementing this!

gagern:
Thanks for review & merge.

@elkins0, since I assume you are working on Windows, are you interested in trying out building CindyJS on Windows? In case you are, either check for my email from Dec 2 or write a comment here so I can provide further instructions.

elkins0:
Yes I'm using Windows. It will probably take me some time to set things up so that I'm in a position to try it out.

gagern:
@elkins0 Sure, no rush. Let us know when you are ready for testing.

PR #120, opened 2015-12-02
Implementation of ConicParabolaPL

elkins0:
Adds support for the algorithm "ConicParabolaPL" type referenced by the html produced by "Export to CindyJS" from Cinderella. This is one of the items on the #24 list.

The code makes use geoOps._helper.conicFromTwoDegenerates and just like geoOps.ConicFromPrincipalDirections and geoOps.ConicInSquare has one of the degenerate conics comprised of a doubled up line.

### DIFF
--- a/examples/ConstructionParabolaPL.html
+++ b/examples/ConstructionParabolaPL.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>ConstructionParabolaPL.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+    <script type="text/javascript">
+createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0 }, 
+	angleUnit: "°", 
+	geometry: [ 
+		{ name: "A", type: "Free", pos: [ -2.5954198473282437, -4.0, -0.7633587786259541 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "a", type: "FreeLine", pos: [ -0.5192878338278932, -1.85459940652819, 4.0 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "C0", type: "ConicParabolaPL", color: [ 0.0, 0.0, 1.0 ], args: [ "A", "a" ], printname: "$C_{0}$" } ], 
+	ports: [ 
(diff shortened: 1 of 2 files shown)


strobelm:
The code looks fine to me, thanks for implementing! 
Oddly i'm having some strange drawing bugs in the conic routine which haven't been there in other geometric operations. Can anyone reproduce?

strobelm:
Found the culprit! It's me (like always in the conic drawing routine). Fix is coming.

gagern:
I'm a bit unhappy with the `normalizeZ` calls in there. One slight improvement would be rewriting this as

    var C = List.add(List.scalmult(B.value[2], A), List.scalmult(A.value[2], B));

which leads to a null vector instead of a division by zero in case one of these points is at infinity or itself the null vector. (I think we might want to use the above in `geoOps.Mid` as well.) The part about the angular bisectors in the comment made me suspicious as well, but since we have a particularly simple situation here, with `dd` and `dp` orthogonal at infinity and of the same length, this should indeed work as advertised.

Cinderella generates five random tangent lines, but I'm not sure I like that any better than what you have here. I can explain details later on but I'd like to think about this a while longer. In particular see whether one approach can handle cases the other cannot.

gagern:
Ah, I see I was too late, and @strobelm already merged this. Will think about it anyway.

strobelm:
I second that, especially since we agreed that a zero vector is better than some NaNs.

gagern:
We have three special cases to consider:

1. Directrix is the line at infinity
2. Focus lies at infinity
3. Focus coincides with directrix

Experimenting a bit with Cinderella and its spherical view, I'd expect the first two to degenerate to the double line at infinity (i.e. *z*²=0), while the third should result in a double line orthogonal to the directrix at the point of the focus. I think the current implementation should be able to deal with the last case, while the other two can be made to result in null matrices instead with the adjustment I suggested.

Can anyone come up with a way to get my expected result for these as well, without introducing special cases in the code? Is my expectation even correct, or would you expect some different behavior in those cases?

gagern:
I decided to do some computer algebra on this problem. See [my gist about this issue](https://gist.github.com/gagern/5a1d6d4663c3da6f52dd) for a lengthy discussion. The essence is that currently the matrix contains these common factors:

    Fz^4 * (Fx*dx + Fy*dy + Fz*dz)^2 * (dx^2 + dy^2)^4

Since `(Fx,Fy,Fz)` is the focus and `(dx,dy,dz)` the directrix, this covers focus at infinity, directrix at infinity, and focus on directrix. Getting rid of these factors using the current approach seems tricky. The desired resulting matrix would be

```
⎛                         Fz^2*dy^2                          -Fz^2*dx*dy                         -(Fx*dx^2 + Fx*dy^2 + Fz*dx*dz)*Fz⎞
⎜                       -Fz^2*dx*dy                            Fz^2*dx^2                         -(Fy*dx^2 + Fy*dy^2 + Fz*dy*dz)*Fz⎟
⎝-(Fx*dx^2 + Fx*dy^2 + Fz*dx*dz)*Fz   -(Fy*dx^2 + Fy*dy^2 + Fz*dy*dz)*Fz  Fx^2*dx^2 + Fy^2*dx^2 + Fx^2*dy^2 + Fy^2*dy^2 - Fz^2*dz^2⎠
```

I've got half a mind to simply hard-code that matrix into the operation, even though doing so will make the code seem rather obscure. In terms of performance it might even be a win.

Makes me wonder whether we should apply similar techniques to other operations as well, to see what degenerate situations we could avoid. I guess there might be quite a few of those around.

PR #123, opened 2015-12-02
Simplify midpoint computation

gagern:
Makes me wonder why we didn't have it like this all along. Am I missing something?

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -109,22 +109,19 @@ geoOps.Meet.visiblecheck = function(el) {
     el.isshowing = visible;
 };
 
+geoOps._helper.midpoint = function(a, b) {
+    return List.normalizeMax(List.add(
+        List.scalmult(b.value[2], a),
+        List.scalmult(a.value[2], b)));
+};
 
 geoOps.Mid = {};
 geoOps.Mid.kind = "P";
 geoOps.Mid.updatePosition = function(el) {
     var x = csgeo.csnames[(el.args[0])].homog;
     var y = csgeo.csnames[(el.args[1])].homog;
-
-    var line = List.cross(x, y);
-    var infp = List.cross(line, List.linfty);
-    var ix = List.det3(x, infp, line);
-    var iy = List.det3(y, infp, line);
-    var z1 = List.scalmult(iy, x);
-    var z2 = List.scalmult(ix, y);
-    el.homog = List.add(z1, z2);
-    el.homog = List.normalizeMax(el.homog);
-    el.homog = General.withUsage(el.homog, "Point");
+    var res = geoOps._helper.midpoint(x, y);
+    el.homog = General.withUsage(res, "Point");
 };
 
 
@@ -1054,7 +1051,7 @@ geoOps.ConicParabolaPL.updatePosition = function(el) {
     var c = List.cross(B, List.sub(dd, dp));
     var d = List.cross(B, List.add(dd, dp));
     // Midpoint of point A and point B
-    var C = List.add(List.normalizeZ(A), List.normalizeZ(B));
+    var C = geoOps._helper.midpoint(A, B);
     // Line c and line d make up the other degenerate conic
     var vc = List.turnIntoCSList([c]);
     var vd = List.turnIntoCSList([d]);


strobelm:
Hi Martin, i get the idea of the cr variant by @richter-gebert but on first look i can't figure out why your approach works, could you fill me in?

gagern:
It's essentially the `normalizeZ` approach, where you normalize both points then add them. But instead of dividing one point by it's *z* coordinate, I multiply the other by that same coordinate. So the final result will be what you'd get from adding the *z*-normalized coordinates, scaled by the product of the *z* coordinates. Which means you get better behavior at infinity.

strobelm:
Oh i see! Currently i can't see where this could cause trouble, perhaps @richter-gebert could comment on that.

gagern:
OK, I had a look at this algebraically. Given points `A=[ax,ay,az]` and `B=[bx,by,bz]`, the two implementations differ only in a common factor which is absent in my new code but in the existing code evaluates to 

```
ay^2*bx^2 + az^2*bx^2 - 2*ax*ay*bx*by + ax^2*by^2 + az^2*by^2
- 2*ax*az*bx*bz - 2*ay*az*by*bz + ax^2*bz^2 + ay^2*bz^2
```

This can lead to additional degenerate situations. One rather important such example, in my opinion, is if `A` and `B` coincide. Not surprising in hindsight, since `line` would become the null vector. But there are other situations as well, like e.g. the midpoint between `[0,0,1]` and `[1,i,*]`. In the latter case, the cross product between `line` and `infp` is zero, making all the matrices zero as well.

To summarize, I can confirm that my code does not add any new degenerate situations, but does remove quite a number of them. I'd suggest to merge this.

gagern:
In https://github.com/CindyJS/CindyJS/pull/126#issuecomment-161519187 @kortenkamp wrote:

> `C = B[2]*A + A[2]*B # Midpoint of point A and point B` seems strange to me as it is not a homogeneous approach.

How so? The midpoint is not a terribly projective concept to start with, but I'd say `B[2]*A + A[2]*B` should be better than `A/A[2] + B/B[2]`. Sure, you can also fool around with cross ratios and the likes, but that only gives you additional degenerate situations, as outlined above.

strobelm:
I'm sufficiently convinced. Thanks @gagern!

elkins0:
@gagern: Makes me wonder why we didn't have it like this all along. Am I missing something?
Yes. The same technique could be applied to geoOps._helper.pointReflection:

```javascript
geoOps._helper.pointReflection = function(center, point) {
    return List.normalizeMax(List.sub(
        List.scalmult(CSNumber.mult(CSNumber.real(2), point.value[2]), center),
        List.scalmult(center.value[2], point)));
};
```

kortenkamp:
Ah, sorry, I mixed up `A[2]` with `A[1]` (other start index).

gagern:
@elkins0: With the description you suggested, the third coordinate would be subject to the same arithmetic, which is not what we want here. Given center `(Cx,Cy,Cz)` and point `(Px,Py,Pz)` we need the result to be `(2CxPz-PxCz, 2CyPz-PyCz, CzPz)`.

elkins0:
@gagern: I agree, we need the result to be `(2CxPz-PxCz, 2CyPz-PyCz, CzPz)`. I checked geoOps._helper.pointReflection symbolicly and get that result along with a scale factor of `2Cz(CxPx+CyPy)Pz-(Cx²+Cy²)Pz²-Cz²(Px²+Py²)`.

So why do we disagree with what the code does? Here, step-by-step, is what I think it does:

1. `((2 Pz) (Cx, Cy, Cz)) - (Cz (Px, Py, Pz))`
2. `(2 Cx Pz, 2 Cy Pz, 2 Cz Pz) - (Cz Px, Cz Py, Cz Pz)`
3. `(2 Cx Pz - Cz Px, 2 Cy Pz - Cz Py, Cz Pz)`
   which matches what we agree is the result:
4. `(2 Cx Pz - Px Cz, 2 Cy Pz - Py Cz, Cz Pz)`

elkins0:
This technique can also be applied to remove normalizeZ from geoOps.Compass. Here each of the homogeneous coordinates of the three points must be scaled by the other two point's z-value. This avoids the divide by zero caused otherwise by normalizeZ:
```javascript
//TODO Must be redone for Points at infinity
//Original Cindy Implementation is not correct either
geoOps.Compass = {};
geoOps.Compass.kind = "C";
geoOps.Compass.updatePosition = function(el) {
    var m = csgeo.csnames[(el.args[2])].homog;
    var b = csgeo.csnames[(el.args[1])].homog;
    var c = csgeo.csnames[(el.args[0])].homog;
    // Was: m = List.normalizeZ(m);
    // Was: b = List.normalizeZ(b);
    // Was: c = List.normalizeZ(c);
    var mZ = m.value[2];
    var bZ = b.value[2];
    var cZ = c.value[2];
    m = List.scalmult(CSNumber.mult(bZ, cZ), m);
    b = List.scalmult(CSNumber.mult(mZ, cZ), b);
    c = List.scalmult(CSNumber.mult(mZ, bZ), c);
    var diff = List.sub(b, c); // Parallel shift distance from point on circle to center of compass
    var p = List.add(diff, m); // to opposite point on circle of compass
    p = List.normalizeMax(p); // Was: List.normalizeZ(p);

    var matrix = geoOps._helper.CircleMP(m, p);
    matrix = List.normalizeMax(matrix);
    el.matrix = General.withUsage(matrix, "Circle");
};
```

gagern:
> So why do we disagree with what the code does?

Because I somehow made a mistake evaluating this in my head. Sign error I guess. Thanks for insisting, @elkins0.

> … a scale factor of `2Cz(CxPx+CyPy)Pz-(Cx²+Cy²)Pz²-Cz²(Px²+Py²)`.

If you set `Cx=Cy=0, Cz=1` this simplifies to `Px²+Py²=0` which is a degenerate conic consisting of two complex lines, which intersect in the center *C* itself. So I would assume that this prefactor has no geometric relevance for purely real constructions, but it is relevant for us both because we want to take complex detours and because any additional factor can have an adverse effect on the numeric stability of the computation.

> This technique can also be applied to remove normalizeZ from geoOps.Compass

I'll welcome any pull request which can get rid of a `normalizeZ` call like this. We might also want to check whether there are any additional common factors remaining after this simplification. If not, I think we can drop the comment up front with the TODO. If there is something left, we could try to get rid of that as well.

PR #124, opened 2015-12-02
Create hopefully persistent anchor names

gagern:
While marked tries to provide anchor names for all its headings, these names
tend to focus on the word parts of the heading, ignoring symbolic content.
Changing the descriptive part of a heading line will change the anchor.  So
we override that, and try to come up with a scheme which will hopefully make
our anchors remain unaffected by many common forms of edits.

In particular, a normal CindyScript command has an anchor composed from the
name of the command and its arity.  An infix operator has its command
derived from the codepoint of its symbol(s).  Some other constructs like the
one for absolute value use `_` as a placeholder for arguments and unicode
codepoints for the non-argument symbols.  Only the first of these is really
easily readable by humans, but all of these should remain pretty unaffected
by changes to the non-essential parts of the headline.

### DIFF
--- a/ref/js/md2html.js
+++ b/ref/js/md2html.js
@@ -79,8 +79,8 @@ MyRenderer.prototype.heading = function(text, level, raw) {
   }
   if (!id && (match = /`‹\w*›\s*(\S+)\s*‹\w*›`/.exec(raw))) {
     // infix operators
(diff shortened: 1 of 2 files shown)


gagern:
Since the step from #142 to this is just one tiny commit, I'll take the liberty of merging this myself.

PR #126, opened 2015-12-03
Simplify ConicParabolaPL to avoid degenerate situations

gagern:
This makes the algorithm simpler insofar as it eliminates a huge common factor of the resulting matrix, thus removing unneccessary singularities. Reading geometric meaning back from the code is now tricky, though.

The following sage code snippet may serve as a documentation of how the previous algorithm by @elkins0 got turned into this new one.

```py
PR.<Fx,Fy,Fz,dx,dy,dz> = QQ[] # A Polynomial ring over the rationals with six indeterminates
A = vector([Fx,Fy,Fz]) # Focus point
a = vector([dx,dy,dz]) # Directrix line
linfty = vector(PR, [0,0,1]) # The line at infinity
dd = a.cross_product(linfty) # The direction of the directrix line
dp = dd.cross_product(linfty) # The perpendicular direction of the directrix line
e = A.cross_product(dd) # Parallel line through focus point in the direction of the directrix line forms both lines of one degenerate conic
b = A.cross_product(dp) # Line through focus point in the perpendicular direction of directrix line
B = a.cross_product(b) # Point B is the projection of the focus point onto the directrix line
c = B.cross_product(dd - dp) # Create both both angle bisector lines through point B from the
d = B.cross_product(dd + dp) # direction of the directrix and its perpendicular direction
C = B[2]*A + A[2]*B # Midpoint of point A and point B
deg1 = c.column() * d.row() # One degenerate conic with factors c and d
deg2 = e.column() * e.row() # Another with e as a double line
m = (C*deg1*C)*deg2 - (C*deg2*C)*deg1 # Pick from pencil to coincide with C
m += m.transpose() # Symmetrize
m/gcd(m.list())/2 # Print the simplified form, as seen in the source comment
```

For details see https://github.com/CindyJS/CindyJS/pull/120#issuecomment-161315538 and https://gist.github.com/gagern/5a1d6d4663c3da6f52dd.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1041,6 +1041,9 @@ geoOps.ConicParabolaPL.updatePosition = function(el) {
      *  [-(Fx*dx^2 + Fx*dy^2 + Fz*dx*dz)*Fz,
      *   -(Fy*dx^2 + Fy*dy^2 + Fz*dy*dz)*Fz,
      *   Fx^2*dx^2 + Fy^2*dx^2 + Fx^2*dy^2 + Fy^2*dy^2 - Fz^2*dz^2]]
+     * For derivation see https://github.com/CindyJS/CindyJS/pull/126
+     * or http://math.stackexchange.com/a/1557496/35416
+     * or https://gist.github.com/gagern/5a1d6d4663c3da6f52dd
      */
     var mult = CSNumber.mult;
     var neg = CSNumber.neg;
@@ -1066,11 +1069,7 @@ geoOps.ConicParabolaPL.updatePosition = function(el) {
     var zz = sub(
         mult(add(mult(Fx, Fx), mult(Fy, Fy)), dx2pdy2),
         mult(Fz2, mult(dz, dz)));
-    var m = List.turnIntoCSList([
-        List.turnIntoCSList([xx, xy, xz]),
-        List.turnIntoCSList([xy, yy, yz]),
-        List.turnIntoCSList([xz, yz, zz])
-    ]);
+    var m = geoOps._helper.buildConicMatrix([xx, xy, yy, xz, yz, zz]);
     m = List.normalizeMax(m);
     el.matrix = General.withUsage(m, "Conic");
 };
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1033,32 +1033,46 @@ geoOps.ConicBy1p4l.updatePosition = function(el) {
 geoOps.ConicParabolaPL = {};
 geoOps.ConicParabolaPL.kind = "C";
 geoOps.ConicParabolaPL.updatePosition = function(el) {
-    var A = csgeo.csnames[(el.args[0])].homog; // focus point
-    var a = csgeo.csnames[(el.args[1])].homog; // directrix line
-    // The direction of the directrix line
-    var dd = List.cross(a, List.linfty);
-    // The perpendicular direction of the directrix line
-    var dp = List.cross(dd, List.linfty);
-    // Parallel line through focus point in the direction of the directrix line
-    // forms both lines of one degenerate conic
-    var e = List.cross(A, dd);
-    // Line through focus point in the perpendicular direction of directrix line
-    var b = List.cross(A, dp);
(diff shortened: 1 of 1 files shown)


kortenkamp:
I did not check all the details, but `C = B[2]*A + A[2]*B # Midpoint of point A and point B` seems strange to me as it is not a homogeneous approach.

gagern:
@kortenkamp: This discussion of how to compute a midpoint might be better located at #123 since it's the core of the request there but just a minor point here.

strobelm:
Hi! The code troubles me a bit, simply because i can't read the math behind it anymore. I would rather like to have (bug free) readable code and not go down on the algebra level if there are no performance issues involved. Or as a middle ground the sage code as comment.

gagern:
The Sage code is by no means the most obvious way to arrive at this formula. I just found essentially the same starting from the elementary definition of a parabola in terms of equal distance to focus and directrix. For details please read [here on Math Stack Exchange](http://math.stackexchange.com/a/1557496/35416). I think any detailed explanation would be too long and therefore hard to read in the 80-columns fixed-width setup of that source file. Could we instead some links to provide that information? Links to [this pull request](https://github.com/CindyJS/CindyJS/pull/126), [my gist](https://gist.github.com/gagern/5a1d6d4663c3da6f52dd), the above [Stack Exchange post](http://math.stackexchange.com/a/1557496/35416), or all three of them? I think this is far easier to read in a browser.

elkins0:
I have checked the code. It does implement the correct matrix. The matrix can also be split this way

```
          ⎛    0      0   -Fx*Fz  ⎞       ⎛  dy²   -dx*dy -dx*dz  ⎞
(dx²+dy²)*⎜    0      0   -Fy*Fz  ⎟ + Fz²*⎜ -dx*dy  dx²   -dy*dz  ⎟
          ⎝ -Fx*Fz -Fy*Fz Fx²+Fy² ⎠       ⎝ -dx*dz -dy*dz -dz²    ⎠
```


The equation for the conic can be written for (x y 1) as:
```
(dx²+dy²)*((Fx²+Fy²)-2*Fz*(Fx*x+Fy*y))-Fz²*(dz²-(dy*x-dx*y)²+2*dz*(dx*x+dy*y))==0
```
I do have a concern. Try this construction. It specifically tests the condition where the focus point lies on the directrix. With my implementation the result is truly chaotic giving incorrect results even with the midpoint fix. With the matrix implementation, the correct result is shown about half the time. The rest of the time nothing is shown. Drag the focus point and spin the directrix line to see the trace flicker on and off. In the initial position no trace is shown:

```javascript
	geometry: [ 
		{ name: "A", type: "Free", pos: [ 1, 0.3636363636363636, 0.2272727272727273 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "a", type: "Through", pos: [ -0.16666666666666666, -0.16666666666666666, 1.0 ], args: [ "A" ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
        { name: "C0", type: "ConicParabolaPL", color: [ 0.0, 0.0, 1.0 ], args: [ "A", "a" ], printname: "$C_{0}$" } ], 
```

elkins0:
The result matrix could be setup using:
```javascript
    var m = geoOps._helper.buildConicMatrix([xx, xy, yy, xz, yz, zz]);
```
instead of
```javascript
    var m = List.turnIntoCSList([
        List.turnIntoCSList([xx, xy, xz]),
        List.turnIntoCSList([xy, yy, yz]),
        List.turnIntoCSList([xz, yz, zz])
    ]);
```

gagern:
`buildConicMatrix` might come at a slight performance cost, but it's much shorter so I switched to it. I also added some comments as I suggested above. I can reproduce the flickering rendering in the degenerate case @elkins0 reported. But I'm fairly certain that this is only a rendering issue. @strobelm can you please have a closer look?

strobelm:
Hi, i can confirm the flickering in the branch. A patch (https://github.com/CindyJS/CindyJS/commit/927e87cd702d2f48693a1f5acded7ff70d413774) is already in master, the branch needs only a rebase onto master.

strobelm:
I think i can live with the documentation online, although comments in the sources are usually more sturdy.

PR #128, opened 2015-12-03
Use adjoint instead of inverse

gagern:
If a scalar multiple is of no effect, using the adjoint matrix instead of the inverse will speed up some operations, since it avoids a division by the determinant.  It may also avoid some numeric problems, possibly leading to all-zero objects instead of NaN or Infinity for certain situations.

The occurrence in `OpDrawing` is the only one not immediately followed by a renormalization step, so that one might warrant the closest examination.

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -422,7 +422,7 @@ eval_helper.drawconic = function(aConic, modifs) {
 
     var ttMatrix = General.mult(tMatrix2, tMatrix1); // get transformation matrix
 
-    var ittMatrix = List.inverse(ttMatrix);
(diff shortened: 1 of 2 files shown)


strobelm:
Hi! Getting rid of inverse is always a good idea. With the adjoint i ran into drawing bugs in examples/65c_Conic_2P_3l.html, adding a normalization fixes this, but now examples/66_Conic_5lines.html went into degenerate situations too early. 

Perhaps we should rethink how we transform the conic from cindy to canvas coordinates, this always felt not really right.

strobelm:
Hi, i reverted back the inverse in conic drawing, rebased the branch and merged it directly into master. Thanks @gagern for finding this. Closed.

PR #129, opened 2015-12-04
Replaces geoOps.ConicInSquare with PT version

elkins0:
This code implements geoOps.ConicInSquare using a projective transformation of the four corners of a square whose four sides are tangent to a unit circle into the four points given. The transformation is applied to the matrix of that unit circle resulting in the desired conic which is then tangent to the lines through those given points in the same manner as the square and the circle.

This is a direct translation of the sage code by @garern shown at [ConicInSquare.md](https://gist.github.com/gagern/9fdd5175933ce6a5e557).

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1146,19 +1146,19 @@ geoOps.ConicInSquare.updatePosition = function(el) {
             List.turnIntoCSList([z, a[1], z]),
             List.turnIntoCSList([z, z, a[2]])
         ]);
-    };
+    }
     // Compute projective transformation from basis to given points
     function ptstep(a, b, c, d) {
         var m = List.transpose(List.turnIntoCSList([a, b, c]));
         var v = List.productMV(List.adjoint3(m), d);
         return List.productMM(m, diagonalMatrix3(v));
-    };
+    }
     // Compute projective transformation from first set of points to second
     function pt(a1, b1, c1, d1, a2, b2, c2, d2) {
         var m1 = ptstep(a1, b1, c1, d1); // first set of points
         var m2 = ptstep(a2, b2, c2, d2); // second set of points
         return List.productMM(m2, List.adjoint3(m1));
-    };
+    }
     var A = csgeo.csnames[(el.args[0])].homog;
     var B = csgeo.csnames[(el.args[1])].homog;
     var C = csgeo.csnames[(el.args[2])].homog;
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1142,12 +1142,7 @@ geoOps.ConicInSquare.updatePosition = function(el) {
     var C = csgeo.csnames[(el.args[2])].homog;
     var D = csgeo.csnames[(el.args[3])].homog;
     // Compute projective transformation from basis to given points (A, B, C, D)
-    var m1T = List.transpose(List.turnIntoCSList([A, B, C]));
-    var v = List.productMV(List.adjoint3(m1T), D).value;
-    // Setup a 3x3 diagonal matrix using the 3 element vector result.
-    var z = CSNumber.zero;
-    var v1dm = geoOps._helper.buildConicMatrix([v[0], z, v[1], z, z, v[2]]);
-    var m1 = List.productMM(m1T, v1dm);
+    var m1 = eval_helper.basismap(A, B, C, D);
     // Compute projective transformation from basis to the corners of a square
     // tangent to a unit circle combined with applying this to the unit circle
     // matrix. The pre-computed constant result scaled by 1/16 is created here.
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1137,43 +1137,30 @@ geoOps._helper.coHarmonic = function(a1, a2, b1, b2) {
 geoOps.ConicInSquare = {};
 geoOps.ConicInSquare.kind = "C";
 geoOps.ConicInSquare.updatePosition = function(el) {
-    // Setup a 3x3 diagonal matrix using a 3 element vector
-    function diagonalMatrix3(v) {
-        var z = CSNumber.zero;
-        var a = v.value;
-        return List.turnIntoCSList([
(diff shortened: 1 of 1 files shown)


elkins0:
@gagern: Not much to it. Works as advertised. I used Cinderella to make the same projective transformation with no difficulty. The export to CindyJS gives the message "TransformC not implemented yet" which is I guess what you meant by "once we can apply projective transformations to conics". Of course that is just what we did here in this coding.

Everyone else, please discuss the code before merging. I am curious where the weak spots were in the original coding in #116 in terms of test cases. I suppose they likely involve points at infinity. I will check back tomorrow.

gagern:
> I am curious where the weak spots were in the original coding in #116 in terms of test cases.

Well, the extra factors in your original implementation are essentially the four determinants [A,B,D], [A,B,C], [B,C,D], [A,C,D] which are zero if these three points are collinear. So your code would lead to an all-zero matrix and hence an ill-defined conic if any three of the input points are collinear.

On the other hand, a projective transformation which attemots to map three collinear points onto non-collinear ones should, in my understanding, be undefined as well. Nevertheless, I don't get a clean zero matrix for those situations out of my approach.

I guess I should investigate this some more before we merge this. I should understand whether the results we get from the projective transformation steps in such collinear configurations have any valid gemetric meaning.

gagern:
It appears that the resulting matrix in the case of three collinear points does indeed describe the line of that collinearity, as a double-line. Which makes a lot of sense geometrically, since it's the limit case if one converges to this situation. In the process of establishing this, I also verified that the resulting matrix remains the same even if the four points are rotated cyclically, or reversed in order. So it would seem to be well-behaved.

I still don't fully understand *why* this works. The key point here is that a transformation matrix of rank 1 contains more information than I had realized. If I can describe this in a nice way, I'll write another comment or gist with details on the math here.

kortenkamp:
@gagern The rank 1 matrix can be decomposed into u v^t. So it gives you two lines... (Just a quick thing while being in a meeting, so please excuse if this is too short or wrong)

gagern:
Some further improvement in terms of performance: If you look at the code (again in sage notation since that's easier to read), you see that essentially the resulting matrix is

```py
m1.adjoint().transpose() * m2.transpose() * diagnonal_matrix([1,1,-1]) * m2 * m1.adjoint()
#                          \__________________ const. ____________________/
```

The constant expression here is the result of ConicInSquare with the projective basis (1,0,0), (0,1,0), (0,0,1) and (1,1,1) as input. Its matrix is (after canceling a common factor of 16):

```
⎛ 1  1 -3⎞
⎜ 1  1  1⎟
⎝-3  1  1⎠
```

You can hard-code that using [`List.realMatrix`](https://github.com/CindyJS/CindyJS/blob/2422bed69c2cd2d74c7f68454a799e3817dd02da/src/js/libcs/List.js#L68-L75), then you don't need the [`pt`](https://github.com/elkins0/CindyJS/blob/de2d950436966facba1cdd402702fb35673eb6d9/src/js/libgeo/GeoOps.js#L1156-L1161) function (which is similar to [`evaluator.map$8`](https://github.com/CindyJS/CindyJS/blob/927e87cd702d2f48693a1f5acded7ff70d413774/src/js/libcs/Operators.js#L3267-L3284) except that the latter is designed to be called from CindyScript, not just internally) any more. You don't really need [`ptstep`](https://github.com/elkins0/CindyJS/blob/de2d950436966facba1cdd402702fb35673eb6d9/src/js/libgeo/GeoOps.js#L1150-L1155) either, since it's functionally equivalent to [`eval_helper.basismap`](https://github.com/CindyJS/CindyJS/blob/927e87cd702d2f48693a1f5acded7ff70d413774/src/js/libcs/Operators.js#L3254-L3265). And once you build on that, you have eliminated both applications of [`diagonalMatrix`](https://github.com/elkins0/CindyJS/blob/de2d950436966facba1cdd402702fb35673eb6d9/src/js/libgeo/GeoOps.js#L1140-L1149) so you might delete that as well. Although that might make a good addition to `List.js` and a nice CindyScript function as well, but that's a different story.

gagern:
@kortenkamp: A rank 1 matrix taken as a conic would be a pair of lines, yes. In case it's symmetric, you'd even have *u* = *v* i.e. a double line, which is indeed the outcome here in the degenerate situations under discussion. But he rank 1 matrix I was referring to is an intermediate step, where that matrix represents a projective transformation. Which in my previous naive understanding just maps the whole plane to a single point.

But nevertheless, such a matrix decomposes, too. If you read it as (*u* ∙ *v*^T), then it means you can read *u* as a point and *v* as a line. Any point in the plane will get mapped to (some representative of) *u*, except for points on line *v* which will get mapped to the null vector. So if you apply that transformation to a point and then plug the resulting point into an arbitrary conic, you can be sure that you get zero if the point happened to be on  line *v*. And in the case at hand, *v* would be exactly the line given by the three collinear points.

elkins0:
@gagern: Oh goody. Now this time the build fails and looks like its not my fault. I'm finished anyway.

eval_helper.basismap could be changed to not use List.inverse:

```javascript
eval_helper.basismap = function(a, b, c, d) {
    var mat = List.turnIntoCSList([a, b, c]);
    mat = List.inverse(List.transpose(mat));
    var vv = General.mult(mat, d);
    mat = List.turnIntoCSList([
        General.mult(vv.value[0], a),
        General.mult(vv.value[1], b),
        General.mult(vv.value[2], c)
    ]);
    return List.transpose(mat);

};
```

and instead use List.adjoint3 (General has also been replaced):

```javascript
eval_helper.basismap = function(a, b, c, d) {
    var mat = List.turnIntoCSList([a, b, c]);
    mat = List.adjoint3(List.transpose(mat));
    var v = List.productMV(mat, d).value;
    mat = List.turnIntoCSList([
        List.scalmult(v[0], a),
        List.scalmult(v[1], b),
        List.scalmult(v[2], c)
    ]);
    return List.transpose(mat);
};
```

gagern:
The inversion in basismap is to be addressed in #128. The build error Travis reported looks really strange. Seems something was seriously wrong with that run. I simply restarted it, and it's happy now.

PR #130, opened 2015-12-05
implemented moebius transformation

strobelm:
Hi!

i implemented Möbius transformations. With this pull request we can calculate the transformation matrix by 3 homogeneous coordinates and use this to transform points. Lines are not supported currently, i'll implement them later.

### DIFF
--- a/examples/111_Moebius.html
+++ b/examples/111_Moebius.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+
+// Initialization code, executed once up front.
+
+</script>
+<script id="csdraw" type="text/x-cindyscript">
+
+// Drawing code, executed whenever the canvas gets redrawn.
+
+</script>
+<script type="text/javascript">
+
+var cdy = createCindy({ // See ref/createCindy documentation for details.
+  ports: [{id: "CSCanvas"}],
+  scripts: "cs*",
+  language: "en",
+  snap:true,
+  grid:1,
+  defaultAppearance: {
+    // See GeoBasics.js for possible attributes.
+  },
+  geometry: [
+    {name:"A1", type:"Free", pos:[0,5], color:[255,0,0], labeled:true},
+    {name:"A2", type:"Free", pos:[0,3], color:[255,0,0], labeled:true},
+    {name:"A3", type:"Free", pos:[0,0], color:[255,0,0], labeled:true},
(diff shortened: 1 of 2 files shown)


gagern:
Unless I'm completely missing something, you are using real-valued arithmetic here, completely ignoring the imaginary parts of the inputs. That's a no-go, since it would break complex tracing.

I'd prefer to reserve the `Tr` kind for projective transformations, and use some other name for Möbius transformations. This will become particularly important once we verify signatures for geometric operations. How about `Mo` or `Mt`?

I guess I'd prefer to store `mat1` and `mat2` with the transformation object, instead of recomputing them as part as the transformation operation.

strobelm:
Concerning the complex values: its somewhat problem inherent since the Möbius transform lives in CP^1 and not in CP^2. I'll think about it (or ask @richter-gebert ).

Renaming the projection and storing the matrices is fine to me.

gagern:
You can break all complex arithmetic down to real arithmetic in carthesian representation. You know, things like

* `(a + i*b) + (c + i*d) = (a + c) + i*(b + d)`
* `(a + i*b) * (c + i*d) = (a*c - b*d) + i*(a*d + b*c)`

Then you can use computations like these even in situations where `a,b,c,d` themselves are complex numbers. The most tricky part is the division, since you want to avoid having a complex number in the denominator. But the code you have already takes care of that aspect.

Instead of speaking about elementary arithmetic operations, it might make sense to talk about matrices instead. You can represent a complex number as a matrix:

```
a + i*b ↦ ⎛a -b⎞
          ⎝b  a⎠
```

Then you can build block matrices of these. Computing a generic 4×4 inverse or adjoint is no fun, but since you know that it's essentially 2×2 with these blocks as entries, it should be possible to describe the adjoint more easily.

Note that this representation is almost [the matrix representation for quaternions](https://en.wikipedia.org/wiki/Quaternion#Matrix_representations) except for a few complex conjugations. I wonder whether doing these conjugations as well leads to different results. I guess I should try this out in my CAS, and come back with the results so we can compare the three approaches (elementary arithmetic á la Cinderella, simple conjugation-free matrices and proper quaternions).

kranich:
>Instead of speaking about elementary arithmetic operations, it might 
>make sense to talk about matrices instead. You can represent a complex 
>number as a matrix:
>
>```
>a + i*b ↦ ⎛a -b⎞
>          ⎝b  a⎠
>```

Beware that the matrix representation of complex numbers is numerically 
less stable than complex number arithmetic, sometimes catastrophically 
so! Due to floating point errors, a matrix representing a complex number 
may cease to have the special form

⎛a -b⎞
⎝b  a⎠

(For the same reason, quaternions are numerically more stable than 
rotation matrices.)

-- 
Stefan Kranich

gagern:
I've posted a follow-up at #138. If that one gets merged, this one here should get merged implicitely.

PR #131, opened 2015-12-05
Remove normalizeZ from geoOps.Compass

elkins0:
Using normalizeZ on the homogeneous coordinates of the three points produces `(m/mZ, b/bZ, c/cZ)`, scaling these by `mZ * bZ * cZ` leaves `(bZ * cZ * m, mZ * cZ * b, mZ * bZ * c)` instead and removes the possiblity of a divide_by-zero.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -656,12 +656,17 @@ geoOps.Compass.updatePosition = function(el) {
     var m = csgeo.csnames[(el.args[2])].homog;
     var b = csgeo.csnames[(el.args[1])].homog;
     var c = csgeo.csnames[(el.args[0])].homog;
-    m = List.normalizeZ(m);
-    b = List.normalizeZ(b);
-    c = List.normalizeZ(c);
+    // Scale each point's homogeneous coordinates by the other two
+    // point's z-value to allow addtion and subtraction to be valid.
+    var mZ = m.value[2];
+    var bZ = b.value[2];
+    var cZ = c.value[2];
+    m = List.scalmult(CSNumber.mult(bZ, cZ), m);
+    b = List.scalmult(CSNumber.mult(mZ, cZ), b);
+    c = List.scalmult(CSNumber.mult(mZ, bZ), c);
     var diff = List.sub(b, c);
     var p = List.add(diff, m);
-    p = List.normalizeZ(p);
+    p = List.normalizeMax(p);
 
     var matrix = geoOps._helper.CircleMP(m, p);
     matrix = List.normalizeMax(matrix);
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -637,44 +637,44 @@ geoOps.CircleMr.updatePosition = function(el) {
     var r = getStateComplexNumber();
     putStateComplexNumber(r); // copy param
     var m = csgeo.csnames[(el.args[0])].homog;
-    m = List.normalizeZ(m);
-    var p = List.turnIntoCSList([r, CSNumber.zero, CSNumber.zero]);
-    p = List.add(p, m);
-    var matrix = geoOps._helper.CircleMP(m, p);
-    matrix = List.normalizeMax(matrix);
+    var rs = CSNumber.mult(m.value[2], r); // scale the radius
+    var matrix = geoOps._helper.ScaledCircleMrr(m, CSNumber.mult(rs, rs));
     el.matrix = General.withUsage(matrix, "Circle");
     el.radius = r;
 };
 geoOps.CircleMr.stateSize = 2;
 
 
+// M is center and rr is radius squared; both must be pre-scaled
+geoOps._helper.ScaledCircleMrr = function(M, rr) {
+    var x = M.value[0];
+    var y = M.value[1];
+    var mz = CSNumber.neg(M.value[2]); // minus z
+    var v = List.scalmult(mz, List.turnIntoCSList([x, y, mz])).value;
+    var vxy = List.turnIntoCSList([x, y]);
+    var zz = CSNumber.sub(List.scalproduct(vxy, vxy), rr);
+    var matrix = geoOps._helper.buildConicMatrix([v[2], CSNumber.zero, v[2], v[0], v[1], zz]);
+    return List.normalizeMax(matrix);
+};
+
+
 geoOps.Compass = {};
 geoOps.Compass.kind = "C";
 geoOps.Compass.updatePosition = function(el) {
-    var m = csgeo.csnames[(el.args[0])].homog;
+    var a = csgeo.csnames[(el.args[0])].homog;
     var b = csgeo.csnames[(el.args[1])].homog;
-    var c = csgeo.csnames[(el.args[2])].homog;
+    var m = csgeo.csnames[(el.args[2])].homog;
     // Scale each point's homogeneous coordinates by the other two
     // point's z-value to allow addtion and subtraction to be valid.
-    var mZ = m.value[2];
+    var aZ = a.value[2];
     var bZ = b.value[2];
-    var cZ = c.value[2];
-    m = List.scalmult(CSNumber.mult(bZ, cZ), m);
-    b = List.scalmult(CSNumber.mult(mZ, cZ), b);
-    c = List.scalmult(CSNumber.mult(mZ, bZ), c);
-    var cX = c.value[0]; // bZ*cX*mZ
-    var cY = c.value[1]; // bZ*cY*mZ
-    cZ = c.value[2]; // bZ*cZ*mZ
-    var z2 = CSNumber.abs2(cZ);
-    var xz = CSNumber.neg(CSNumber.mult(cX, cZ));
-    var yz = CSNumber.neg(CSNumber.mult(cY, cZ));
-    var dXY = List.turnIntoCSList([cX, cY]);
-    var sqXY = List.scalproduct(dXY, dXY);
-    var dR = List.sub(b, m);
-    var sqR = List.scalproduct(dR, dR);
-    var zz = CSNumber.sub(sqXY, sqR);
-    var matrix = geoOps._helper.buildConicMatrix([z2, CSNumber.zero, z2, xz, yz, zz]);
-    matrix = List.normalizeMax(matrix);
+    var mZ = m.value[2];
+    a = List.scalmult(CSNumber.mult(bZ, mZ), a);
+    b = List.scalmult(CSNumber.mult(aZ, mZ), b);
+    m = List.scalmult(CSNumber.mult(aZ, bZ), m);
+    // Setup circle's matrix with m as center and segment ab length as radius
+    var d = List.sub(b, a);
+    var matrix = geoOps._helper.ScaledCircleMrr(m, List.scalproduct(d, d));
     el.matrix = General.withUsage(matrix, "Circle");
 };
 
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -637,17 +637,19 @@ geoOps.CircleMr.updatePosition = function(el) {
     var r = getStateComplexNumber();
     putStateComplexNumber(r); // copy param
     var m = csgeo.csnames[(el.args[0])].homog;
-    var rabs2 = CSNumber.abs2(r).value.real;
-    var rs = m.value[2]; // Already scaled when r/r=1 is scaled by m
-    if (rabs2 > 1) {
-        // Like homog [1=r/r, 0, 1/r]; Scale m by 1/r; rs already scaled
-        // Note: CSNumber.inv(CSNumber.real(Infinity)) produces NaN - i*NaN
-        m = List.scalmult(isFinite(rabs2) ? CSNumber.inv(r) : CSNumber.zero, m);
-    } else {
-        // Like homog [r, 0, 1]; Scale r by m.value[2]; m already scaled by 1
-        rs = CSNumber.mult(rs, r);
-    }
-    var matrix = geoOps._helper.ScaledCircleMrr(m, CSNumber.mult(rs, rs));
+    /*
+    The circle's radius value may take on values from zero to infinity.
+    However since the squared radius value appears in the circle's matrix,
+    a radius value of 2E+154 or more could also end up as an infinite value.
+    The use of List.normalizeMax everywhere limits coordinate values of m to
+    no more 1.0, so scaling the radius value by m's z-coordinate first will
+    not make the radius value any larger. Then by squaring the radius value,
+    any infinity value produced can be caught here.
+    */
+    var sr = CSNumber.mult(m.value[2], r);
+    var sr2 = CSNumber.mult(sr, sr);
+    if (!CSNumber._helper.isFinite(sr2)) return List.fund;
+    var matrix = geoOps._helper.ScaledCircleMrr(m, sr2);
     el.matrix = General.withUsage(matrix, "Circle");
     el.radius = r;
 };
@@ -656,6 +658,14 @@ geoOps.CircleMr.stateSize = 2;
 
 // M is center and rr is radius squared; both must be pre-scaled
 geoOps._helper.ScaledCircleMrr = function(M, rr) {
+    /*
+    Given M as the circle's homogeneous center point coordinates [x, y, z] and
+    rr as the circle's radius value squared scaled by M's z-coordinate squared,
+    build the following matrix:
+        ⎛   z*z      0      -z*x   ⎞
+        ⎜    0      z*z     -z*y   ⎟
+        ⎝  -z*x    -z*y  x*x+y*y-rr⎠
+    */
     var x = M.value[0];
     var y = M.value[1];
     var mz = CSNumber.neg(M.value[2]); // minus z
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -648,8 +648,6 @@ geoOps.CircleMr.updatePosition = function(el) {
 geoOps.CircleMr.stateSize = 2;
 
 
-//TODO Must be redone for Points at infinity
-//Original Cindy Implementation is not correct either
 geoOps.Compass = {};
 geoOps.Compass.kind = "C";
 geoOps.Compass.updatePosition = function(el) {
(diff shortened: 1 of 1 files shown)


gagern:
Definitely a step in the right direction, but we're not there yet. This code still has `mZ * bZ * cZ` as a common factor. If either of these were zero, then we'd expect the double line at infinity as the result, not the zero matrix. The `bZ * cZ` term can be avoided if you pass the original unscaled centerpoint to `CircleMP`. The `mZ` is harder to avoid. Looking at the Sage version of `CircleMP`:

```py
p = mZ*(cZ*b - bZ*c) - bZ*cZ*m
tang = matrix([[-mZ, 0, mX], [0, -mZ, mY], [mX, mY, 0]])
fund = diagonal_matrix([0,0,1])
mu = p*tang*p
la = p*fund*p
erg = mu*fund - la*tang
gcd([mu, la]) # prints mZ
```

So the last coordinate of `p` already has a factor of `mZ` in it. The `tang` matrix will multiply the other coordinates with `mZ` as well, while the `fund` matrix will discard the other coordinates completely. Now all we need is a nice formulation to avoid that factor. We could hardcode the polynomial for `mu`, but that's hardly elegant, so I'll wait a bit whether someone can come up with something better.

elkins0:
Here's what I'm thinking. Redo symbolically using divide then scale to see what we end up with. I haven't this for mistakes yet :smiley: :

```
m = (mX, mY, mZ)
b = (bX, bY, bZ)
diff = b/bZ - m/mZ = (bX/bZ - mX/mZ, bY/bZ - mY/mZ, 0)
squaredRadius = (bX/bZ - mX/mZ)^2 + (bY/bZ - mY/mZ)^2
newCircle = (x - cX/cZ z)^2 + (y - cY/cZ z)^2 - z^2 squaredRadius
newCircle = (x - cX/cZ z)^2 + (y - cY/cZ z)^2 - z^2 ( (bX/bZ - mX/mZ)^2 + (bY/bZ - mY/mZ)^2) )

Must be scaled by (cZ*bZ*mZ)^2 to avoid divide-by-zero. Expanded, factored I get:

newCircle = 
bZ^2 cZ^2 mZ^2 x^2 + bZ^2 cZ^2 mZ^2 y^2 - 2 bZ^2 cX cZ mZ^2 x z - 2 bZ^2 cY cZ mZ^2 y z + 
 z^2 ( bZ^2 (cX^2 + cY^2) mZ^2 - cZ^2 ((-bZ mX + bX mZ)^2 + (-bZ mY + bY mZ)^2) )

matrix =
( (bZ^2 cZ^2 mZ^2, 0, -bZ^2 cX cZ mZ^2),
  (0, bZ^2 cZ^2 mZ^2, -bZ^2 cY cZ mZ^2), 
  (-bZ^2 cX cZ mZ^2, -bZ^2 cY cZ mZ^2, bZ^2 (cX^2 + cY^2) mZ^2 - cZ^2 ((-bZ mX + bX mZ)^2 + (-bZ mY + bY mZ)^2) )
```

elkins0:
The code has been updated. See @ef591e7e48823b9e0211c29ca44a50849ba5c75f for changes and [98_Compass.html](https://github.com/CindyJS/CindyJS/blob/master/examples/98_Compass.html) for test file.

elkins0:
After consideration I find that the use of abs2 is inappropriate here.
`CSNumber.abs2(a)` is equivalent to `CSNumber.mult(a, CSNumber.conjugate(a))`.
`CSNumber.mult(a, a)` needs to be used.
So `List.scalproduct(a, a)` replaces `List.abs2(a)` in two places.

Vector length `abs` and angle `arg` normally are used together in place of `re` and `im`.
There is no occurrence of `conjugate` nor any `arg` value to balance the use of `abs` in the code.
`CSNumber.arg` would appear like this:
```javascript
CSNumber.arg = function(a) {
    return CSNumber.real(Math.atan2(CSNumber.im(a), CSNumber.re(a)));
};
```
Curiously, there is a List.normSquared which appears to be a stunted version of List.abs2.

Also noted that  zero radius circles are sometimes drawn as a line #139.
For `type: "CircleMr" radius: 0` this seems to be always true.

gagern:
I checked the code. The algebra almost works out; you still have one call to `abs2` which you need to replace by a multiplication. In general it's best to avoid any reference to complex conjugation since code involving that has a tendency to be non-analytic, which can break complex tracing.

I think the new naming scheme is quite confusing. With the old names, *B* and *C* play exchangeable roles; it's only the difference between which matters. *M*, on the other hand, has a different role since it's the center of the new circle.

The other variables didn't convey much meaning to me at first reading. With the comment above, for comparison, things improve somewhat. On the whole, I'm of two minds about the code. On the one hand, the previous code was far more high-level, so in my opinion it was easier to understand what's going on. But on the other hand, the low-level approach (i.e. dealing with individual complex numbers instead of whole vectors and matrices) might be more efficient. And we don't seem to have a high-level approach which can avoid the extra factor.

I think I'll wait a bit for the `abs2` to be replaced, and also see whether anyone can come up with something more elegant. But if not, I think we should merge this in a short while.

elkins0:
I replaced the leftover abs2--that was an oversight on my part. I fixed the naming scheme to be segment end-points a and b and m for the center. I created geoOps._helper.ScaledCircleMrr  so it could also use it use it to remove List.normalizeZ(m) from geoOps.CircleMr.updatePosition which will now also produce a pair of lines at infinity when m is a point at infinity. List.normalizeZ is also used in geoOps.CircleMr.getParamForInput but there appears to be no use case for a divide by zero happening there. If the center point m z-value is 0 then there's no way to stretch the radius when the circle is a line at infinity and the pointer position can't be at point at infinity...

elkins0:
So I thought I'd see what happens when the radius is a large value in geoOps.CircleMr.updatePosition. I find that with the radius set to 1E+160 .. 1E+300 then geoOps._helper.ScaledCircleMrr produces a matrix with the last element set to "-Infinity". If radius is set to Infinity then geoOps._helper.ScaledCircleMrr  produces a matrix with the last element set to "NaN - i*NaN". geoOps.Compass can also have its segment length have values like 1E+160. Perhaps the following should be added as the very first code line in geoOps._helper.ScaledCircleMrr:
```javascript
if (!CSNumber._helper.isFinite(rr)) return List.fund;
```

elkins0:
So the problem stems from the code I added to geoOps.CircleMr:
```javascript
    var rs = CSNumber.mult(m.value[2], r); // scale the radius
    var matrix = geoOps._helper.ScaledCircleMrr(m, CSNumber.mult(rs, rs));
```
Since the r value has a scale of 1.0, m is unchanged. Lets say r is `Infinity+i*0` and m is `[1+i*0, 0.75+i*0, 0.025+i*0]` then we get rs is `Infinity+i*NaN`. This is what happens as the code progresses:

1. `CSNumber.mult(m.value[2], r)`
2. `(0.025+i*0)*(Infinity+i*0)`
3. `((0.025*Infinity)+(0*0))+i*((0*0)+(0*Infinity))`
4. returns `Infinity+i*NaN`

This is because `0*Infinity` is `NaN`. If the radius when Infinity were respesnted as a homogeneous coordinate of `[1, 0, 0]` and otherwise as `[radius, 0, 1]` then all would well. Then the code for geoOps.CircleMr could be as follows:
```javascript
    var mZ = m.value[2];
    var rZ = r.value[2];
    var rs = List.scalmult(mZ, r);
    var ms = List.scalmult(rZ, m);
    var matrix = geoOps._helper.ScaledCircleMrr(ms, List.scalproduct(rs, rs));
```
The deeper problem of course is that CSNumber can not handle Infinity without introducing NaN values. These can happen as `2E+154 * 2E+154` is Infinity when radius squared is produced.

elkins0:
OK, so I've created a separate commit @ with only the changes necessary for geoOps.Compass.updatePosition. Next I'll put in the additional geoOps.CircleMr.updatePosition changes to handle center point at infinity and/or radius set to Infinity all whilst avoiding the dreaded divide-by-zero.

elkins0:
There appears to be nothing else that creates homogeneous coordinates with numbers greater than 1. So I took that approach within `geoOps.CircleMr.updatePosition`. If the radius is greater than 1 then it is scaled to 1 by `1/radius` and then scaled by the z-value of m while m is scaled by `1/radius`. Otherwise we use the radius as is and scale it by the z-value of m and leave m alone (scaled by 1).

We have to handle the radius specially when it is set to infinity because `CSNumber.inv(CSNumber.real(Infinity))` produces `NaN - i*NaN` instead of the desired `CSNumber.zero`.

The result is that the radius looks like the homogeneous coordinate [radius, 0, 1] when it is less than or equal to 1, [1, 0, 1/radius] when it is greater than 1, and [1, 0, 0] when it is set to infinity. Looked at this way it has its own z-value which we can use to scale m.

`geoOps.CircleMr.updatePosition` now no longer has divide-by-zero occurring and produces a matrix equal to List.fund when the center point is at infinity or the radius is set to infinity. When both the conditions are true a matrix of all zeroes is produced.

elkins0:
@gagern, @strobelm this is ready for review and merging also, I think.

gagern:
@elkins0 Sorry, I've been busy with the upcoming CindyJS website today. Will be on vacation from tomorrow on, so I don't know when I'll find time to review this. Don't worry, the ticket system won't allow me to forget about this, so I'll pick it up again after the holidays at the latest.

gagern:
First the `CircleMr` portion. It took me quite a while to grok what's going on here. Which means that this may be bad for maintainability. In particular, the term “prescaled” in the `ScaledCircleMrr` helper is not clear enough in my opinion. If I read things correctly, you require `rr = (M[2] * r)^2`. Better make that clear. The `Like homog […]` are unclear to me as well. *What* is like that?

Here is how I currently think about this: `ScaledCircleMrr` requires `M = [x, y, z]` and `rr = (z*r)^2`. Since you always have `rr = rs^2` you want `rs = z*r`. In the second case, that's exactly what you compute. In the first case, you divide everything by `r`, so you pass `M = [x/r, y/r, z/r], rs = z`. Fine. The fact that you use `1` as a cutoff avoids concerns that you might be inappropriately dealing with just one specific special case, which in turn makes this kind of case distinction somewhat more valid in my eyes. On the other hand, it means you'll be doing the division in more cases than it would really be needed. I'm not sure yet how I feel about that. I also was worried for a moment that you imply that if `isFinite` returns false, then the value must be infinite. But this is indeed correct here, since the only other alternative, NaN, will end up in the second case. Perhaps a comment to this effect would be useful.

Compared to `CircleMr`, the changes to `Compass` look reasonably clear and correct. In particular, I verified once again that the resulting matrix is as simple as it gets, in terms of entry polynomials. Of course this, too, builds on `ScaledCircleMrr` so really understanding `Compass` requires an understanding of its arguments, and would benefit from a clarification there.

@elkins0: Right now I'd like you to improve the comment for `ScaledCircleMrr`, clarifying its `rr` argument. I'd also welcome improvements to the comments in the `CircleMr` case distinctions.

After that I'm happy with the `Compass` part, but would like to hear from @strobelm, @kranich or @richter-gebert whether it's just me who now has trouble understanding `CircleMr`, and whether we can come up with something that is easier to understand and maintain while still dealing with all the inputs your code allows.

To that effect, what are the actual benefits of the new code? As far as I can tell, it allows for center at infinity (due to the dropped `normalizeZ`) and for infinite radius (due to the `isFinite` case distinction). The semantics of infinite radius are as follows: for finite center you get the double line at infinity, while for center at infinity you get the zero matrix. Which is exactly what I'd expect. Somewhat more surprisingly, for center at infinity and finite (or even zero) radius, we get the double line at infinity. At least for real centers; choosing the ideal circle points as center will lead to a zero matrix.

Personally I consider handling center at infinity to be far more important than infinite radius, so I wouldn't rule out sacrificing the infinite radius case for the sake of code clarity.

elkins0:
>>The result is that the radius looks like the homogeneous coordinate [radius, 0, 1] when it is less than or  equal to 1, [1, 0, 1/radius] when it is greater than 1, and [1, 0, 0] when it is set to infinity. Looked at this  way it has its own z-value which we can use to scale m.

>The Like homog […] are unclear to me as well. *What* is like that?

I was simply trying to say that the radius scale factor is 1.0 and that if the radius were scaled by 1/radius then the scaled radius value would be 1.0 and I would then need to scale M by 1/radius. Otherwise the the radius would be scaled by M's z-coordinate and the M would remain unchanged since it would be scaled by 1.0. Never mind though, I was to close to the action, trying to keep the radius value under 1.0.

With regards to `geoOps.CircleMr.updatePosition`:
>On the other hand, it means you'll be doing the division in more cases than it would really be needed.

Actually on further consideration there is no need to divide at all. Since `List.normalizeMax` makes sure that all of m's coordinate values are no greater than 1.0 there is no reason to scale m at all since scaling the radius by m's z-coordinate would not make it larger; rather it would more likely be much smaller. Then the only thing left to do is to check the scaled radius value squared for being finite.

I have made those changes and added the additional comments as @gagern suggested.

With regards to `geoOps._helper.ScaledCircleMrr`, note that `geoOps._helper.CircleMP` could be replaced by it (but `geoOps._helper.ScaledCircleMrr` also does a  `List.normalizeMax` ):
```javascript
geoOps._helper.CircleMP = function(m, p) {
    var mZ = m.value[2];
    var pZ = p.value[2];
    m = List.scalmult(pZ, m);
    p = List.scalmult(mZ, p);
    var d = List.sub(p, m);
    return geoOps._helper.ScaledCircleMrr(m, List.scalproduct(d, d));
};
```

gagern:
> Actually on further consideration there is no need to divide at all. Since `List.normalizeMax` makes sure that all of m's coordinate values are no greater than 1.0 there is no reason to scale m at all since scaling the radius by m's z-coordinate would not make it larger; rather it would more likely be much smaller. Then the only thing left to do is to check the scaled radius value squared for being finite.

I'm writing down some thoughts in this paragraph. The post above and the comment in your latest commit reads as if the code now relies on its input having been passed through `normalizeMax`. There is no such normalization inside the code, and although normally all other operations *should* have performed such renormalization, I'd try to not rely too much on this. To be more precise, I think it is OK if lack of normalization makes numeric problems more likely or more severe, but apart from that should not affect the geometric meaning. If I read it correctly, then a large m.z coordinate combined with a large radius could lead to the double line at infinity even in cases where a finite representation might be possible through renormalization. I think this is acceptable.

The fact that you no longer have the infinite radius case as a special subcase of a more generic “scale by `1/r`” case makes it less obvious that this is the correct result. I hope anyone questioning that will eventually have a look at the git history (which I'd prefer to preserve in this case) or this discussion here, and be convinced.

A more serious problem due to the missing case distinction is the scenario where `sr2` is NaN (in at least one component). So perhaps we should handle the `if (!CSNumber._helper.isNaN(sr2))` case first. I'm somewhat unsure what to return in that case, though. We could return the all-zero matrix, indicating an undefined conic. We could also build a matrix with NaN entries, either everywhere or everywhere except for the mixed xy term which is zero for circles. Building such a matrix would be considerably more code, though, and right now I can't think of a good reason for doing this. So I'd say return the all-zero matrix if `sr2` is NaN.

> With regards to `geoOps._helper.ScaledCircleMrr`, note that most of what `geoOps._helper.CircleMP` does could be replaced by a call to it as shown here […]

Don't do that. The code you propose introduces an additional `mZ` factor into the resulting matrix. As a consequence, the result would no longer be defined (i.e. would be the zero matrix) if the center is a point at infinity. In the current implementation, the result in that case factors into the line at infinity and the line through P in the direction orthogonal to M.

elkins0:
As you point out, in the case where m lacks the expected normalization, the geometric meaning is preserved anyway so it is left up to the upstream code to insure that renormalization occurs in order to get the best numerical results. That appears to be a fundamental assumption throughout all of the `GeoOps.js` code.

With regards to `geoOps.CircleMr.updatePosition`:
>A more serious problem due to the missing case distinction is the scenario where `sr2` is NaN (in at least one component). So perhaps we should handle the `if (!CSNumber._helper.isNaN(sr2))` case first.

JavaScript's `isFinite` function returns `false` if the number is `NaN` or positive or negative infinity. So `CSNumber._helper.isFinite`, shown here for reference and used by `geoOps.CircleMr.updatePosition`, already detects the case where `sr2` is a `NaN`. Right now `geoOps._helper.ScaledCircleMrr` will return the double line at infinity if `sr2` is a `NaN`.
```javascript
CSNumber._helper.isFinite = function(z) {
    return isFinite(z.value.real) && isFinite(z.value.imag);
};
```
Let us reach a consensus as to what should be returned when `sr2` is a `NaN` before coding this any further. In the original coding, `List.normalizeZ(m)` was introducing `NaN`'s whenever m.z was zero. Now `NaN`'s are no longer introduced anywhere in this code. Let's keep it that way. Note that `NaN` is not referenced or deliberately introduced anywhere else in `GeoOps.js` so why should matters be further complicated here? The function `List._helper.isNaN` is unused outside of `tracing2core`, `tracing4core` and `tracingSesq`. Our goal should be to avoid computations like `Infinity-Infinity`, `Infinity/Infinity` or `0/0` and their ilk which produce a `NaN`.  

>Don't do that. The code you propose introduces an additional `mZ` factor into the resulting matrix. As a consequence, the result would no longer be defined (i.e. would be the zero matrix) if the center is a point at infinity. In the current implementation, the result in that case factors into the line at infinity and the line through P in the direction orthogonal to M.

Understood. I wasn't planning to.

gagern:
The idea of NaN in IEEE float semantics is that it taints anything computed from it. So if at *any* stage of a computation, something becomes undefined, then that undefinedness propagates through the computation and makes the final result undefined. At least as long as you only use elementary arithmetic. A user might write a script which does `C.radius = f(…)`, and that function `f` could do *anything*. (I don't think we have a writing `radius` accessor yet, but we should have one day.) We should make no assumption about radius *not* being NaN.

If the input is undefined, then returning the well-defined double line at infinity is wrong. Which is why the current behavior is wrong. Even if we *could* guarantee *at the moment* that the input will never be NaN, such a guarantee has to be represented in code, otherwise someone is bound to unknowingly violate it during future development. So we must make sure that this function, all by itself, doesn't do something wrong for input which is properly typed but outside the expected domain.

The goal should be that for undefined (i.e. NaN) radius, the result must be geometrically undefined. We have various ways to represent an undefined conic, though: we could use the all-zero matrix, or a matrix with at least one NaN entry. Both of these *should* propagate, i.e. using a zero or NaN geometric object as the input to some construction should again lead to a zero or NaN output, with “zero or none” seen as a set. So I think it is valid to turn NaN input into zero output.

In https://github.com/elkins0/CindyJS/commit/424d739c3b9a3bc8e1aa00860b5255f7954bd440 you had a check `if (rabs2 > 1)`. Any comparison involving NaN is false (even `NaN == NaN`), so NaN input would end up in the other case, and you'd do `rs = CSNumber.mult(rs, r)` and eventually plug the square of this `rs` into the matrix. So you'd have a matrix with at least one `NaN` entry, which makes the whole conic undefined. That was OK, I think.

elkins0:
>So you'd have a matrix with at least one NaN entry, which makes the whole conic undefined. That was OK, I think.

To accomplish that only requires this simple code change:
```javascript
-    if (!CSNumber._helper.isFinite(sr2)) return List.fund;
+    if (!CSNumber._helper.isFinite(sr2) && !CSNumber._helper.isNaN(sr2)) return List.fund;
```

gagern:
That would be all right.

elkins0:
@gagern: I've made that change.

PR #137, opened 2015-12-07
Have splitDegenConic return a pair of null vectors instead of nada

### LINKED ISSUE #132: geoOps._helper.IntersectConicConic fails given 2 zeroed matricies
This happens because `if (Aabs2 < myeps) { // both are degenerate` is true causing each matrix to be sent directly to `geoOps._helper.splitDegenConic` which then returns Nada.

The result is "SCRIPT5007: Unable to get property 'value' of undefined or null reference" a little latter  on and so loading stops.

Here is a way to get two zeroed matrices (2 sets of coincident points are used):
```javascript
	geometry: [ 
		{ name: "A", type: "Free", pos: [ 4.0, 4.0, 0.8 ], color: [ 0.0, 0.0, 1.0 ], labeled: true }, 
		{ name: "B", type: "Free", pos: [ 4.0, 4.0, 0.8 ], color: [ 0.0, 0.0, 1.0 ], labeled: true }, 
		{ name: "C", type: "Free", pos: [ 4.0, 4.0, 0.8 ], color: [ 0.0, 0.0, 1.0 ], labeled: true }, 
		{ name: "D", type: "Free", pos: [ 3.2, -4.0, -0.8 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "E", type: "Free", pos: [ 3.2, -4.0, -0.8 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "F", type: "Free", pos: [ 3.2, -4.0, -0.8 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "C0", type: "ConicFromPrincipalDirections", color: [ 0.0, 0.0, 1.0 ], args: ["A", "B", "C"], printname: "$C_{0}$" },
		{ name: "C1", type: "ConicFromPrincipalDirections", color: [ 1.0, 0.0, 0.0 ], args: ["D", "E", "F"], printname: "$C_{1}$" },
		{ name: "Collection__3", type: "IntersectionConicConic", args: [ "C0", "C1" ] } ], 
```

gagern:
While `nada` would be appropriate for CindyScript operations, in the geometry core we expect to continue with any intermediate result without further type checking.  So we should always return a pair of vectors, even if we make them null vectors to indicate undefined geometric elements.

This fixes #132.

### DIFF
--- a/examples/111_VeryDegenerateICC.html
+++ b/examples/111_VeryDegenerateICC.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script type="text/javascript">
+
+var cdy = createCindy({ // See ref/createCindy documentation for details.
+  ports: [{id: "CSCanvas"}],
+  language: "en",
+  geometry: [ 
+    { name: "A", type: "Free", pos: [5,5,1], color: [0,0,1], labeled: true }, 
+    { name: "B", type: "Free", pos: [5,5,1], color: [0,0,1], labeled: true }, 
+    { name: "C", type: "Free", pos: [5,5,1], color: [0,0,1], labeled: true }, 
+    { name: "D", type: "Free", pos: [3,-4,1], color: [1,0,0], labeled: true }, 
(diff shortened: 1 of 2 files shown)


strobelm:
nada was actually always wrong there, good to have it fixed!

PR #138, opened 2015-12-08
Doubly complex Möbius transformations

gagern:
This is a follow-up to #130. There are extensive comments in the code indicating what's going on.

@kranich wrote in https://github.com/CindyJS/CindyJS/pull/130#issuecomment-162884813:

> Beware that the matrix representation of complex numbers is numerically less stable than complex number arithmetic, sometimes catastrophically so!

This is still somewhat confusing to me. I thought (and still think) that the elementary low-level arithmetic operations one would need to perform with this matrix representation are *exactly* the same ones one would also perform when doing complex arithmetic the usual way. Can you give a more explicit example, or point out some reference on this subject? Or can you in the example at hand point out any part of the code where a more elementary approach would lead to a significantly different computation at the floating-point level and therefore might conceivably lead to better numeric stability?

Since the pull request now involves code by both me and @strobelm, I'd welcome a review by a knowledgeable third party. In the light of the concerns expressed above preferably @kranich I think.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1570,6 +1570,115 @@ geoOps.SelectL.updatePosition = function(el) {
     el.homog = General.withUsage(el.homog, "Line");
 };
 
+geoOps.TrMoebius = {};
+geoOps.TrMoebius.kind = "Tr";
+geoOps.TrMoebius.updatePosition = function(el) {
+    // input RP^2 coordinates mapped to complex numbers
+    // tranform A1, A2, A3, to B1, B2, B3
+
+    var A1 = List.normalizeZ((csgeo.csnames[el.args[0]]).homog);
+    var B1 = List.normalizeZ((csgeo.csnames[el.args[1]]).homog);
+
+    var A2 = List.normalizeZ((csgeo.csnames[el.args[2]]).homog);
+    var B2 = List.normalizeZ((csgeo.csnames[el.args[3]]).homog);
+
+    var A3 = List.normalizeZ((csgeo.csnames[el.args[4]]).homog);
+    var B3 = List.normalizeZ((csgeo.csnames[el.args[5]]).homog);
+
+    // map A1.x, A1.y to A1.x + i*A1.y ...
+    var a1 = CSNumber.complex(A1.value[0].value.real, A1.value[1].value.real);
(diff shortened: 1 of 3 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -1570,6 +1570,149 @@ geoOps.SelectL.updatePosition = function(el) {
...
+    /*
+    Now we conceptually want B * A.adjoint()
+
+    ⎛ B0 -B1  B4 -B5⎞ ⎛ A6 -A7 -A4  A5⎞   ⎛ ar -ai  br -bi⎞
+    ⎜ B1  B0  B5  B4⎟ ⎜ A7  A6 -A5 -A4⎟ = ⎜ ai  ar  bi  br⎟
+    ⎜ B2 -B3  B6 -B7⎟ ⎜-A2  A3  A0 -A1⎟   ⎜ cr -ci  dr -di⎟
+    ⎝ B3  B2  B7  B6⎠ ⎝-A3 -A2  A1  A0⎠   ⎝ ci  cr  di  dr⎠
+
+    But since we only care about two columns of the result, it's
+    enough to use two columns of the adjoint of A, namely the second
```
kranich: Should be 'enough to use two columns of the adjoint of A, namely the **first**'

gagern:
I just verified that the resulting point has a gcd of 1. So there should be no easily avoidable degeneracies.

kranich:
@gagern The operations that you are actually performing using complex numbers in matrix representation should be fine.

kranich:
If you trust the maths in [GeoOps.js#L1691](https://github.com/gagern/CindyJS/blob/CalcMoebius/src/js/libgeo/GeoOps.js#L1691) then the pull request can be merged.

strobelm:
Great! Since i wrote a prove for exactly GeoOps.js#L1691 we are good to go.

gagern:
Yes, I've verified [the statement in that comment](https://github.com/CindyJS/CindyJS/blob/2df669c75d1539e744ce304409079a57e2eaa542/src/js/libgeo/GeoOps.js#L1687-L1692) as well.

gagern:
N.b.: I just compared the implementation merged here against the original Cinderella implementation. Except for the fact that Cinderella normalizes its input to z=1 and we don't, the two implementations are algebraically equivalent.

PR #140, opened 2015-12-08
Include Moebius transformations in generic Transform macro

gagern:
We need this since Cinderella exports `Transform` operations for projective transformations and Möbius transformations alike, but we use different operations to implement these cases so we need more points to delegate to.

I already anticipated the acceptal of this request in my recent edit to #24. Once it's in line, the delegation to `TrMoebiusP` can be checked as done, since I verified that this works.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -7,7 +7,8 @@ geoOps._helper = {};
  * S  - Segment
  * C  - Conic (including circle)
  * *s - Set of *
- * Tr - Transformation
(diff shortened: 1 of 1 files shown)


strobelm:
Oh i missed that request, if we merge this we will have to resolve a conflict with https://github.com/strobelm/CindyJS/commit/9a18562377a0eae184550817feffafded9f036c4

Actually i like the version with the object map even better than my switch case variant.

gagern:
If you want to, you can merge this and rebase #141 onto this.

Issue #139, opened 2015-12-08
Zero radius circles sometimes drawn as a line


elkins0:
Example geometry:
```javascript
	geometry: [ 
		{ name: "A", type: "Free", pos: [ 4.0, 3.0, 1.0 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
		{ name: "C0", type: "CircleMr", color: [ 0.0, 0.0, 1.0 ], args: [ "A" ], radius: 0, size: 1, printname: "$C_{0}$" } ], 
```

strobelm:
Hi,

yes it's a known problem, since the circle ops use conic by 5 which for
small radius circles yields a degenerate conic in the conic drawing
routine. I'll fix that asap.

On 12/08/2015 07:49 PM, Alexander Elkins wrote:
>
> Example geometry:
>
>     geometry: [ 
>         { name: "A", type: "Free", pos: [ 4.0, 3.0, 1.0 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
>         { name: "C0", type: "CircleMr", color: [ 0.0, 0.0, 1.0 ], args: [ "A" ], radius: 0, size: 1, printname: "$C_{0}$" } ], 
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CindyJS/CindyJS/issues/139>.
>


-- 
Michael Strobel, M.Sc.

Technische Universität München -- Zentrum Mathematik
Lehrstuhl für Geometrie und Visualisierung (M10)
Boltzmannstr. 3, 85747 Garching bei München 
https://www-m10.ma.tum.de/bin/view/Lehrstuhl/MichaelStrobel

PR #141, opened 2015-12-09
Möbius transform for arcs and lines

strobelm:
Hi after points here is the Möbius transform for Lines and Arcs. Feel free to comment on the pull request.

The last thing missing are conics, Cinderella treats them as circles (as far as i can tell), seems a bit odd to me. Any input on that?


gagern:
> The last thing missing are conics, Cinderella treats them as circles (as far as i can tell), seems a bit odd to me. Any input on that?

We shouldn't allow Möbius transformations of conics which are not circles, since the result will in general be quite complicated. Might revisit that if we ever have algebraic curves of higher degree as primitive elements, but not before. If you want to, you can try to work out what degree exactly we're talking about here, but in general I'd expect degree 4. That's because the Möbius transformation is already quadratic in the input coordinates, and plugging that into a quadratic form will lead to degree 4 unless by some lucky coincidence the highest degrees cancel out somehow.

src/js/libgeo/GeoOps.js
```diff
@@ -1724,12 +1737,68 @@ geoOps.TrMoebiusP.kind = "P";
...
+};
+
+geoOps.TrMoebiusL = {};
+geoOps.TrMoebiusL.kind = "C";
+geoOps.TrMoebiusL.updatePosition = function(el) {
+    var t = csgeo.csnames[(el.args[0])];
+    var l = csgeo.csnames[(el.args[1])].homog;
+
+    var getRandLine = function() {
+        var rline = List.realVector([100 * (Math.random() - 0.5), 100 * (Math.random() - 0.5), 100 * (Math.random() - 0.5)]);
```
gagern: Why the factor 100? I don't see that affecting the geometric meaning in any way, and I can't imagine it improving numerics either.

src/js/libgeo/GeoOps.js
```diff
@@ -1724,12 +1737,68 @@ geoOps.TrMoebiusP.kind = "P";
...
+
+    el.homog = geoOps._helper.TrMoebiusP(p, t.mat1, t.mat2);
     el.homog = General.withUsage(el.homog, "Point");
 };
 
+geoOps._helper.TrMoebiusP = function(p, mat1, mat2) {
+    var l1 = List.productMV(mat1, p);
+    var l2 = List.productMV(mat2, p);
+    return List.normalizeMax(List.cross(l1, l2));
+};
```
gagern: I'd suggest you pass in the Möbius transformation as a complete object, instead of passing the two matrices in separately. Everyone calling this code will be holding both in any case, so splitting things up only here means the rest of the code won't have to care about how a Möbius transformation is actually performed internally.

src/js/libgeo/GeoOps.js
```diff
@@ -1724,12 +1737,68 @@ geoOps.TrMoebiusP.kind = "P";
...
+    var a3 = List.cross(getRandLine(), l);
+
+    var Pts = geoOps._helper.TrMoebiusPts(t, a1, a2, a3);
+
+    el.matrix = List.normalizeMax(geoOps._helper.ConicBy5(null, Pts[0], Pts[1], Pts[2], List.ii, List.jj));
+    el.matrix = General.withUsage(el.matrix, "Circle");
+};
+
+geoOps.TrMoebiusArc = {};
+geoOps.TrMoebiusArc.kind = "C";
```
gagern: Can you try to get this hooked up to the Cinderella export via `geoMacros`?

src/js/libgeo/GeoOps.js
```diff
@@ -1211,19 +1211,32 @@ geoOps.CircleBy3 = {};
...
     var erg = geoOps._helper.ConicBy5(el, a, b, c, d, p);
     el.matrix = List.normalizeMax(erg);
     el.matrix = General.withUsage(el.matrix, "Circle");
 
 };
+geoOps.CircleBy3.initialize = function(el) {
+    el.startPoint = csgeo.csnames[(el.args[0])].homog;
+    el.viaPoint = csgeo.csnames[(el.args[1])].homog;
+    el.endPoint = csgeo.csnames[(el.args[2])].homog;
+};
```
gagern: Why do we store this information with a circle that is not an arc? Seems like we shouldn't need it, and it might tempt some future contribution to use it inappropriately if we are not careful enough.

strobelm:
Here is the update: i fixed a bug in geoOps.RandomLine which prevented to have lines in the first sector (and removed the factor), added Moebius to geoMacros.Transform and prepared for further transformation and incorporated most of @gagern suggetions, thanks for that! 

Comments are very welcome.

src/js/libgeo/GeoOps.js
```diff
@@ -1211,10 +1211,14 @@ geoOps.CircleBy3 = {};
...
 geoOps.CircleBy3.updatePosition = function(el) {
     var a = csgeo.csnames[(el.args[0])].homog;
+    el.startPoint = a;
     var b = csgeo.csnames[(el.args[1])].homog;
+    el.viaPoint = b;
     var c = List.ii;
     var d = List.jj;
     var p = csgeo.csnames[(el.args[2])].homog;
+    el.endPoint = p;
+
```
gagern: I'd still prefer to have these points not stored for a normal circle. Unless you have a good reason to do otherwise.

strobelm:
Just forgot about them, should be fixed now!

strobelm:
By the way Cinderella generates complete garbage when applying Möbius tranforms to conics, perhaps somebody with an account could open a bugreport (or even better create one for me ;-) ).

strobelm:
And here is the rebased version.

gagern:
> Cinderella generates complete garbage when applying Möbius tranforms to conics, perhaps somebody with an account could open a bugreport

Created issue 109 for this.

src/js/libgeo/GeoOps.js
```diff
@@ -7,8 +7,7 @@ geoOps._helper = {};
  * S  - Segment
  * C  - Conic (including circle)
  * *s - Set of *
- * Tr - Projective transformation
- * Mt - Moebius transformation
+ * Tr - Transformation
```
gagern: Looks like something went wrong here during rebase. Or is this deliberate?

src/js/libgeo/GeoOps.js
```diff
@@ -1725,12 +1735,68 @@ geoOps.TrMoebiusP.kind = "P";
...
+    var Arc = csgeo.csnames[(el.args[1])];
+
+    var a1 = Arc.startPoint;
+    var a2 = Arc.viaPoint;
+    var a3 = Arc.endPoint;
+
+    var Pts = geoOps._helper.TrMoebiusPoints(t, a1, a2, a3);
+    var b1 = Pts[0],
+        b2 = Pts[1],
+        b3 = Pts[2];
```
gagern: Makes me wonder about the use of `TrMoebiusPoints`. Why have a function which combines the three values to a single array, just to decompose that array again here? Right now, including the content of `TrMoebiusPoints` here would actually save you one line. Things would be different in ES2015 and its destructuring assignments, I guess.

src/js/libgeo/GeoOps.js
```diff
@@ -1928,14 +2018,23 @@ geoMacros.Calculation = function(el) {
...
+};
+
 geoMacros.Transform = function(el) {
-    var trKind = csgeo.csnames[el.args[0]].kind;
-    var argKind = csgeo.csnames[el.args[1]].kind;
+    var arg = csgeo.csnames[el.args[1]];
+    var tr = csgeo.csnames[el.args[0]];
+    // workaround for Arcs since we treat them as circles
+    var akind = arg.kind;
+    if (arg.kind === "C" && arg.type === "ArcBy3") akind = "Arc";
```
gagern: Better use `arg.isArc` here, so you can repeatedly apply Möbius transformations to a segment or arc. I think you could drop the `kind === "C"` check and just check the `isArc` property alone.

Perhaps we should consider adding a kind for arcs, if and when we add a kind for circles. These things would be simpler if we had a type hierarchy here, so that we could add subtypes without having to change type checks in a million places. We might want to try the performance impact of such a change one day.

src/js/libgeo/GeoOps.js
strobelm: No it wasn't, reverted that.

src/js/libgeo/GeoOps.js
strobelm: You're right, i just wanted to avoid code doubling. Sometimes i really miss calling function by reference (without the object detour). I will upload a fix in a minute.

src/js/libgeo/GeoOps.js
strobelm: Good point, i think we need  ```arg.type === "ArcBy3" || arg.isArc``` since the macros are evaluated before initialization takes place.

I thought about introducing a new type, but since cinderella never had real use cases despite ArcBy3 i abandoned the idea.

gagern: > … the macros are evaluated before initialization takes place.

Then how about moving the `isArc` property from the geometric element to the geometric operator? So you can do `geoOps[arg.type].isArc`. There definitely should be a test case for repeated transformations, since this is more tricky than it would seem at first glance.

By the way, #144 will make sure that every element gets fully initialized before any subsequent macro gets expanded.

strobelm: Moving to geoOps isn't a bad idea, but we would have to keep the
el.isArc property since GeoRender.js will only have the element handed
over and can't distinguish circles and arcs anymore in case we remove this.

Perhaps our best shot is #144. I already read the code, but i still have
to run some examples.
 
On 12/12/2015 04:04 PM, Martin von Gagern wrote:
>>  geoMacros.Transform = function(el) {
>> -    var trKind = csgeo.csnames[el.args[0]].kind;
>> -    var argKind = csgeo.csnames[el.args[1]].kind;
>> +    var arg = csgeo.csnames[el.args[1]];
>> +    var tr = csgeo.csnames[el.args[0]];
>> +    // workaround for Arcs since we treat them as circles
>> +    var akind = arg.kind;
>> +    if (arg.kind === "C" && arg.type === "ArcBy3") akind = "Arc";
>> … the macros are evaluated before initialization takes place.
> Then how about moving the `isArc` property from the geometric element to the geometric operator? So you can do `geoOps[arg.type].isArc`. There definitely should be a test case for repeated transformations, since this is more tricky than it would seem at first glance.
>
> By the way, #144 will make sure that every element gets fully initialized before any subsequent macro gets expanded.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/CindyJS/CindyJS/pull/141/files#r47432651


-- 
Michael Strobel, M.Sc.

gagern: You could easily use the same check using `geoOps` over in ` GeoRender.js` as well. But if this is a motivation to go for #144, I'll be more than happy to see that proceed.

strobelm:
Here is the rebased version with create merged. Since the initialization problem is gone it's easier to detect arcs.

elkins0:
Alone these appear to work fine. However with TrInverse the geometry won't load:

```javascript
geoOps.TrInverse = {};
geoOps.TrInverse.kind = "Tr";
geoOps.TrInverse.updatePosition = function(el) {
    var tr = csgeo.csnames[(el.args[0])];
    var m = tr.matrix;
    el.dualMatrix = List.transpose(tr.matrix); <--
    el.matrix = List.transpose(tr.dualMatrix);
};

List.transpose = function(a) {
    var erg = [];
    var n = a.value[0].value.length; <--
Unable to get property 'value' of undefined or null reference
```
You have setup mat1 and mat2. However matrix and dualMatrix are undefined.

strobelm:
Hi,

yes this is true, simply because inversion for Möbiustransformations
isn't that easy. For inverting a transform (az+b)/(cz+d) we have to
compute (dz -b)/(-cz+a) and therefor recalculate mat1 and mat2. This
needs it's own function.


Am 15.12.2015 um 22:43 schrieb Alexander Elkins:
>
> Alone these appear to work fine. However with TrInverse the geometry
> won't load:
>
> geoOps.TrInverse = {};
> geoOps.TrInverse.kind = "Tr";
> geoOps.TrInverse.updatePosition = function(el) {
>     var tr = csgeo.csnames[(el.args[0])];
>     var m = tr.matrix;
>     el.dualMatrix = List.transpose(tr.matrix); <--
>     el.matrix = List.transpose(tr.dualMatrix);
> };
>
> List.transpose = function(a) {
>     var erg = [];
>     var n = a.value[0].value.length; <--
> Unable to get property 'value' of undefined or null reference
>
> You have setup mat1 and mat2. However matrix and dualMatrix are undefined.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/141#issuecomment-164906234>.
>

-- 
Michael Strobel, M.Sc.

gagern:
We should also discuss how to handle this from Cinderella. The Cinderella representation has a tendency to use the same name for different but related operations. Like `Transform` for transforming “an object”, no matter what object that is, or what transformation. But also `TrInverse` which it apparently uses to invert both projective and Möbius transformations. The rub here is that we are already using that name, exclusively for projective transformations.

So I think we have these options:

1. Change the name Cinderella exports. We have name-changing machinery in place, and we at some point decided to use the more descriptive names, which in this case and in my opinion means differentiating by type of transformation. Should we change other names as well, to cover the full set of CindyJS transformation primitives? I think it would be consistent to do so.

2. Change the name of what's now called `TrInverse` (to something like `ProjectiveInverse` or whatever) so that we can make `TrInverse` a macro. The renaming is neccessary since macro expansion is deliberately recursive, so we can't have the same name for a macro and a primitive.

3. Write some special case code. I'd rather avoid this unless there is some *very* good reason.

I think I'd go with 1. @kortenkamp, what's your opinion here?

examples/111_Moebius_CinderellaExportTest.html
```diff
@@ -0,0 +1,61 @@
...
+		{ name: "R", type: "Free", pos: [ 4.0, 2.5891783567134263, 0.20040080160320642 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "S", type: "Free", pos: [ 4.0, 2.0696055684454757, 0.2320185614849188 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "Tr1", type: "TrProjection", color: [ 0.0, 0.0, 1.0 ], args: [ "L", "S", "M", "R", "N", "Q", "O", "P" ] }, 
+		{ name: "C2", type: "Transform", color: [ 0.0, 0.0, 1.0 ], args: [ "Tr1", "C0" ], size: 1, printname: "$C_{2}$" }, 
+		{ name: "T", type: "Transform", pos: [ 4.0, 3.0120032375808106, 0.12321971730404363 ], color: [ 1.0, 0.0, 0.0 ], args: [ "Tr1", "G" ], labeled: true }, 
+		{ name: "U", type: "Transform", pos: [ 4.0, 3.155084971423055, 0.1144918446644713 ], color: [ 1.0, 0.0, 0.0 ], args: [ "Tr1", "H" ], labeled: true }, 
+		{ name: "V", type: "Transform", pos: [ 4.0, 2.926257936039025, 0.12327138542933927 ], color: [ 1.0, 0.0, 0.0 ], args: [ "Tr1", "K" ], labeled: true } ], 
+	ports: [ 
+		{ id: "CSCanvas", width: 3200, height: 1548, transform: [ { visibleRect: [ -24.44, 32.4, 103.56, -29.52 ] } ], background: "rgb(168,176,192)" } ] });
```
gagern: Those dimensions are pretty hard to use for mere mortals who don't run a HiDPI-screen at native resolution…

Seriously, if you have that Cinderella file lying around, it would be nice to see it re-exported at a lower zoom and window size. Otherwise we might have to adjust the viewport manually to something more reasonable. `111_Moebius_It` is a bit on the large side as well.

strobelm:
Your wish is my command ;-). I updated the examples to a more SD friendly resolution, the iterated one shows nicely the drawing bug  #139 which i will tackle as soon this gets merged.

gagern:
Yes, I've seen #139 in both the updated examples. But that won't keep the merge here. Thanks for all the work!

gagern:
Looking at #24, `TrMoebiusS` seems to be still missing, right?

PR #142, opened 2015-12-09
Improve links between ref docs, make md2html usable as a module

gagern:
These things are neccessary for the automatic inclusion of the ref manual into the website. This is an extension of #124, so it might make sense to review that first.


ref/Geometric_Operators.md
```diff
@@ -238,21 +238,21 @@ Returns a matrix that represents a translation that maps `‹point1›` to `‹p
...
-#### Affine transformation: `map(‹point1›,‹point2›,‹point3›,‹point4›, ‹point5›,‹point6›)`
+#### Affine transformation: `map(‹point1›,‹point3›,‹point5›,‹point2›, ‹point4›,‹point6›)`
 
 **Description:**
 Returns a matrix that represents an affine transformation that maps `‹point1›` to `‹point2›`, `‹point3›` to `‹point4›`, and `‹point5›` to `‹point6›`.
 
 ------
 
-#### Projective transformation: `map(‹point1›,‹point2›,‹point3›,‹point4›, ‹point5›,‹point6›,‹point7›,‹point8›)`
+#### Projective transformation: `map(‹point1›,‹point3›,‹point5›,‹point7›, ‹point2›,‹point4›,‹point6›,‹point8›)`
```
strobelm: Although  the mapping is correct it's somehow hard to read, what about ```map(‹pointA1›,‹pointA2›,‹pointA3›,‹pointB1›, ‹pointB2›,‹pointB3›)```?

gagern: Well, I copied this from the Cinderella documentation. The correct approach would probably be adjusting this in the Cinderella docs. Not sure why that wasn't done.

PR #143, opened 2015-12-10
Fix HTML validation problems reported for example files

gagern:
While adding validation to the website, I noticed these issues.

I performed the validation using [node-validator-nu 2.1.15](https://github.com/hiroaki-yamamoto/node-validator-nu/releases/tag/2.1.15) which uses [v.Nu](https://validator.github.io/validator/) [15.4.12](https://github.com/validator/validator/releases/tag/15.4.12).

### DIFF
--- a/examples/19_EuclidsAlgX2.html
+++ b/examples/19_EuclidsAlgX2.html
@@ -177,7 +177,7 @@ t=t+"$";
 </div>
  <div style="position:absolute; top:630px;left:10px">
 
-        <input id="ch1" onclick="check()" checked=true type="checkbox" >einrasten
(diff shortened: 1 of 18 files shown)


strobelm:
Hi! Thanks for fixing the bugs. I checked the examples they all work as expected, despite 74_Csad* where the heading is now rendered below the canvas.

gagern:
Ah, you meant “except”, not “despite”. Fixed the affected examples now. The reason was the unneccessary absolute positioning in there.

Which makes me think that we should do a more thorough run over all examples to eliminate absolute positioning in all of them, unless there is a really good reason for it. Many examples align buttons absolutely, and some others still use MathJax. Different pull request though, at some later time.

strobelm:
Great!

PR #144, opened 2015-12-10
Create geometric elements

gagern:
This adds CindyScript functions `create` and `createpoint`, along with all the machinery to properly integrate this with the rest of the geometry module, complex tracing in particular. As a benefit, we're forbidding invalid arguments for geometric elements not just for those dynamically created, but for the initial configuration as well.

I've been sitting on this branch for far too long, I think. The code isn't perfect yet, but the amount of change is large enough already to make reviewing this quite a task, so I shouldn't add to that.

One known deficiency is that it doesn't cope well with elements being added during animation. As discussed on the mailing list, the best approach there would be to remove them once animation stops, since they don't have a valid backup position to revert to. But that would mean the ability to remove elements, which in turn should remove depending elements as well, so there is quite a significant amount of code to be rewritten here to deal with what should be an *extremely* rare situation. Not worth blocking the merge for, I think, but once it got a positive review we should open a ticket to keep track of this known deficiency.

Note that the addition of signatures to all geometric elements means that I'll have to update this branch to include signatures for any operations which get merged into master between now and the time this request gets merged. I'm not sure whether I'll rebase on a regular basis, or just once after I get a positive review.


strobelm:
Hi! I read the code yesterday and had no major concerns. The examples are working as expected despite 65c_Conic_2P_3l.html and 67_Conic_1P_4l.html due to reversed arguments. If we fix this i don't see why this shouldn't be merged. It's great to have createpoint and also signature will safe us some headaches. Thanks @gagern !

strobelm:
The examples (and functions) are fixed and the branch is merged to master, thanks again!

Issue #49, opened 2015-10-29
Inconsistent default text properties


gagern:
Geometrix labels and `drawtext` use different font families and different text sizes by default. This is somewhat counter-intuitive, and also in contrast to what Cinderella does.

Example, exported from Cinderella:

```js
createCindy({ 
  scripts: {draw: 'drawtext(A.xy + [1,-1], "B");'},
  defaultAppearance: {}, 
  geometry: [ 
    { name: "A", type: "Free", pos: [ 4.0, -1.804878048780488, -1.2195121951219514 ], color: [ 1.0, 0.0, 0.0 ], labeled: true } ], 
  ports: [{ id: "CSCanvas", width: 680, height: 336, transform: [ { visibleRect: [ -9.04, 9.32, 18.16, -4.12 ] } ], background: "rgb(168,176,192)" } ] });
```

CindyJS renders this as ![CindyJS screenshot](https://cloud.githubusercontent.com/assets/190759/10816839/2930f808-7e36-11e5-8e36-339c46316288.png) while Cinderella does ![Cinderella screenshot](https://cloud.githubusercontent.com/assets/190759/10816859/4fc3a8da-7e36-11e5-9ef0-13cd7fec5c18.png).

kortenkamp:
Isn't the real problem that the Cinderella export does not export the properties completely?

gagern:
@kortenkamp: I don't think that's the whole story. Different text-creating things using different defaults feels strange for those who write CindyJS code by hand, so there is definitely something to fix purely on the CindyJS side, in my opinion.

Cinderella might and probably should export text sizes for its labels explicitely, but it should not modify the text of the script to include an explicit `size ->` modifier which was not entered by the user. So `drawtext` should fit in with its surroundings so that the whole result looks about the same in Cinderella and in CindyJS. But since `drawtext` uses `csport.drawingstate.textsize` it should honour `defaultAppearance`, so setting that on export should fix things on export from Cinderella, even if CindyJS might choose to use a different font size by default.

[...]

PR #145, opened 2015-12-11
Change default text size

gagern:
This changes the default text for `drawtext` from 20 to 12, since that's what Cinderella uses. Adding `drawtext(20)` to the move or draw script (but not the init script!) can restore the old behavior, as was done with the examples.

There is also a new setting `defaultAppearance.textsize` which controls the text size for geometric elements, like labels for points. These could only be set per element so far, with the default being fixed at 12.

@richter-gebert, please have a look at this, either merge it yourself or delegate to e.g. @strobelm if the general approach is OK. And be prepared to change all existing content outside this repository which makes use of `drawtext` with no accompanying `textsize` call or `size->` modifier.

This pull request addresses #49, but doesn't unify the font family yet.


gagern:
Before I forget it: @richter-gebert decided to not change the default to `12`. So we should uniformly make the default `20`, while explicitely settling on a default of `12` in cooperation with Cinderella. There are several ways how we could achieve this:

1. Use the `cinderella` property stored in the arguments to `createCindy`. This makes the distinction somewhat obscure, which might be confusing for people who copy&paste code from one widget to another.

2. Have Cinderella export its default font size as `defaultAppearance.textsize`. Use that setting as the default for both labels of `geometry` elements and scripted calls to `drawtext`. This is somewhat unprecedented, since so far `defaultAppearance` only affects `geometry`, not script calls. But it has the nice benefit that users can consistently change their font size in a single place.

3. Have Cinderella export `defaultAppearance.textsize` as above, but also have it issue a call to the `textsize(…)` function at the beginning of any `move` or `draw` script it might contain. This will cause trouble if some widget designer decides to overwrite the unary `textsize` function. In our most recent discussion of this issue, this was what @richter-gebert decided to use, but I hadn't thought of the overwrite issue at that time, so I'm no longer certain this is a good idea.

In case compatibility with past versions of CindyJS-Widgets exported from Cinderella is an issue, then 1. is the *only* viable alternative. And even that doesn't work for Cinderella versions which didn't export that setting yet. Luckily, the most recent versions of MC² are converting files on the fly now, so we don't have to worry about those. Currently my personal preference is 2.

kranich:
@gagern As far as I recall, @richter-gebert agreed upon 2. Can this be merged now?

gagern:
@kranich: The code as it stands does not implement 2., so it needs to be re-implemented before getting merged.

gagern:
Closing in favor of #345 which takes approach 2.

PR #149, opened 2015-12-17
implemented moebius transform for segments

strobelm:
The forgotten Möbius transformation for segments with an example exported from Cinderella (which works out of the box).

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -1826,6 +1826,31 @@ geoOps.TrMoebiusL.updatePosition = function(el) {
 };
 
 
+geoOps.TrMoebiusS = {};
+geoOps.TrMoebiusS.kind = "C";
+geoOps.TrMoebiusS.signature = ["Mt", "S"];
+geoOps.TrMoebiusS.updatePosition = function(el) {
+    var tr = csgeo.csnames[(el.args[0])];
+    var s = csgeo.csnames[(el.args[1])];
+
+    var a1 = s.startpos;
+    var a3 = s.endpos;
+    // gen further point on segment -- farpoint isn't really useful here
+    var a2 = geoOps._helper.midpoint(a1, a3);
+
+    var b1 = geoOps._helper.TrMoebiusP(a1, tr);
+    var b2 = geoOps._helper.TrMoebiusP(a2, tr);
+    var b3 = geoOps._helper.TrMoebiusP(a3, tr);
+    el.startPoint = b1;
+    el.viaPoint = b2;
+    el.endPoint = b3;
+
+    el.isArc = true;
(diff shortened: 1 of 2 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -1826,6 +1826,31 @@ geoOps.TrMoebiusL.updatePosition = function(el) {
...
+geoOps.TrMoebiusS.kind = "C";
+geoOps.TrMoebiusS.signature = ["Mt", "S"];
+geoOps.TrMoebiusS.updatePosition = function(el) {
+    var tr = csgeo.csnames[(el.args[0])];
+    var s = csgeo.csnames[(el.args[1])];
+
+    var a1 = s.startpos;
+    var a3 = s.endpos;
+    // gen further point on segment -- farpoint isn't really useful here
+    var a2 = geoOps._helper.midpoint(a1, a3);
```
gagern: This may do the wrong thing if the input is a projectively transformed line segment, in particular one which passes through infinity. Possible solutions:

1. Change the implementation of segments (i.e. `Segment` and `TransformS`) to use `via` (which is (the image of) the original midpoint) instead of `farpoint`. Also change `PointOnSegment` accordingly. Cinderella uses this approach, and @richter-gebert tends to think along these lines. Personally I find 0,1,∞ to be a more intuitive frame than 0,1,½ which is why I chose the approach using `farpoint` instead. But I can see how the alternative would be easier here.

2. Have a look at how [`PointOnSegment.updatePosition`](https://github.com/strobelm/CindyJS/blob/ba3cc2a321b95c4869b0ac21dcf755dda661678b/src/js/libgeo/GeoOps.js#L574-L587) works. If you imagine `param = 0.5` you get the projectively correct version of the midpoint. The comment in that function suggests that we are not doing this right for projectively transformed segments either, so this may need more work there.

3. Change `Segment`, `TransformS` and `PointOnSegment` to avoid both `farpoint` and `via`. Instead, make use of suitably chosen representatives, so that `end` - `start` *is* the farpoint and `end` + `start` *is* the midpoint. This approach means one would have to be careful with `normalizeMax` calls for these vectors, since they could only be normalized in tandem. One could e.g. call `normalizeMax` on a 2×3 matrix containing these vectors. On the other hand this formalism would probably simplify a number of computations since you wouldn't have to adjust the scale of vectors before doing a linear combination of them. Since `.startpos` and `.endpos` are not `.homog`, the chances of anyone accidentially modifying these parameters are not that large, and we can easily grep for occurrences and see they do things right.

Personally I'm in favor of the last option. What do you think?

gagern:
Re-assign back to me once this is ready for review again.

gagern:
Superseded by #453, resembling option 3 above.

PR #150, opened 2015-12-18
implemented arrows in gslp

strobelm:
Until now we can't have arrows in the gslp, this commit changes this. 

Furthermore we need to update the Cinderella export which doesn't hand over arrows at all currently.

### DIFF
--- a/examples/arrow_playground.html
+++ b/examples/arrow_playground.html
@@ -110,6 +110,15 @@ draw(H,K,size->2,alpha->0.5, arrow->true, arrowshape->"full", arrowsides->"==>",
                       {name:"X", type:"Free", pos:[3,5],color:[1,0,0],size:3},
                       {name:"Y", type:"Free", pos:[9,5],color:[1,0,0],size:3},
 
+                      {name:"XX", type:"Free", pos:[8,-5],color:[1,0,0],size:3},
+                      {name:"YY", type:"Free", pos:[12,-5],color:[1,0,0],size:3},
+		      {name:"GSLP", type:"Segment", args:["XX","YY"], arrow:true, arrowsides: "<==>", arrowshape: "full", arrowsize: 2, arrowposition: 0.8, color: [1,0,0],alpha:0.5 },
+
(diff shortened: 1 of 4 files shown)


gagern:
It took me a moment to remember that since ac24e5da7a096face87db7197e8cf5f01d0a6a9e it's OK to pass `undefined` as modifier value internally. Now I'm happy with this, but would like to see interaction with Cinderella export before merging this into master. Will start working on the export there now.

gagern:
After correctly translating and scaling attributes, export seems to work quite well. Posted as Cinderella merge request 50. We still lack some arrowheads, see #152.

The arrow translations in my Cinderella merge request read like this:

```java
        add("arrow.mode", 0).map("arrowsides")
            .mapValue(ArrowType.NO, null)
            .mapValue(ArrowType.START, "==>")
            .mapValue(ArrowType.END, "<==")
            .mapValue(ArrowType.BOTH, "<==>")
            .mapExhaustive();
        add("arrow.fraction", 1).map("arrowposition").scale(0.01)
            .depends(ArrowType.modeAtt, ArrowType.NO);
        add("arrow.size", null).map("arrowsize").scale(0.2)
            .depends(ArrowType.modeAtt, ArrowType.NO);
        add("arrow.type", null).map("arrowshape")
            .depends(ArrowType.modeAtt, ArrowType.NO)
            .mapValue(0, "default")
            .mapValue(1, "hollow")
            .mapValue(2, "full")
            .mapValue(3, "delta")
            .mapExhaustive();
```

The arrow position scale factor 0.01 is to convert from integer percentage to floating-point fraction. The size scale factor 0.2 is less obvious. On the one hand, arrow sizes in CindyJS are relative to a default of 10px. On the other hand, the internal arrow description in Cinderella has an edge length of 2 for the base edge (and √5 for the legs) so we need to scale by (approximately) 2. Perhaps the factor 10 in CindyJS was not the best idea, but on the other hand it's probably not serious enough to break existing content over this. Particularly since Cinderella has no arrow drawing capabilities from script code at all. @kortenkamp, if Cinderella gains arrow drawing support one day, would it be OK to follow the CindyJS size interpretation?

PR #151, opened 2015-12-27
Implements reflection transforms

elkins0:
Reflections involving points and lines are general transformations while circle reflections are not. Reflection involving general conics is undefined. Other than points, all elements map to circles in a circle reflection. In a circle reflection angles are preserved and there is a one-to-one correspondence of points within the circle to those that lie outside of it. Points on the circle of reflection map to themselves. The point at the center of the circle of reflection maps to the ideal point 0:0:0. Reflection in a circle of infinite radius is equivalent to reflecting in a line. This particular tour-de-force is shown in the [ReflectionInThreePointCircle.html](https://github.com/elkins0/CindyJS/commit/f349f716b1277d5a4a3257ceb4d924780c9160d6) example.
This is an item on the #24 list.

### DIFF
--- a/examples/ReflectionInCircle.html
+++ b/examples/ReflectionInCircle.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>ReflectionInCircle.cdy</title>
+    <style type="text/css">
+        * {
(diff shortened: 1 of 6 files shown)


strobelm:
Hi, code looks good, examples are working fine. Merged.

Issue #153, opened 2016-01-07
some make errors hard to understand


montaga:
Hi folks!
I just compiled the CindyJS on a new machine and spent some time notice that java is required for closure.
The error that was displayed was the following:

java ... [a bunch of several lines]
execvp(): No such file or directory
Error: spawn ENOENT
    at errnoException (child_process.js:988:11)
    at Process.ChildProcess._handle.onexit (child_process.js:779:34)
From previous event:
    at Array.reduce (native)
    at Task.run (/home/aaron/m10/CindyJS/make/Task.js:225:22)
From previous event:
    at /home/aaron/m10/CindyJS/make/Task.js:193:18
Makefile:2: recipe for target 'all' failed

(I had problems reading from that that java is missing)
It would be cool if there was an automatic test that detects whether there is a working version of java, npm, node

gagern:
Good point, will see to it.

gagern:
On my system, node does include the name of the command in the error message. Nevertheless, the stack trace is still confusing, and we don't want to force people to use my version of node either. The fix https://github.com/gagern/CindyJS/commit/56b8b45f370d07c287a17a9055329daa5528355d is included in #154 and waiting for review.

gagern:
Closed via #154.

PR #154, opened 2016-01-07
Build system output improvements

gagern:
This makes the build system output more readable:

* Colors all over the place, which also helps identifying the beginning of wrapped lines
* Output lines prefixed by task name
* Clear errors if command not found, fixing #153


strobelm:
Hi! The code works nicely on my fedora system.

I also tested the code in a docker image with ubuntu 14.04 and found the following:

After installing build-essential and git the make fails with
```/bin/sh: 1: node: not found ```
this is fine and obvious.

after installing nodejs the output is the same since ubuntu 14.04 installs calls the node executable nodejs. Perhaps we should check for both: node and nodejs

If npm is not installed we have
```
Running npm install to make sure we have all dependencies
execvp(): No such file or directory

/root/CindyJS/make/index.js:62
        throw err;
              ^
Error: spawn ENOENT
    at errnoException (child_process.js:988:11)
    at Process.ChildProcess._handle.onexit (child_process.js:779:34)
make: *** [all] Error 8
```
It's quite clear whats wrong, so no urgent need to change this. 

I think we are fine with a check for nodejs and if you're in the mood perhaps a better warning for npm.

gagern:
> After installing build-essential and git the make fails with
> `/bin/sh: 1: node: not found`
> this is fine and obvious.

I disagree. I deliberately left the “download and unpack node” part included in the Makefile. If that doesn't work, it's because I forgot to wire up the [`js_make` target](https://github.com/CindyJS/CindyJS/blob/f39b7d04926bb51ec9a8ece4c6fc45c0ca0c24fd/Makefile#L54) appropriately.

I'll try to wire up the download, add the name autodetection, have a look at the version autodetection, and then notify you here on this PR.

gagern:
@strobelm I think this should be ready now. Travis had some problems today, but I guess they were unrelated.

The use of `nodejs` in Debian-based systems is kind of problematic. https://github.com/gagern/CindyJS/commit/fbd6471ffa1c98d2da21b948e8dcc59672121381 is my attempt to deal with that, by adding a `node` wrapper to the `PATH`. Other solutions would include requiring the [nodejs-legacy package](https://packages.debian.org/stretch/nodejs-legacy), [fixing expect to use `$NODE`](https://github.com/mjackson/expect/issues/64), or switching from [expect](https://www.npmjs.com/package/expect) to e.g. [Chai](https://www.npmjs.com/package/chai).

Note that https://github.com/gagern/CindyJS/commit/7fdc737671c772bfa39ce65e88d1407ab6b4dc0c changes the default for manual compilation from closure to plain, as we recently discussed. The default for Cindy3D remains unaffected.

strobelm:
Much better thanks!

PR #155, opened 2016-01-08
implemented indicator for non-degenerate circles

strobelm:
Hi following up on bug #139 i wrote some routines which fix drawing bugs if circles are getting very small after applying Möbius transforms. 

The key idea is to indicate that known non-degenerate circles will be still circles after the Möbius transform and loop in that property until we reach the conic drawing routine.

On the way i refactored functions using random lines (which is used quite often to generate points on the circles).

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -400,7 +400,7 @@ eval_helper.drawconic = function(conicMatrix, modifs, isNotDegenCircle) {
         return;
     Render2D.preDrawCurve();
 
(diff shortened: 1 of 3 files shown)


kranich:
@strobelm Please `git rebase master` to fix the failing Travis CI build.

strobelm:
Hi! I just rethought the whole process and perhaps there is a more elegant way: the center of the circle.
I'll code some lines and let you know if it worked out.

gagern:
Superseded by #179.

PR #157, opened 2016-01-14
Declare inputs of the shader resource build step

gagern:
Without this, a change to one of these inputs will not trigger a recreation of the corresponding output.


strobelm:
Small commit, large impact. Rebuilding seems to work.

PR #160, opened 2016-01-14
Add getImage function to the Plugin API

gagern:
Using this function, it becomes possible for plugins to operate on the same collection of images used in CindyJS itself.


strobelm:
Looks good!

PR #164, opened 2016-01-15
Gl

montaga:
First Version of semi-working GL :)


kranich:
@montaga Please `git rebase master` to fix the failing Travis CI build.

gagern:
Rebase not neccessary, but we need this beautified. See updated Travis output.

montaga:
Thank you for your hints. Now everything should be fine for merging.

strobelm:
Perhaps we could squash some commits and rebase into master (we could do this together in the office @montaga )

gagern:
Don't squash too much: we should be able to see the core ideas in the development process, and to check out any previous versions which might have worked differently and thus might serve for comparison. Squashing minor fixes and getting rid of non-viable code seems all right and is probably a good idea, but I'd leave substantial code edits intact. Note that rebase has trouble handling merges.

strobelm:
@montaga and i came to terms that the rebase quite tedious, if there are no objections i will merge the branch this evening.

PR #165, opened 2016-01-15
Reimplement tokenize

gagern:
This fixes various problems, see individual commits. Among them are problems newly reported by recent jshint.


gagern:
Thanks a lot!

PR #171, opened 2016-01-18
Fix typo

kranich:
`node make alltests` complains about `tokens` being used out of scope. I believe this should be `token` instead (typo).

### DIFF
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -3127,7 +3127,7 @@ evaluator.tokenize$2 = function(args, modifs) { //TODO der ist gerade sehr uneff
 
         var token = v1.value[0];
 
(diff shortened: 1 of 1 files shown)


strobelm:
I think i already discussed message by jshint some time ago with @gagern but i can't remember why. @gagern any ideas?

montaga:
Probably #165 should properly fix that?

kranich:
@montaga You are right, #165 should fix that.

gagern:
@kranich If you want to, you might do the review for #165. It feels like a pretty annoying showstopper right now that we can't rely on Travis results due to this, so I'd like to see this resolved today, one way or another.

kranich:
@gagern OK, I will review #165.

PR #172, opened 2016-01-19
Angular bisector

gagern:
This makes `AngularBisector` ready to be exported from Cinderella. It requires a version of Cinderella which is recent enough to export `SelectL` not `SelectP`. The implementation of `SelectL` required a rework similar to what I did to `SelectP` in 3eefed8632fb0d7a4584c573137689b77c7d8ec0. For reasons unknown (to me), Cinderella exports the point of intersection as a third argument. Since our internal operation only expects two lines, I'm using the macro step to also ensure that the third argument gets dropped. I will probably file a Cinderella merge request to not export that argument at all.

### DIFF
--- a/examples/AngularBisectorCdy.html
+++ b/examples/AngularBisectorCdy.html
@@ -1,0 +1,47 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>AngularBisectorCdy.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+    <script type="text/javascript">
+createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0 }, 
+	angleUnit: "°", 
+	geometry: [ 
+		{ name: "A", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 1.0, 0.0, 0.0 ] }, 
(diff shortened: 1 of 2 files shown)


kortenkamp:
Here is the reason for the third argument: for two lines with the same coordinates the second bisector is undefined, unless you determine a point for it (as the two lines have no single point of intersection). So I am quite sure that your implementation will fail in such cases (that are quite common).

kranich:
I think we should consider renaming the operator to `AngleBisector'. This is shorter and [*much* more common in English](https://books.google.com/ngrams/graph?content=angular+bisector,angle+bisector&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1;,angular%20bisector;,c0;.t1;,angle%20bisector;,c0). I could not even find "angular bisector" in any English dictionary.

gagern:
@kortenkamp thanks for the clarification. Perhaps we should implement `AngleBisector` or `AngularBisector` as a ternary operation like this, and turn the old name into a macro which constructs its point of intersection internally.

kranich:
@gagern Please `git rebase master` to fix the failing Travis CI build.

strobelm:
@gagern are you planing to rework the code or should i have a look at it?

gagern:
I had intended to look at this sooner or later. But I've got enough other things to keep me busy, and the angle bisector was originally your thing so you will probably be faster to adapt it. Plus with the third argument added to the game, my original idea of trying a different formulation here won't work any more, so my motivation to do my own thing here has been reduced (at least until I understand some corner cases a bit better).

So please go ahead and have a look, @strobelm. If you take this up, perhaps do your own pull request branched off https://github.com/gagern/CindyJS/commit/319cda969b38f18562c815faa2907061479a5525. Let me know whether you agree with the first two commits I made here, then I'll gladly review anything you have to add to that.

strobelm:
I'll look into it. Generally i wanted to implement the angle n-sector in the near future. The "trick" with the point won't work for this since we have a larger variety of possibilities for n >=3. Perhaps we get this done right before we tackle the larger problem.

elkins0:
@strobelm I have continued this as *Continuation of Angular bisector* #183

gagern:
This is no longer an active proposal. Better things are being worked on.

Issue #166, opened 2016-01-17
Problem with triangles and lines


bellezaodiada:
The following code to demonstrate Desargue's theorem does not work in CindyJS but perfect in the java version. The triangles no longer fill with colour and some lines are cut short.
The original Cinderella file is attached and so is the "exported to CindyJS" file. All using version 2.8 build 1781

[desargues.zip](https://github.com/CindyJS/CindyJS/files/93420/desargues.zip)

strobelm:
Hi! The problem is caused by Poly which is not implemented as geoOps operator. I'll have a look.

bellezaodiada:
Many thanks.

Graham

bellezaodiada:
Michael

Was there any progress on this problem?
My examples of Desargues Theorem still don't work on Cindy JS. Is there 
any likelihood of this being fixed in the reasonably near future.

I attach the Cinderella file as a reminder of the problem.

Many thanks

Graham

bellezaodiada:
Here is the Cinderella file for the Desargues demonstration. Works perfectly in Cinderella with Java but doesn't work in CindyJS
[desargues3d.zip](https://github.com/CindyJS/CindyJS/files/190878/desargues3d.zip)

bellezaodiada:
Closed in error :-)

strobelm:
Hi Graham,

i haven't forgot about the issue and already started writing some code:

https://github.com/CindyJS/CindyJS/pull/173/files

It's not finished yet, i hope i can implement it within the next one or
two weeks.

best regards

Michael

On 03/27/2016 01:25 PM, bellezaodiada wrote:
> Michael
>
> Was there any progress on this problem?
> My examples of Desargues Theorem still don't work on Cindy JS. Is there 
> any likelihood of this being fixed in the reasonably near future.
>
> I attach the Cinderella file as a reminder of the problem.
>
> Many thanks
>
> Graham
>
>
> El 19/01/2016 a las 10:41, Michael Strobel escribió:
>> Hi! The problem is caused by Poly which is not implemented as geoOps 
>> operator. I'll have a look.
>>
>> —
>> Reply to this email directly or view it on GitHub 
>> <https://github.com/CindyJS/CindyJS/issues/166#issuecomment-172793568>.
>>
>> No se encontraron virus en este mensaje.
>> Comprobado por AVG - www.avg.com <http://www.avg.com>
[...]

PR #173, opened 2016-01-19
initial version of polygons in gslp

strobelm:
Hi,

this is not really intended to be a pull request, rather code to open up discussion.

I was looking into issue #166 which deals also with polygons in the gslp. Currently the signatures in geoOps expect the number of arguments to be of fixed size. 

This raises some questions:
1) How to handle arbitrary sized arguments (like in polygons)
2) do we want to have mixed sets of Lines, Points, Conics etc... as arguments? Is there a use case?
3) Do we need a new GeoOps type for polygons?

Additionally i wasn't sure how to handle the arguments in a proper way fitting for eval_helper.drawpolygon. I would really appreciate comment on that.


gagern:
Only had a cursory look at the code so far.

> 1) How to handle arbitrary sized arguments (like in polygons)

Good question, But your `signature = "Ps"` looks pretty good to me. Seeing how we'd probably write `args: ["A", "B", "C"]` and not `args: [["A", "B", "C"]]`, it might make sense to not write this as  `signature = ["Ps"]`.

Initially I thought we might want to distinguish between sets of points and sequences of points. But the sets of points (as returned from the multi-valued functions) are kept in order using tracing, so they are in fact sequences as well. So the key distinction is that a `Ps` as a single argument is identified by a single name, whereas a `Ps` as the whole signature corresponds to a sequence of names.

> 2) do we want to have mixed sets of Lines, Points, Conics etc... as arguments? Is there a use case?

Can't think of one. If one turns up, I'm certin we'll find a suitable solution. The choices made here don't feel like they'd be blocking any roads, particularly seeing how all that signature codes is just internal anyway. As long as we keep the JSON structure intact, it shouldn't matter what we call our arguments.

> 3) Do we need a new GeoOps type for polygons?

Definitely. `Poly` sounds good. The names exported from Cinderella, e.g. when transforming a polygon, will explicitely use `Polygon` to be perfectly clear, but I see no urgent need to follow that here.

> Additionally i wasn't sure how to handle the arguments in a proper way fitting for `eval_helper.drawpolygon`. I would really appreciate comment on that.

Looking at [that helper](https://github.com/CindyJS/CindyJS/blob/6a3ead8ecc56aa47b6d00bd0b45dbac8c85f225e/src/js/libcs/OpDrawing.js#L850-L855), it seems it understands both a list and a shape as first and only element of its `args` array. So I'd say the `args` argument should be something like this:

```js
var args = [el.args.map(function(pt) {
    return csgeo.csnames[pt].homog;
})];
```

kranich:
@strobelm Please `git rebase master` to fix the failing Travis CI build.

gagern:
No need to rebase, I triggered Travis to rerun this job.

gagern:
What's the status here? It would be nice to have polygon support soon. Do you want to continue this, @strobelm, or should someone else pick this up when they find the time?

strobelm:
Hi! I looked at the branch some time ago and there were some nasty merge
conflicts with master. I think taking the ideas and rewriting the code
would be the best way to go. It's on my task list after randomized
proving but if you got some spare time Martin i won't mind at all ;-).

Am 31.05.2016 um 10:46 schrieb Martin von Gagern:
> What's the status here? It would be nice to have polygon support soon.
> Do you want to continue this, @strobelm <https://github.com/strobelm>,
> or should someone else pick this up when they find the time?
> 
> —
> <https://github.com/CindyJS/CindyJS/pull/173#issuecomment-222628222>, or
> mute the thread
> <https://github.com/notifications/unsubscribe/ANhz4hZTP1V7woLYxnRhUOPbxJNrUsJsks5qG_VmgaJpZM4HH4hj>.
> 

-- 
Michael Strobel, M.Sc.

strobelm:
I think polygons are done. Export from Cinderella is currently broken, @gagern already has a PR for that (thanks a lot!).

gagern:
@strobelm Please make Travis happy, by making `CSCanvas` a `<div>` instead of a `<canvas>`.

strobelm:
Done.

By the way: something seems to be odd with error handling, i got these message instead of the actual output:
```
cindyjs/make/commands.js:55
                    cmdline + " exited with code " + code));
                    ^
ReferenceError: cmdline is not defined
```

kranich:
> Please make Travis happy, by making CSCanvas a <div> instead of a <canvas>.

Don't you think that it is slightly weird that `CSCanvas` is a `<div>`? Maybe we should change that before 1.0.0?

gagern:
> By the way: something seems to be odd with error handling

Yes, I noticed. https://github.com/CindyJS/CindyJS/pull/386/commits/d9d957e84c9fa2dfd4c2a076e2fc9167a75da1ed for and https://github.com/CindyJS/CindyJS/pull/382/commits/d8355918dc675cbf8465db0c4b6d437d03abab8d fix this, so if either of these gets merged the issue should be resolved. Perhaps I should cherry-pick this into a pull request in its own right, though.

> Don't you think that it is slightly weird that `CSCanvas` is a `<div>`? Maybe we should change that before 1.0.0?

Change what? It doesn't *have* to be a `<div>`, but if it is a `<canvas>`, CindyJS will rebuild it into a `<div>` nonetheless, moving attributes as it goes. This could be very confusing for CSS stylesheets, since e.g. a rule like `canvas#CSCanvas` wouldn't match even though the canvas appears in the original HTML source. That's the reason I'm suggesting `<div>` instead.

We could change the name in all our examples, but a name is just that: a name. Do you have a good alternative name to suggest?

kranich:
> We could change the name in all our examples, but a name is just that: a name. Do you have a good alternative name to suggest?

How about `CSWidget`?

src/js/libgeo/GeoBasics.js
```diff
@@ -118,6 +119,12 @@ function segmentDefault(el) {
     el.clip = General.string("end");
 }
 
+
+function polygonDefault(el) {
+    el.filled = (el.filled !== "undefined" ? General.bool(el.filled) : General.bool(true));
```
gagern: You mean `el.filled !== undefined` without the quotes. And you should use `General.bool(!!el.filled)` or some such, since we don't know whether `el.filled` really is a boolean value.

src/js/libgeo/GeoBasics.js
```diff
@@ -154,13 +161,27 @@ function addElementNoProof(el) {
 
     // Detect unsupported operations or missing or incorrect arguments
     var op = geoOps[el.type];
+    var isSet = false;
+    var getKind = function(el) {
+        return csgeo.csnames[el].kind;
```
gagern: I think it would be better to name that argument `name`, since `el` is typically the object for a geometric element, not its name. Since you are only using this in a single location, I personally would prefer an anonymous function at the point of use, so one doesn't have to scroll to relate these two.

src/js/libgeo/GeoBasics.js
```diff
@@ -154,13 +161,27 @@ function addElementNoProof(el) {
...
+
     if (!op) {
         console.error(el);
         console.error("Operation " + el.type + " not implemented yet");
         return null;
     }
     if (op.signature !== "**") {
-        if (op.signature.length !== (el.args ? el.args.length : 0)) {
+        // check for sets
+        if (!Array.isArray(op.signature) && op.signature.charAt(1) === "s") {
```
gagern: I *think* I had originally chosen the `**` syntax for arbitrary parameters because I wanted `P*` for points and so on. You use `Ps` instead. I think there is some risk of confusion, since `["Ps"]` would bean a single set of points, as returned by a multi-valued operation, while `"Ps"` without brackets would be an arbitrary number of points. Easy to mix up, I think. Furthermore, `P*` allows for a fairly canonical `P+` one day in cases where we at least one element. It also allows for `["P", "P", "P*"]` or similar if we ever have an operaion that takes two distinguished points and a set of other points. What do you think?

src/js/libgeo/GeoBasics.js
```diff
@@ -174,9 +195,13 @@ function addElementNoProof(el) {
                 return null;
             }
             if (op.signature !== "**") {
+                var opSigs;
+                if (isSet) {
+                    opSigs = el.args.map(getKind);
```
gagern: This feels kind of weird. If I understand you correctly, you generate a list of actual kinds to compare against the actual kinds. A check which will never fail. So I think it should be skipped in the first place. The code here is also inside the `for i` loop, so you do this again for each argument, which comes at a slight performance cost and at some potential for confusion as well. Can't you just add `|| !isSet` to the `**` check above? Or, in case you decide on `*` instead of `s` as the suffix, make that `op.signature.charAt(1) !== "*"` or `/[^*]$/.test(op.signature)` or something like that?

Another alternative would be moving the semantics from the `el.args.forEach` above to this `for (i…` loop here. In the code above you'd set `isSet` but not do any type checking. Then here you'd set `opSig` to `op.signatures[i]` if you are in the `!isSet` case, but to `op.signature.charAt(1)` in the `isSet` case. Then you compare `opSig` against `argKind`. That would concentrate all the looping over arguments to a single location, which I think would make this easier to maintain.

src/js/libgeo/GeoOps.js
```diff
@@ -2519,6 +2520,12 @@ geoOps._helper.initializeLine = function(el) {
 };
 
 
+geoOps.Poly = {};
+geoOps.Poly.kind = "Poly";
+geoOps.Poly.signature = "Ps";
+geoOps.Poly.updatePosition = noop;
```
gagern: The `updatePosition` code should collect all the `homog` vectors from the arguments into a list. That's because one day we'll want to transform polygons (e.g. apply projective transformations to them), and then we can't derive those coordinates from the arguments without case distinctions.

Do we need points at infinity or via points or some such, in order to identify the line segments correctly after a projective transformation? If unsure we can postpose adding them, but I fear we are going to need them.

src/js/libgeo/GeoRender.js
```diff
@@ -241,6 +241,30 @@ function drawgeotext(el) {
...
 }
 
+function drawgeopolygon(el) {
+    if (!el.isshowing || el.visible === false)
+        return;
+
+    var modifs = {};
+    modifs.color = el.color;
+    modifs.alpha = el.alpha;
+    modifs.size = el.size;
```
gagern: Shouldn't these properties be set by `polygonDefault`? Otherwise they won't be wrapped as CindyJS data types.

gagern:
I finally found the time to give this a closer look. Sorry for the delay, @strobelm! Will you be able to address the issues? If not, we might want to merge this as is, since it's still way better than not having polygons at all. But it would be even better if we could merge this in a day or two with my concerns addressed.

strobelm:
Hi, thanks for the comments! I'll fix the problems asap (probably today or tomorrow).

src/js/libgeo/GeoRender.js
strobelm: iirc this is done in polygonDefault und just rewrapped for the drawing function.

strobelm:
Here is the update, i think i resolved the issues. No work has been done in order to save the polygon point homogs for transformations, should be noted on the todo list.

src/js/libgeo/GeoRender.js
gagern: Ah, I missed the fact that `polygonDefault` defers to `lineDefault` which does these things.

gagern:
I did export a bunch of polygons with different settings from Cinderella and noticed that the appearance looks quite different. In particular, in Cinderella polygons are by default filled *and* stroked. This is exported using `fillcolor` and `fillalpha` in addition to `color` and `alpha`. If `fillalpha` is absent, it defaults to zero, thus leading to an unfilled polygon.

The `fill` property used by the current code doesn't seem to correspond to anything Cinderella exports. Is this being used elsewhere? I found no further references in `GeoBasics` either.

gagern:
I'm working on a pull request, please stand by…

PR #176, opened 2016-01-20
Fix CSNumber.pow for base 0

### LINKED ISSUE #175: CSNumber.pow does not work for base 0
CindyJS yields `NaN - i*NaN` instead of `0` for `0^(1/4)`.

Cinderella yields `1` for `0^0`. Although I am not sure `0^0` is mathematically well-defined, CindyJS should yield `1` for compatibility with Cinderella.

I will prepare a pull request to fix this.

kranich:
This should fix #175.

### DIFF
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -114,6 +114,10 @@ For noninteger values of `b` only one principal value of `a^b` will be returned.
     < 0.2
     > 2^(1/2)
     < 1.4142
+    > 0^(1/4)
+    < 0
+    > 0^0
+    < 1
 
 ------
 
--- a/src/js/libcs/CSNumber.js
+++ b/src/js/libcs/CSNumber.js
@@ -465,31 +465,13 @@ CSNumber.log = function(a) {
 
 
 CSNumber.pow = function(a, b) {
-
     if (CSNumber._helper.isZero(b))
         return CSNumber.one;
     if (CSNumber._helper.isZero(a))
         return CSNumber.zero;
-
-    if (b.value.real === Math.round(b.value.real) && b.value.imag === 0) { //TODO später mal effizienter machen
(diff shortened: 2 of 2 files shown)


gagern:
I wonder whether it would be better to do two special cases:

1. If exponent is 0, return 1
2. If base is 0, return 0

That way, we implicitely get 0<sup>0</sup> = 1 and simplify the computation for that case as well.

While you are at it, how about improving some other special cases as well? If the exponent is real, you can delegate to `pow2`, which should make things like `i^(2^30)` feasible.

gagern:
I'd also like to wait on #165 before merging this, to get a proper Travis run. If you want to do the review there, be my guest.

kranich:
@gagern I have implemented your suggested improvements.

kranich:
@gagern I find the name `CSNumber.pow2` a bit misleading. I would expect a 'to-the-power-of-two', i.e. `^2`, function under that name. Should we rename it to `CSNumber.powRealExponent` or something similar?

gagern:
Good idea. It took me several close looks to see what this actually was about. Better to be a bit more verbose.

`powRealExponent` is a bit funny in that it spells out the less important term “exponent” while abbreviating the core “power”. `powerRealExponent` would be the very verbose form, but it's a bit long for my taste. `powRealExp` might increase the risk of people assuming an `exp` function in there. `realPow` on the other hand might imply a real base as well. So all things considered, I think your suggestion makes the most sense even if it is a bit funny with respect to its emphasis.

kranich:
@gagern I think this is ready to merge now.

gagern:
Added one more testcase in e0f1bac2616ae2e729bb079e92f9ed66be9972db and merged in f39b7d04926bb51ec9a8ece4c6fc45c0ca0c24fd. Thanks for the fix!

PR #177, opened 2016-01-20
Warn about division by zero only in user-defined script

### LINKED ISSUE #168: Segment to far point causes division by zero
Consider the segment between two finite points A and B. If we set `B.homog` to coordinates of a far point, drawing the segment leads to a division by zero.

kranich:
This should fix #168.

It also changes the error message from "DIVISION BY ZERO" to "WARNING: Division by zero!" as emitted by Cinderella.


gagern:
I think the message should be directed at the `csconsole` now.

kranich:
@gagern You are right. I fixed this and also adjusted the corresponding test cases.

PR #179, opened 2016-01-20
Circle of small radius should not be drawn as lines

### LINKED ISSUE #139: Zero radius circles sometimes drawn as a line

strobelm:
Hi, here is the promised code. It checks if the center of our circle is a finite point, if not we should not draw the circle as line even the radius gets very small (as in 111_Moebius_It.html). The approach is much more lightweight and reliable than the previous pull request. Are there any drawbacks i have not thought of?

@gagern / @kranich: @richter-gebert asked me to merge a fix for the circle problem quickly since he needs this for an applets (or widget or whats the fancy term right now).

This should fix #139 and supersede #155.

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -407,7 +407,7 @@ eval_helper.drawconic = function(conicMatrix, modifs) {
     // check for complex values
     for (var i = 0; i < 2; i++)
         for (var j = 0; j < 2; j++) {
-            if (Math.abs(mat.value[i].value[j].value.imag) > eps) return;
+            if (Math.abs(mat.value[i].value[j].value.imag) > CSNumber.eps) return;
         }
 
     // transform matrix to canvas coordiantes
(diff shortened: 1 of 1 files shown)


gagern:
A definite improvement. We might want to try center of circle as opposed to center of conic one day, since the former is still defined when a circle degenerates to a finite line times the line at infinity. But even so it's a huge step forward, and we want this included.

strobelm:
@gagern wow thanks for looking at this virtually instant! 
@richter-gebert the bugs should be fixed now

Issue #105, opened 2015-11-27
Translate code to English


gagern:
Now that CindyJS is open source and might attract international contributors, we should try to keep the sources consistently in English. Things that come to my mind where this is not the case yet are the occasional German comment and the common use of `erg` to denote “Ergebnis” i.e. “result”. The latter could be consistently renamed to `res` in most situations, the former will need some systematic searching.

PR #181, opened 2016-01-21
remove redundant line in geoOps.Conicby5

strobelm:
The routine geoOps._helper.ConicBy5 possesses a redundant line which is never used in the routine itself. I think it's a leftover from a re-factoring.


kranich:
We also have some

```JavaScript
var erg = List.sub(m1, m2);
return erg;
```

and similar stuff in there. `erg` is short for German 'Ergebnis'; `result` or `res` would be better in an international project. But why not simply 

```JavaScript
return List.sub(m1, m2)
```

Should we fix this?

gagern:
@kranich see #105 for the `erg` stuff.

PR #182, opened 2016-01-22
accessing external textures in CindyGL

montaga:
These commits enable to use imagergb on included images and loaded images as rendering targets.


montaga:
This must not merged yet, since it will break all colorplots on Jürgens MacBook.
If this does not work on any other System, please let me know, in order to find out the bug.

PR #183, opened 2016-01-24
 Continuation of Angular bisector #172 

elkins0:
@gagern wrote:
This makes `AngularBisector` ready to be exported from Cinderella. It requires a version of Cinderella which is recent enough to export `SelectL` not `SelectP`. The implementation of `SelectL` required a rework similar to what I did to `SelectP` in [3eefed8](https://github.com/CindyJS/CindyJS/commit/3eefed8632fb0d7a4584c573137689b77c7d8ec0). For reasons unknown (to me), Cinderella exports the point of intersection as a third argument. Since our internal operation only expects two lines, I'm using the macro step to also ensure that the third argument gets dropped. I will probably file a Cinderella merge request to not export that argument at all.

What I've done is simplify `angleBisector.updatePosition` to not create or make use of the intersection of the two lines. I've also modified @gagern's `AngularBisectorCdy.html` to  to run lines a and b through separated points, create the intersection point of the two lines and  show the coordinates of the intersection point using `CindyScript`. The axes, grid and point snapping are also turned on.

Since Cinderella currently exports the `AngularBisector` name instead of the name @kranich and I prefer  `AngleBisector` I have left that name as is. I would welcome any example of @kortenkamp's that shows where this implementation of `angleBisector.updatePosition` fails. When the lines are identical, this implementation produces one line identical to the other two lines and one ideal line [0,0,0]. When the lines are parallel, this implementation produces a line mid-way and parallel to the other two lines and one line at infinity.

### DIFF
--- a/examples/AngularBisectorCdy.html
+++ b/examples/AngularBisectorCdy.html
@@ -21,8 +21,6 @@
     </style>
     <script type="text/javascript" src="../build/js/Cindy.js"></script>
 <script id="csdraw" type="text/x-cindyscript">
-scale=1;//4 in Cinderella
-drawtext((-9,8),"C"+C.homog/scale,color->(0,0.7,0),size->20);
 draw((-9.06,0),(14.54,0),arrow->true,size->2,color->(128/255,128/255,128/255));//x-axis
 draw((0,-6.34),(0,9.34),arrow->true,size->2,color->(128/255,128/255,128/255));//y-axis
 </script>
@@ -35,10 +33,8 @@ createCindy({
 	geometry: [ 
 		{ name: "A", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 1.0, 0.0, 0.0 ] }, 
 		{ name: "a", type: "Through", pos: [ -0.8739495798319329, -4.0, 0.1250420168067227 ], color: [ 1.0, 0.0, 0.0 ], args: [ "A" ] }, 
-		{ name: "B", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 0.0, 0.0, 1.0 ] }, 
-		{ name: "b", type: "Through", pos: [ 3.2682926829268295, -4.0, 0.2907317073170732 ], color: [ 0.0, 0.0, 1.0 ], args: [ "B" ] }, 
-		{ name: "C", type: "Meet", pos: [ -0.16, 0.16, 4.0 ], color: [ 0.0, 1.0, 0.0 ], args: [ "a", "b" ] }, 
-		{ name: "Collection__1", type: "AngularBisector", args: [ "a", "b", "C" ] }, 
+		{ name: "b", type: "Through", pos: [ 3.2682926829268295, -4.0, 0.2907317073170732 ], color: [ 0.0, 0.0, 1.0 ], args: [ "A" ] }, 
+		{ name: "Collection__1", type: "AngularBisector", args: [ "a", "b", "A" ] }, 
 		{ name: "c", type: "SelectL", pos: [ 0.957605970573069, -4.0, 0.19830423882292275 ], color: [ 1.0, 1.0, 0.0 ], args: [ "Collection__1" ] }, 
 		{ name: "d", type: "SelectL", pos: [ 4.0, 0.9576059705730687, 0.12169576117707727 ], color: [ 0.0, 1.0, 1.0 ], args: [ "Collection__1" ] }, 
 		{ name: "Collection__2", type: "AngularBisector", args: [ "c", "d" ] }, 
--- a/examples/AngularBisectorCdy.html
+++ b/examples/AngularBisectorCdy.html
@@ -1,0 +1,55 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>AngularBisectorCdy.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csdraw" type="text/x-cindyscript">
+scale=1;//4 in Cinderella
+drawtext((-9,8),"C"+C.homog/scale,color->(0,0.7,0),size->20);
+draw((-9.06,0),(14.54,0),arrow->true,size->2,color->(128/255,128/255,128/255));//x-axis
+draw((0,-6.34),(0,9.34),arrow->true,size->2,color->(128/255,128/255,128/255));//y-axis
+</script>
+
+    <script type="text/javascript">
+createCindy({
+	scripts: "cs*", grid: 1, snap: true, 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0, dimDependent: 0.7 }, 
+	angleUnit: "°", 
+	geometry: [ 
+		{ name: "A", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 1.0, 0.0, 0.0 ] }, 
+		{ name: "a", type: "Through", pos: [ -0.8739495798319329, -4.0, 0.1250420168067227 ], color: [ 1.0, 0.0, 0.0 ], args: [ "A" ] }, 
+		{ name: "B", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 0.0, 0.0, 1.0 ] }, 
+		{ name: "b", type: "Through", pos: [ 3.2682926829268295, -4.0, 0.2907317073170732 ], color: [ 0.0, 0.0, 1.0 ], args: [ "B" ] }, 
+		{ name: "C", type: "Meet", pos: [ -0.16, 0.16, 4.0 ], color: [ 0.0, 1.0, 0.0 ], args: [ "a", "b" ] }, 
+		{ name: "Collection__1", type: "AngularBisector", args: [ "a", "b", "C" ] }, 
+		{ name: "c", type: "SelectL", pos: [ 0.957605970573069, -4.0, 0.19830423882292275 ], color: [ 1.0, 1.0, 0.0 ], args: [ "Collection__1" ] }, 
+		{ name: "d", type: "SelectL", pos: [ 4.0, 0.9576059705730687, 0.12169576117707727 ], color: [ 0.0, 1.0, 1.0 ], args: [ "Collection__1" ] }, 
+		{ name: "Collection__2", type: "AngularBisector", args: [ "c", "d" ] }, 
+		{ name: "e", type: "SelectL", pos: [ 4.0, -2.454728389053679, 0.2581891355621472 ], color: [ 0.7, 0.0, 0.0 ], args: [ "Collection__2" ], dashtype: "dashed" }, 
+		{ name: "f", type: "SelectL", pos: [ -2.454728389053677, -4.0, 0.061810864437852914 ], color: [ 0.0, 0.0, 0.7 ], args: [ "Collection__2" ], dashtype: "dashed" } ], 
+	ports: [ 
+		{ id: "CSCanvas", width: 590, height: 392, transform: [ { visibleRect: [ -9.06, 9.34, 14.54, -6.34 ] } ], background: "rgb(168,176,192)" } ], 
+	cinderella: { build: 1813, version: [ 2, 9, 1813 ] } });
+    </script>
+</head>
+<body>
+    <canvas id="CSCanvas"></canvas>
+</body>
+</html>
--- a/examples/FreeAngleBisector.html
+++ b/examples/FreeAngleBisector.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>FreeAngleBisector.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csdraw" type="text/x-cindyscript">
+draw((-9.06,0),(14.54,0),arrow->true,size->2,color->(128/255,128/255,128/255));//x-axis
+draw((0,-6.34),(0,9.34),arrow->true,size->2,color->(128/255,128/255,128/255));//y-axis
+</script>
+
+    <script type="text/javascript">
+createCindy({
+	scripts: "cs*", 
+	defaultAppearance: { fontFamily: "sans-serif", lineSize: 1, pointSize: 5.0, dimDependent: 0.7 }, 
+	angleUnit: "°",  grid: 1, snap: true, 
+	geometry: [ 
+    { name: "A", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 1.0, 0.0, 0.0 ] }, 
+		{ name: "a", type: "Through", pos: [ -0.8739495798319329, -4.0, 0.1250420168067227 ], color: [ 1.0, 0.0, 0.0 ], args: [ "A" ] }, 
+		{ name: "B", type: "Free", pos: [ -0.16, 0.16, 4.0 ], color: [ 0.0, 0.0, 1.0 ] }, 
+		{ name: "b", type: "Through", pos: [ 3.2682926829268295, -4.0, 0.2907317073170732 ], color: [ 0.0, 0.0, 1.0 ], args: [ "B" ] }, 
+		{ name: "Collection__1", type: "AngularBisector", args: [ "a", "b" ] }, 
+		{ name: "c", type: "SelectL", pos: [ 0.957605970573069, -4.0, 0.19830423882292275 ], color: [ 1.0, 1.0, 0.0 ], args: [ "Collection__1" ] }, 
+		{ name: "d", type: "SelectL", pos: [ 4.0, 0.9576059705730687, 0.12169576117707727 ], color: [ 0.0, 1.0, 1.0 ], args: [ "Collection__1" ] } ], 
+	ports: [ 
+		{ id: "CSCanvas", width: 590, height: 392, transform: [ { visibleRect: [ -9.06, 9.34, 14.54, -6.34 ] } ], background: "rgb(168,176,192)" } ], 
+	cinderella: { build: 1813, version: [ 2, 9, 1813 ] } });
+    </script>
+</head>
+<body>
+    <canvas id="CSCanvas"></canvas>
+</body>
+</html>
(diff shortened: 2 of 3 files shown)


elkins0:
@strobelm I have added you as a collaborator to hopefully make things easier on you.

elkins0:
An instrumented example using JavaScript and CindyScript to manipulate lines and points in CindyJS is [AngularBisectorCS.html](https://gist.github.com/elkins0/0baf165f01ff9440c9b4). This allow lines a and b to be set parallel at the push of a button. You'll need #184 for that to work.

gagern:
> What I've done is simplify `angleBisector.updatePosition` to not create or make use of the intersection of the two lines.

In https://github.com/CindyJS/CindyJS/pull/172#issuecomment-172908556 @kortenkamp pointed out that we *should* have the point of intersection as a third argument. It is neccessary to have the angular bisectors still defined even if the two defining lines are identical. This is a common scenario, e.g. having two `Through` lines through a common point, so we should be sure not to break that.

gagern:
If the two input lines are finite and identical, then the correct answer is one line which is identical to these two, and one line which is perpendicular *at the defining point*, which in this case can't de deduced from the two lines.

Once we define the three-argument version of an angle bisector, we can use macros to define the two-argument version as one step which defines the point of intersection, and one step which does the bisectors using that. The result obviously won't be defined in the corner case, but we'd be off no worse than with the current implementation, which should be enough for backward compatibility with existing examples.

elkins0:
Ah I see. So if a defining point *is* passed in, then it is only a matter of checking for a degenerate result and replacing it with the perpendicular to the other result through the given *defining point*.

I just checked Cinderella and it does seem to do that *some* of the time. As you move the *defining point* around in the display port the result is `infinity invalid` about half the time. I had grid snapping turned on. With both lines at 45&deg; this most noticeable with `infinity invalid` always in the -x+y and +x-y quadrants. The `*.homog` value is [0,0,0] when this happens.

That is now Cinderella [issue 112](https://gitlab.cinderella.de:8082/cinderella/driftin/issues/112).

gagern:
We shouldn't have to do any special case handling in the code, it should be just automatic. @strobelm said he'd be looking into this as well.

elkins0:
@gagern, @strobelm: The ternary version of the AngularBisector code is now implemented as I envisioned it. The `geoMacros.AngularBisector` code has support for leaving off the point should you desire this feature. Anyone could just pass in a pinned, invisible point with coordinates of [0,0,0] to achieve the same result so I don't truly see the need to keep that feature.

gagern:
I had a cursory look at the code, and I think it has too many case distinctions, too many local functions, and more complexity than I'd like for this task. It's also pretty far away from how we tend to think about this problem in the development team, which is in terms of cross ratios involving the ideal circle points. Let me or @strobelm come up with an implementation and then we can better compare their respective merits and behavior in corner cases.

gagern:
I had a closer look. I also did think a bit more about the problem. @strobelm I'd like your opinion on what I write below. Here is what I'd expect from an ideal solution, given input *g, h, P*:

1. If *g, h* are distinct and finite, and *P* ist their finite point of intersection,  
then do the obvious thing.
2. If *g=h* are identical and finite, and *P* is a finite point incident to both of them,  
then return one line identical to *g=h* and one which is orthogonal in *P*.
3. If *g* and *h* are finite, distinct but parallel, and *P* is the corresponding point at infinity,  
then return one line which is half way between them and one which is the line at infinity.
4. If *g=h* are identical and finite, and *P* is their far point,  
then return one line which is identical to *g=h* and one which is the line at infinity.
5. If *g* is the line at infinity and *h* is finite, and furthermore *P* is their intersection at infinity,  
then return the line at infinity twice.  
This may need a bit more thorough verification but I believe that's what you get from limit processes.
6. If *g=h* are identical to one another and to the line at infinity,  
then the result is undefined, i.e. twice the null vector.
7. If any of *g, h, P* is a null vector,  
then the result is twice the null vector.

An ideal solution would achieve all of the above without introducing a case distinction using the same distinctions as above. That's because by placing the branch points of the function in some other locations in the parameter space, we can ensure that the cases distinguished above are the logical continuation of more general cases. This might be hard to achieve, and if it is, we should be particularly careful to make sure that my expectations above are indeed justified, and don't depend on a particular limit process.

Your method of normalizing the direction vectors of the lines works nicely for finite and distinct lines. The special case you have would correctly cover many of the problematic cases. As far as I can tell, all the cases where *g* and *h* are finite should be handled correctly. So perhaps case 5 above is the only problematic situation. I'd like to test all those cases first, and see whether I can think of more relevant cases than those mentioned above.

One solution we started from would be using cross ratios of points. If you take *G* and *H* as the intersections of *g* and *h* with the line ar infinity, then CR(*G*, *M*; *I*, *J*) = CR(*M*, *H*; *I*, *J*) describes the far point *M* through which the bisectors must pass. Joining *M* with *P* will give the bisectors themselves. This nicely covers the case of *g=h*, but it breaks down terribly if *P* is at infinity, since in that case *M=P* and the line joining them becomes undefined.

Another way to base this on cross ratios would be to use the cross ratio of lines. This is what I originally had in mind. You could combine *g* and *h* with the lines joining *P* to *I* resp. *J*. The current implementation is based on this. It starts with `poi = List.normalizeMax(List.cross(xx.homog, yy.homog))` so it would be easy to replace this by the third argument. Then it uses `myI` and `myJ` to refer to these lines joining *P* to *I* resp. *J*, and computes the result as a linear combination of these. However, if *P* is at infinity, then `myI` and `myJ` will be the line at infinity, and the general case can't handle those situations. But there is some more code for handling degenerate cases. @strobelm, can you elaborate the rationale behind that part, and which of the cases it covers?

I think we definitely should set up some testing framework, to automatically (i.e. without manually inspecting some HTML file) ensure that any implementation we choose actually has the behavior we expect in all the cases mentioned above.

strobelm:
Hi,

well the code generates the angle bisector as a appropriate linear
combination of the connecting lines from I and J to the intersection of
g and h. If now these connecting lines coincide (which happens if the
intersection of g and h is on l_inf) one of resulting lines will be
l_infty and the other the zero vector. Therefor the special case (for O
= cross(g,h)):

p = [0,0,1] and
q = [O,g,l_inf]*h+[O,h,l_inf]*g

I have a proof for that somewhere for the the special case formula which
yields cross-ration -1.

This also means if m / l are the l_inf we have one 0 prefactor and
return also l_inf (Case 5).
Case 6/7 is currently not covered in our code (simply there is no check
for the zero vector), we should add this with the rewrite.

I will have a look at the proposed code tomorrow.

On 01/26/2016 05:58 PM, Martin von Gagern wrote:
>
> I had a closer look. I also did think a bit more about the problem.
> @strobelm <https://github.com/strobelm> I'd like your opinion on what
> I write below. Here is what I'd expect from an ideal solution, given
> input /g, h, P/:
>
>  1. If /g, h/ are distinct and finite, and /P/ ist their finite point
>     of intersection,
>     then do the obvious thing.
>  2. If /g=h/ are identical and finite, and /P/ is a finite point
>     incident to both of them,
>     then return one line identical to /g=h/ and one which is
>     orthogonal in /P/.
>  3. If /g/ and /h/ are finite, distinct but parallel, and /P/ is the
>     corresponding point at infinity,
>     then return one line which is half way between them and one which
>     is the line at infinity.
>  4. If /g=h/ are identical and finite, and /P/ is their far point,
>     then return one line which is identical to /g=h/ and one which is
>     the line at infinity.
>  5. If /g/ is the line at infinity and /h/ is finite, and furthermore
>     /P/ is their intersection at infinity,
>     then return the line at infinity twice.
>     This may need a bit more thorough verification but I believe
>     that's what you get from limit processes.
>  6. If /g=h/ are identical to one another and to the line at infinity,
>     then the result is undefined, i.e. twice the null vector.
>  7. If any of /g, h, P/ is a null vector,
>     then the result is twice the null vector.
>
> An ideal solution would achieve all of the above without introducing a
> case distinction using the same distinctions as above. That's because
> by placing the branch points of the function in some other locations
> in the parameter space, we can ensure that the cases distinguished
> above are the logical continuation of more general cases. This might
> be hard to achieve, and if it is, we should be particularly careful to
> make sure that my expectations above are indeed justified, and don't
> depend on a particular limit process.
>
> Your method of normalizing the direction vectors of the lines works
> nicely for finite and distinct lines. The special case you have would
> correctly cover many of the problematic cases. As far as I can tell,
> all the cases where /g/ and /h/ are finite should be handled
> correctly. So perhaps case 5 above is the only problematic situation.
> I'd like to test all those cases first, and see whether I can think of
> more relevant cases than those mentioned above.
>
> One solution we started from would be using cross ratios of points. If
> you take /G/ and /H/ as the intersections of /g/ and /h/ with the line
> ar infinity, then CR(/G/, /M/; /I/, /J/) = CR(/M/, /H/; /I/, /J/)
> describes the far point /M/ through which the bisectors must pass.
> Joining /M/ with /P/ will give the bisectors themselves. This nicely
> covers the case of /g=h/, but it breaks down terribly if /P/ is at
> infinity, since in that case /M=P/ and the line joining them becomes
> undefined.
>
> Another way to base this on cross ratios would be to use the cross
> ratio of lines. This is what I originally had in mind. You could
> combine /g/ and /h/ with the lines joining /P/ to /I/ resp. /J/. The
> current implementation is based on this. It starts with |poi =
> List.normalizeMax(List.cross(xx.homog, yy.homog))| so it would be easy
> to replace this by the third argument. Then it uses |myI| and |myJ| to
> refer to these lines joining /P/ to /I/ resp. /J/, and computes the
> result as a linear combination of these. However, if /P/ is at
> infinity, then |myI| and |myJ| will be the line at infinity, and the
> general case can't handle those situations. But there is some more
> code for handling degenerate cases. @strobelm
> <https://github.com/strobelm>, can you elaborate the rationale behind
> that part, and which of the cases it covers?
>
> I think we definitely should set up some testing framework, to
> automatically (i.e. without manually inspecting some HTML file) ensure
> that any implementation we choose actually has the behavior we expect
> in all the cases mentioned above.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/183#issuecomment-175115285>.
>


-- 
Michael Strobel, M.Sc.

strobelm:
Hi, the results seem to be correct (at least for finite points) therefor i plugged in the formulas to see if the math is correct too. Sage tells me that is not the case, but i can't rule out that i got something wrong. Here is the code:

```
sage: a1,a2,a3,b1,b2,b3 = var('a1 a2 a3 b1 b2 b3', domain='real')
sage: a = vector([a1,a2,a3]); b = vector([b1,b2,b3])
sage: M = a.cross_product(b)
sage: inf = vector([0,0,1])
sage: l = sqrt(a1*a1+a2*a2)*b + sqrt(b1*b1+b2*b2)*a;    m = -sqrt(a1*a1+a2*a2)*b + sqrt(b1*b1+b2*b2)*a
sage: linf = l.cross_product(inf); minf = m.cross_product(inf)
sage: O = vector([0,0,1])
sage: def crossr(a, b, c, d, o):
...   	re1 = det(Matrix([o,a,c]))*det(Matrix([o,b,d]))
...   	re2 = det(Matrix([o,a,d]))*det(Matrix([o,b,c]))
...   	return re1/re2
sage: II = vector([-I, 1,0]);JJ = vector([I, 1,0]);
sage: erg = (crossr(linf,M,II,JJ,O) == crossr(M,minf,II,JJ,O))
sage: erg.test_relation()
False
```

strobelm:
Mathematica says its correct, cool. I'm wondering if this works for other n-sectors.

```
In[1]:= l = {ll1, ll2, ll3}; m = {mm1, mm2, mm3}; inf = {0, 0, 1};

In[2]:= L = Cross[l, inf]; M = Cross[m, inf];

In[4]:= l1 = Sqrt[ll1*ll1 + ll2*ll2]*m + Sqrt[(mm1*mm1 + mm2*mm2)]*l;


In[5]:= l2 = Sqrt[ll1*ll1 + ll2*ll2]*m - Sqrt[(mm1*mm1 + mm2*mm2)]*l;

In[6]:= A1 = Cross[l1, inf];

In[7]:= A2 = Cross[l2, inf];

In[8]:= II = {-I, 1, 0}; JJ = {I, 1, 0};

In[9]:= OO = {0, 0, 1};

In[10]:= A1inf = Cross[A1, inf]; A2inf = Cross[A2, inf];

In[11]:= cr1 = (Det[{OO, L, II}]*
     Det[{OO, A1inf, JJ}])/ (Det[{OO, L, JJ}]*Det[{OO, A1inf, II}]);

In[12]:= cr2 = (Det[{OO, A1inf, II}]*
     Det[{OO, M, JJ}])/ (Det[{OO, A1inf, JJ}]*Det[{OO, M, II}]);

In[13]:= Simplify[cr1 - cr2]

Out[13]= 0
```

elkins0:
@gagern wrote:
>I had a cursory look at the code, and I think it has too many case distinctions, too many local functions, and more complexity than I'd like for this task. It's also pretty far away from how we tend to think about this problem in the development team, which is in terms of cross ratios involving the ideal circle points.

In response to that critique :

1. Flattened local functions.
2. Fixed imaginary results based on consulting *Mathematica*'s `Normalize` function and the coding of `List.projectiveDistMinScal` by replacing the `CSNumber.sqrt(CSNumber.add(CSNumber.mult(line.value[0], line.value[0]), CSNumber.mult(line.value[1], line.value[1])))` code pattern with the `List.abs(List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero]))` code pattern.
3. Choose higher-level code patterns over performance, replacing the `List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero])` code pattern with the `List.cross(List.cross(List.linfty, line), List.linfty)` code pattern in four places.
4. Used the `var func = List.func;` code pattern to make the code portion more readable.

```javascript
geoOps.angleBisector.updatePosition = function(el) {
    var a = csgeo.csnames[el.args[0]].homog;
    var b = csgeo.csnames[el.args[1]].homog;
    var p = csgeo.csnames[el.args[2]].homog;
    var add = List.add;
    var sub = List.sub;
    var abs = List.abs;
    var cross = List.cross;
    var sm = List.scalmult;
    var nm = List.normalizeMax;
    var isAlmostZero = List._helper.isAlmostZero;
    var linfty = List.linfty;
    var na = sm(abs(cross(cross(linfty, b), linfty)), a);
    var nb = sm(abs(cross(cross(linfty, a), linfty)), b);
    var res1 = sub(na, nb);
    var res2 = add(na, nb);
    if (isAlmostZero(res1)) res1 = cross(cross(cross(linfty, res2), linfty), p);
    if (isAlmostZero(res2)) res2 = cross(cross(cross(linfty, res1), linfty), p);
    el.results = tracing2(nm(res1), nm(res2));
};
```
Applying the same treatment to `geoOps.Perp.updatePosition` would result in:
```javascript
geoOps.Perp.updatePosition = function(el) {
    var l = csgeo.csnames[(el.args[0])].homog;
    var p = csgeo.csnames[(el.args[1])].homog;
    var cross = List.cross;
    var linfty = List.linfty;
    var nm = List.normalizeMax;
    el.homog = General.withUsage(nm(cross(cross(cross(linfty, l), linfty), p)), "Line");
};
```

elkins0:
@strobelm wrote:
```sage
sage: l1,l2,l3,m1,m2,m3 = var('l1 l2 l3 m1 m2 m3', domain='real')
sage: l = vector([l1,l2,l3]); m = vector([m1,m2,m3])
sage: inf = vector([0,0,1])
sage: L = l.cross_product(inf); M = m.cross_product(inf)
sage: l1 = sqrt(l1*l1+l2*l2)*m + sqrt(m1*m1+m2*m2)*l;   l2 = -sqrt(l1*l1+l2*l2)*m + sqrt(m1*m1+m2*m2)*l
sage: A1 = l1.cross_product(inf); A2 = l2.cross_product(inf)
sage: O = vector([0,0,1])
sage: def crossr(a, b, c, d, o):
...     re1 = det(Matrix([o,a,c]))*det(Matrix([o,b,d]))
...     re2 = det(Matrix([o,a,d]))*det(Matrix([o,b,c]))
...     return re1/re2
sage: II = vector([-I, 1,0]);JJ = vector([I, 1,0]);
sage: erg1 = simplify(crossr(L,A1,II,JJ,O) - crossr(A1,M,II,JJ,O))
sage: erg2 = simplify(crossr(L,A2,II,JJ,O) - crossr(A2,M,II,JJ,O))
```
The differences I can see is that in your *Sage* versus *Mathematica* input are:

1. `var('l1 l2 l3 m1 m2 m3', domain='real')` whereas in *Mathematica* all variables can be complex
2. Additional `cross_product(inf)` on `A1` and `A2` which changes the signs on `erg1` and `erg2`.
3. ` l2 = sqrt(l1*l1+l2*l2)*m - sqrt(m1*m1+m2*m2)*l ` which has no effect on `erg1` and `erg2`.

I reworked your *Mathematica* code to use a flattened version of `List.crossratio3`:
```Mathematica
Block[{crossratio3,l,m,inf,L,M,l1,l2,A1,A2,II,JJ,OO,cr1,cr2,cr3,cr4},
crossratio3=Function[{a,b,c,d,x},Det[{a,c,x}]*Det[{b,d,x}]/(Det[{a,d,x}]*Det[{b,c,x}])];
 l={ll1,ll2,ll3};
 m={mm1,mm2,mm3};
 inf={0,0,1};
 L=Cross[l,inf];
 M=Cross[m,inf];
 l1=Sqrt[ll1*ll1+ll2*ll2]*m+Sqrt[(mm1*mm1+mm2*mm2)]*l;
 l2=Sqrt[ll1*ll1+ll2*ll2]*m-Sqrt[(mm1*mm1+mm2*mm2)]*l;
 II={-I,1,0};
 JJ={I,1,0};
 OO={0,0,1};
 A1=Cross[Cross[l1,inf],inf];
 A2=Cross[Cross[l2,inf],inf];
 cr1=FullSimplify[crossratio3[L,A1,II,JJ,OO]];
 cr2=FullSimplify[crossratio3[A1,M,II,JJ,OO]];
 cr3=FullSimplify[crossratio3[L,A2,II,JJ,OO]];
 cr4=FullSimplify[crossratio3[A2,M,II,JJ,OO]];
 {cr1==cr2,cr3==cr4,cr1==-cr3,cr4}]
{True, True, True, ((ll1 + I*ll2)*(mm1 - I*mm2))/
    (Sqrt[ll1*ll1 + ll2*ll2]*Sqrt[(mm1*mm1 + mm2*mm2)])}
```

elkins0:
Oh...and this doesn't quite match `List.crossratio3`:
```sage
sage: def crossr(a, b, c, d, o):
...     re1 = det(Matrix([o,a,c]))*det(Matrix([o,b,d]))
...     re2 = det(Matrix([o,a,d]))*det(Matrix([o,b,c]))
...     return re1/re2
```

gagern:
Status update: I haven't forgotten about this. I intend to write a testing framework which allows us to elegantly write unit tests for geometric primitives. Once I have that, I'd like to use that to check [the points I mentioned above](#issuecomment-175115285) for all proposed implementations, and take it from there.

strobelm:
Hi! I was thinking about this, too. The current code breaks 82_angle_bisector.html since it expects now a third argument, this makes me wonder how to deal with (almost) optional arguments in geoOps. I think current signature handling can't deal with this.

gagern:
@strobelm We could allow for signature elements which contain some special character like `?` or `=` to indicate optional arguments. I've seen `Angle` exported from Cinderella with only two arguments, too. Or as an alternative we could allow the same names for macros and implementations, and use the macro version to unify input, possibly introducing a third argument as needed. In the case of angle bisectors, the previous name is different from the one Cinderella uses, so we could use macros to provide the missing point.

strobelm:
Hi, i just thought about default arguments. I think it worth discussing these before we continue with the bisector:
https://github.com/CindyJS/CindyJS/issues/229

gagern:
I've picked parts of this to #237 and intend to continue discussion there if you don't mind. Personally I'm satisfied with the code, so I'll most likely let @strobelm review those parts which I wrote.

strobelm:
Since we merged #237 this can be closed.

Issue #95, opened 2015-11-24
Programatically setting positions for lines


gagern:
It would be nice if we could set the position of a line from CindyScript. In particular the completely free `FreeLine` and the semi-free lines `Through`, `HorizontalLine` and `VerticalLine`. We already have write access for the `angle` of a `Through`, but changing `homog` should be made to work as well, I think.

gagern:
Some pointers: [`Accessor.setField`](https://github.com/CindyJS/CindyJS/blob/0fb147c879568a98de0f3dbe8615c98791bdbd01/src/js/libcs/Accessors.js#L167-L180) where we'd have to add the additional cases and [`Through.getParamForInput`](https://github.com/CindyJS/CindyJS/blob/0fb147c879568a98de0f3dbe8615c98791bdbd01/src/js/libgeo/GeoOps.js#L186-L193) for an example of how to distinguish between mouse and script input.

kranich:
This is related to #282, #284, #285, #286, #287.

kranich:
@gagern How would you like to update homogeneous coordinates of `HorizontalLine` and `VerticalLine`? Set first respectively second coordinate entry to zero?

PR #184, opened 2016-01-24
Allow manipulating lines in CindyJS

elkins0:
A call to `Accessor.setField(geo, "angle", value)` on a line through a point results in a call `movepointscr(geo, dir, "dir")`. `geoOps.Through.getParamForInput` did not support `"dir"` as a valid type. Also CindyScript couldn't set a line's `homog` value.

I ran into this problem when I was instrumenting an example file for #183. The instrumented example using JavaScript and CindyScript to manipulate lines and points in CindyJS is [AngularBisectorCS.html](https://gist.github.com/elkins0/0baf165f01ff9440c9b4)

### DIFF
--- a/src/js/libcs/Accessors.js
+++ b/src/js/libcs/Accessors.js
@@ -181,7 +181,7 @@ Accessor.setField = function(geo, field, value) {
     }
 
 
-    if (field === "homog" && geo.kind === "P" && geo.movable && List._helper.isNumberVecN(value, 3)) {
+    if (field === "homog" && (geo.kind === "P" || geo.kind === "L") && geo.movable && List._helper.isNumberVecN(value, 3)) {
         movepointscr(geo, value, "homog");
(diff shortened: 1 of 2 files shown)


elkins0:
An instrumented example using JavaScript and CindyScript to manipulate lines and points in CindyJS is [AngularBisectorCS.html](https://gist.github.com/elkins0/0baf165f01ff9440c9b4). This allow lines a and b to be set parallel at the push of a button. You'll need #184 for that to work.

elkins0:
This ties into *Programmatically setting positions for lines* #95.

gagern:
This pull request appears to be superseded by #286. Sorry it took me so long to properly look at this, and thanks for the contribution nonetheless.

gagern:
Can't remember why I repoened this, so I guess it was by accident. Closing again.

Issue #121, opened 2015-12-02
Native convexhull3d


kranich:
As suggested in #51, instead of using GWT to transpile a Java library, we may want to reimplement `convexhull3d` in JavaScript.

kranich:
Fixed by #584.

PR #191, opened 2016-01-26
Restructuring plugins

### LINKED ISSUE #186: Restructuring plugins
We currently have three plugins, namely `cindy3d`, `cindygl`, and `katex`.

Their sources reside in [src/js/](https://github.com/CindyJS/CindyJS/tree/master/src/js/) and  [src/str/](https://github.com/CindyJS/CindyJS/tree/master/src/str/).

I think it would be cleaner to move plugins to a `plugins` directory at the root of the repository, with a subdirectory for every plugin.

This would also make it much easier to integrate plugins that are developed in a separate repository.

What do you think?

kranich:
This should fix #186.


kranich:
@gagern Please do not merge yet. Plugin examples are broken.

gagern:
Symlinks are problematic, as I just explained in https://github.com/CindyJS/CindyJS/issues/186#issuecomment-174958362. This commit also doesn't move the `quickhull3d` plugin.

kranich:
OK, then let's leave the examples where they were for the time being.
I will also move the `quickhull3d` plugin as suggested.
I will rebase accordingly.

kranich:
>This commit also doesn't move the `quickhull3d` plugin.

As far as I see it, `quickhull3d` is not really a plugin. No `use` needed, no separate source file to be included. Once #121 has been addressed, `quickhull3d` is definitely part of CindyJS core.

@gagern Please feel free to move `quickhull3d` yourself if you consider it reasonable. I just do not know how you would like that to be handled.

kranich:
@gagern Please note that due to the move the plugins are no longer covered by build targets `beautify`, `jshint`, and `alltests`. Is this side-effect desirable or should we change this?

gagern:
> Please note that due to the move the plugins are no longer covered by build targets `beautify`, `jshint`, and `alltests`. Is this side-effect desirable or should we change this?

@kranich I see no change here. [`beautify`](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L94-L102) operates on [`sources.ours`](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/sources.js#L43) which does not include the plugins. [`jshint`](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L108-L115) operates on `ours.js` which is [built](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L43-L45) by concatenating that same set of sources. [`alltests`](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L155-L163) is an alias for several other tasks, none of which sould be affected by this move.

The most reasonable way such a move could affect a file is by a change in glob matching. We use globs for [the reference manual](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L186-L187), [glsl shader sources](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L305), [gwt modules](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L379-L382), [katex sources](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L426-L427) and – [via the `forbidden` command](https://github.com/CindyJS/CindyJS/blob/master/make/commands.js#L186) – in [the forbidden patterns for examples](https://github.com/CindyJS/CindyJS/blob/5be655ca6f42e9cfd45fb7873b542c2cba30bb9d/make/build.js#L143-L148).

At least for the Cindy3D plugin, this fact is quite deliberate. There I like to maintain my own coding style, which differs from what beautify does to the rest of the code. In particular, I only use 2 spaces indentation (to make deeply nested code more maintainable) and ES2015 syntax. I also use advanced compilation from the closure compiler, which usually gives even better warnings than jshint would. @montaga adopted many of my conventions since he based his plugin off of the one for Cindy3D. The KaTeX plugin could be subjected to these checks, but that would be a different issue since the move doesn't change the status quo.

kranich:
@gagern OK. Then this is ready to merge, in my opinion.

kranich:
@gagern Thanks!

PR #195, opened 2016-01-31
Change default lineJoin to round

gagern:
This fixes a regression for drawpoly.  In the long run modifiers to change line join and line cap style would be nice, but this here is just a hot fix.


gagern:
Superseded by #196.

Issue #156, opened 2016-01-10
Example using CindyJS showing an unfortunate result of normalizeMax


elkins0:
[IntersectionOfSameLines.html](https://gist.github.com/elkins0/bbdd9a8f2e5dbea559df) shows an example where the expected result for the homogeneous coordinates for a point (E) should be [0,0,0] but instead end up as [1,-0.5,-0.125] because normalizeMax amplifies the numerical residue into an invalid real result. This does not occur in Cinderella. This sample JavaScript code performs the same calculation giving the same wrong results:
```javascript
//The debugger shows these real values (simplified form):
var a = [0.11111111111111109, -0.22222222222222224, 1];
document.write("<br>a = [" + a + "]");
var b = [0.1111111111111111, -0.2222222222222222, 1];
document.write("<br>b = [" + b + "]");
// (a - b)
var d = [a[0] - b[0], a[1] - b[1], a[2] - b[2]];
document.write("<br>(a - b) = [" + d + "]");
// cross(a,b)
var c = [a[1] * b[2] - a[2] * b[1], a[2] * b[0] - a[0] * b[2], a[0] * b[1] - a[1] * b[0]];
document.write("<br>cross(a,b) = [" + c + "]");
var max = c.reduce(function(a, b, i, to) { return Math.abs(a) > Math.abs(b) ? a : b; }, 0);
document.write("<br>max = ", max);
var E = c.map(function(a) { return a / max; });
document.write("<br>E = [" + E + "]");
var twoToMinus52 = Math.pow(2, -52);
document.write("<br>2^-52 = " + twoToMinus52);
document.write("<br>(1+2^-52)-1 = " + ((1+twoToMinus52)-1));
var twoToMinus53 = Math.pow(2, -53);
document.write("<br>2^-53 = " + twoToMinus53);
document.write("<br>(1+2^-53)-1 = " + ((1+twoToMinus53)-1));
```
This is the result of the JavaScript:
```
a = [0.11111111111111109,-0.22222222222222224,1]
b = [0.1111111111111111,-0.2222222222222222,1]
(a - b) = [-1.3877787807814457e-17,-2.7755575615628914e-17,0]
cross(a,b) = [-2.7755575615628914e-17,1.3877787807814457e-17,3.469446951953614e-18]
max = -2.7755575615628914e-17
E = [1,-0.5,-0.125]
2^-52 = 2.220446049250313e-16
(1+2^-52)-1 = 2.220446049250313e-16
[...]

PR #199, opened 2016-01-31
Clean-up CSNumber.log coding

elkins0:
Adds CSNumber.log tests for 0 and 8 values on the complex unit circle because CSNumber.log appears to be cluttered with unnecessary special case code. The CSNumber_tests.js tests are submitted first to record that the original code passes before making any changes to CSNumber.log.

### DIFF
--- a/src/js/libcs/CSNumber.js
+++ b/src/js/libcs/CSNumber.js
@@ -434,27 +434,11 @@ CSNumber.powRealExponent = function(a, b) {
 CSNumber.log = function(a) {
     var re = a.value.real;
     var im = a.value.imag;
-    var s = Math.sqrt(re * re + im * im);
-    var i = im;
-
-
-    var imag = Math.atan2(im, re);
-    if (i < 0) {
-        imag += (2 * Math.PI);
-    }
-    if (i === 0 && re < 0) {
-        imag = Math.PI;
-    }
-    if (imag > Math.PI) {
-        imag -= (2 * Math.PI);
-    }
-    var real = Math.log(s);
-
     return CSNumber.snap({
         "ctype": "number",
         "value": {
-            'real': real,
-            'imag': imag
+            'real': Math.log(Math.sqrt(re * re + im * im)),
+            'imag': Math.atan2(im, re)
         }
     });
 };
--- a/tests/CSNumber_tests.js
+++ b/tests/CSNumber_tests.js
@@ -32,6 +32,15 @@ var f_a = CSNumber.complex(100, -0.5);
 var f_b = CSNumber.complex(-0.5, 1);
 var f_a_plus_b = CSNumber.add(f_a, f_b);
 var f_a_mult_f_b = CSNumber.complex(-49.5,100.25);
+var f_0 = CSNumber.real(0);
+var f_minus_rh2_minus_rh2i = CSNumber.complex(-Math.SQRT1_2, -Math.SQRT1_2);
+var f_minus_1i = CSNumber.complex(0, -1);
+var f_rh2_minus_rh2i = CSNumber.complex(Math.SQRT1_2, -Math.SQRT1_2);
+var f_1 = CSNumber.real(1);
+var f_rh2_plus_rh2i = CSNumber.complex(Math.SQRT1_2, Math.SQRT1_2);
+var f_1i = CSNumber.complex(0, 1);
+var f_minus_rh2_plus_rh2i = CSNumber.complex(-Math.SQRT1_2, Math.SQRT1_2);
+var f_minus_1 = CSNumber.real(-1);
 
 
 // basic operations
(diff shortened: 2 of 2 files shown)


elkins0:
OK. [Build #416](https://travis-ci.org/CindyJS/CindyJS/builds/106100177) passes.
```
487 [unittests   ]     ✓ log
```

elkins0:
Examining this portion of `CSNumber.log`,
```javascript
    var imag = Math.atan2(im, re);
```
`Math.atan2` produces a result which lies between `-Math.PI` and `0` when `im` is less than `0`.
```javascript
    if (im < 0) {
        imag += (2 * Math.PI);
    }
```
So now we have a result a result which lies between `Math.PI` and `(2 * Math.PI)` when `im` is less than `0`.

`Math.atan2` produces a result which is `-Math.PI` when `im` is `0` and `re` is less than `0`.
```javascript
    if (im === 0 && re < 0) {
        imag = Math.PI;
    }
```
So this code does nothing.

`Math.atan2` produces a result which lies between `0` and `Math.PI` when `im` is greater than `0`.
So this code only comes into play when `im` is less than `0`:
```javascript
    if (imag > Math.PI) {
        imag -= (2 * Math.PI);
    }
```
So the result is to subtract the `(2 * Math.PI)` added earlier in the coding back off again.

gagern:
I like the idea of cleaning this up, and the fact that you added a test case to verify the change! The test case itself is a bit hard to read. Perhaps we want a function to run the test, something like

```js
function expectComplex(expectedReal, expectedImag, actual) {
  var expected = CSNumber.complex(expectedReal, expectedImag);
  expect(CSNumber._helper.isAlmostEqual(expected, actual)).toBe(true);
}
```

which collects all the red tape and allows you to concentrate on the actual issue at hand in each line of the test. But that's not a core issue here, I think.

Looking at [the ECMAScript 5.1 spec for `atan2`](http://www.ecma-international.org/ecma-262/5.1/#sec-15.8.2.5) I see that things are somewhat more subtle than they appear at first glance. In particular:

> * If *y* is +0 and *x*<0, the result is an implementation-dependent approximation to +*π*.
> * If *y* is −0 and *x*<0, the result is an implementation-dependent approximation to −*π*.

So the *sign* of the zero may decide which imaginary portion we may expect here. As far as I understand the previous code, we would *never* end up with a result of `-Math.PI`, since after the first two special cases we have `imag >= 0` and the last case won't apply if `imag == Math.PI`. I wonder whether this is a property worth preserving. I can't imagine why, but would welcome a second opinion.

I'm also a bit concerned about the `CSNumber.snap` call there. It has been [present](https://github.com/CindyJS/CindyJS/blob/e26335c9dfa838e8fb73d46451a7987badde5e1a/lib/Number.js#L80-L101) since e26335c9dfa838e8fb73d46451a7987badde5e1a, which is since the function was first introduced. Was this a deliberate choice, or some copy&paste from `Number.mod`? If it was deliberate, what's the rationale? @richter-gebert, do you remember what led to that line?

elkins0:
If it is unavoidable, `if (imag === -Math.PI) imag = -imag;` might be added, however be aware that while it may be possible for CSNumber.log to force the range -Math.PI < imag <= Math.PI any negation of imag latter results in the range -Math.PI <= imag < Math.PI.

I'm all for removing snap from CSNumber.log. Notice that the 1e-7 setting of CSNumber.eps near-by only to be heedlessly reset 260 lines latter to 1e-10!
```javascript
CSNumber.eps = 0.0000001;

CSNumber.snap = function(a) {

... // About 260 lines later on the CSNumber.eps value is replaced:

CSNumber._helper.seed = 'NO';

CSNumber.eps = 0.0000000001;
CSNumber.epsbig = 0.000001;

```
The proximity suggests the original effect intended may have long ago been compromised. Certainly there is no effect at all on the values generated in any of the current CSNumber.log tests. In fact when I tried using `isEqual` instead of `isAlmostEqual` there were 9 of 11 passes with the old code and 10 of 11 passes with the new code. The old code's add/subtract causes a loss of accuracy in the second test and while both new and old match each other on the first test, the value expected is a high-precision value which is a tiny bit different so that is not identical.

Also you probably wouldn't want to snap the imag value to just any integer other than zero since its value is typically some fraction of *π*.

I don't believe having CSNumber.snap in CSNumber.log helps resolve problems such as *Detect degenerate situations* #57 or *Example using CindyJS showing an unfortunate result of normalizeMax* #156.

As to `-0` and `+0`, I did some checking and found that `-1/Infinity`, negation of a variable set to zero `-(0)` and multiplying `-#*0` or dividing `0/-#` by a negative value all generate a `-0`. In general addition and subtraction produce positive zero. The exceptions are `(-0) + (-0)` and `(-0) - (0)`. And here's the rub, `(-0) === (0)` is `true`! The only way to test for a `-0` is "1 / (-0) === -Infinity`.

strobelm:

PR #200, opened 2016-01-31
bugfix: colorplot on HiDPI displays was not handled correctly

montaga:
Now the 1-adic colorplot renders in "real resolution" instead of CSS-resolution. Therefore content looks nicer on HiDPI-displays.

### DIFF
--- a/examples/cindygl/18_hidpitest.html
+++ b/examples/cindygl/18_hidpitest.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


kranich:
I am not sure whether this is the right way to handle this. You might want to have a look at http://www.html5rocks.com/en/tutorials/canvas/hidpi/

montaga:
The steps mentioned in the tutorial above were already done in `src/js/Setup.js`. Hence it also works fine in the example as well.

What has been changed in my commit is that the internal rendering resolution is now not the hardware resolution, i.e. the resolution of the "backing store", instead of the logical resolution.

PR #201, opened 2016-02-01
Upgrade Google Closure Compiler to latest version (v20160208)

kranich:
This pull request updates Google Closure Compiler to the latest version.

Among the benefits are:

* official support for ECMAScript 6 (since v20151015)
* many ES6 related bug fixes
* support for ES6 modules (opt-in)
* slightly smaller output size

### DIFF
--- a/make/Settings.js
+++ b/make/Settings.js
@@ -18,7 +18,7 @@ module.exports = function Settings() {
         closure_urlbase: "http://dl.google.com/closure-compiler",
         closure_language: "ECMASCRIPT5_STRICT",
         closure_level: "SIMPLE",
-        closure_version: "20150126",
+        closure_version: "20160201",
         verbose: "true",
         logprefix: "true",
         c3d_closure_level: "ADVANCED",
--- a/make/Settings.js
+++ b/make/Settings.js
@@ -18,7 +18,7 @@ module.exports = function Settings() {
         closure_urlbase: "http://dl.google.com/closure-compiler",
         closure_language: "ECMASCRIPT5_STRICT",
         closure_level: "SIMPLE",
-        closure_version: "20160201",
+        closure_version: "20160208",
         verbose: "true",
         logprefix: "true",
         c3d_closure_level: "ADVANCED",
--- a/make/build.js
+++ b/make/build.js
@@ -269,6 +269,7 @@ module.exports = function build(settings, task) {
         var opts = {
             language_in: "ECMASCRIPT6_STRICT",
             language_out: "ECMASCRIPT5_STRICT",
+            dependency_mode: "LOOSE",
             create_source_map: "build/js/Cindy3D.js.map",
             compilation_level: this.setting("c3d_closure_level"),
             warning_level: this.setting("c3d_closure_warnings"),
@@ -285,7 +286,7 @@ module.exports = function build(settings, task) {
             })),
         };
         if (this.setting("cindy3d-dbg") !== undefined) {
-            opts.transpile_only = true;
+            opts.compilation_level = "WHITESPACE_ONLY";
             opts.formatting = "PRETTY_PRINT";
         }
         this.closureCompiler(closure_jar, opts);
@@ -336,6 +337,7 @@ module.exports = function build(settings, task) {
         var opts = {
             language_in: "ECMASCRIPT6_STRICT",
             language_out: "ECMASCRIPT5_STRICT",
+            dependency_mode: "LOOSE",
             create_source_map: "build/js/CindyGL.js.map",
             compilation_level: this.setting("cgl_closure_level"),
             warning_level: this.setting("cgl_closure_warnings"),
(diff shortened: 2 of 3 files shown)


gagern:
You changed `--transpile_only` to `--compilation_level=WHITESPACE_ONLY`. According to google/closure-compiler@8d8fe2a216a25c388e46e8095bbb3a095b096103, the correct replacement should be `--skip_non_transpilation_passes`. It is my experience that `WHITESPACE_ONLY` will *not* transpile, i.e. will result in ES6 output which causes problems since last I checked, there was no was to include ES6 sources that would work for all browsers.

Looking at the `Cindy3D.js` resulting from a simple `node make clean cindy3d`, I see that it looks *way* too short:

```js
(function(){
'use strict';
})();//# sourceMappingURL=Cindy3D.js.map
```

Did you observe anything different than this? Or didn't you test this? If you don't observe this outcome, we should investigate why I do. If you do as well, we should figure out what's causing this, whether we can use the closure compiler without entry points or whether we have to provide entry points these days.

kranich:
>You changed `--transpile_only` to 
>`--compilation_level=WHITESPACE_ONLY`. According to 
>google/closure-compiler@8d8fe2a216a25c388e46e8095bbb3a095b096103, the 
>correct replacement should be `--skip_non_transpilation_passes`. It is 
>my experience that `WHITESPACE_ONLY` will *not* transpile, i.e. will 
>result in ES6 output which causes problems since last I checked, there 
>was no was to include ES6 sources that would work for all browsers.

OK, I just changed/rebased that.

>Looking at the `Cindy3D.js` resulting from a simple `node make clean cindy3d`, I see that it looks *way* too short:
>
>```js
>(function(){
>'use strict';
>})();//# sourceMappingURL=Cindy3D.js.map
>```
>
>Did you observe anything different than this? Or didn't you test this? If you don't observe this outcome, we should investigate why I do. If you do as well, we should figure out what's causing this, whether we can use the closure compiler without entry points or whether we have to provide entry points these days.

I had not looked at `Cindy3D.js`. That's disappointing! Same for 
`CindyGL.js`. We need to investigate this, as you have said.

kranich:
Oops, now I get `--skip_non_transpilation_passes` is not a valid option when I run `make cindy3d-dbg`.

kranich:
Apparently, google/closure-compiler@8d8fe2a removed `--transpile_only` but forgot to add `--skip_non_transpilation_passes` instead. See google/closure-compiler#1479.

kranich:
@gagern At google/closure-compiler#1479, @dimvar has suggested to use `--compilation_mode WHITESPACE_ONLY` instead. My issue and pull request over there have been closed.

I will revert my previous rebase.

kranich:
`Cindy3D.js` and `CindyGL.js` are still empty. So this must not be merged right now.

kranich:
@gagern If we fix the error message `plugins/cindy3d/src/js/Ops3D.js:321: ERROR - variable k is undeclared` by adding `let k;` as line 309, then we can upgrade to `v20151216`. This does not truncate `Cindy3D.js` nor `CindyGL.js`. Should we do this?

gagern:
I'd say give it a bit more time to work out what's actually going on. I've got half a mind to git bisect this on the closure compiler repository. Let my try whether I can automate that…

gagern:
Bisected this to google/closure-compiler@62ca536d0a184a8852646e2bc2db178f700f5cfc. Looking at [its test case](https://github.com/google/closure-compiler/blob/62ca536d0a184a8852646e2bc2db178f700f5cfc/test/com/google/javascript/jscomp/CommandLineRunnerTest.java#L1790-L1794), perhaps we should prefix each JavaScript input individually with `--js`.

gagern:
Proposed solution works, see kranich#1 for PR. Since I don't think this is intentional, I also filed google/closure-compiler#1482 about it. I'd like to wait for a comment on that, and perhaps use a later version which again doesn't need those prefixes. The command line is already pretty long as it is.

kranich:
@gagern OK, thank you! You may want to get the fix in `plugins/cindy3d/src/js/Ops3D.js` into master sooner than that, though.

kranich:
v20160125 has been rolled back, see [Releases](https://github.com/google/closure-compiler/wiki/Releases). I will keep an eye on the releases.

kranich:
@gagern I have rebased and squashed our commits. We are now using `v20160201` which no longer needs the `--js` prefix for every source file. Once the Travis build has passed, this should be ready to merge.

gagern:
Travis doesn't check plugins, so I'll want to perform some manual checks there before accepting this.

kranich:
@gagern Latest version is now v20160208. I have just checked manually that all cindy3d and cindygl examples are working, so I think you can merge this now.

gagern:
Lacking time for more thorough testing, I'll take your word for it.

PR #208, opened 2016-02-03
Rudementary saving of geometric state

gagern:
This only saves the geometry portion of the current state, not the current assignment of variables or the properties of a physics simulation. So it's a first step. Since the API hasn't stabilized yet, there intentionally is no documentation. The idea is that the object returned from `cdy.saveState()` will gain properties as we learn how to store more and more aspects of internal state; at the moment there is only a `geometry` object which can be used in place of the original `geometry` component of the `createCindy` options object. One day we may have an accompanying `cdy.restoreState(savedState)` call to reinsert a saved state into a running instance.


strobelm:
I think this is heading in the right direction. I also like the small portions, way easier to see whats happening.

PR #209, opened 2016-02-04
all operations and guess more incidences

strobelm:
Hi,

i implemented the allpoints, alllines, allwhatever operations and for that i also updated guessIncidences for conics.

### DIFF
--- a/examples/114_allops.html
+++ b/examples/114_allops.html
@@ -1,0 +1,126 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Konstruktion.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+<script id='csmove' type='text/x-cindyscript'>
+println("allpoints, should be A, B, C, D, E, F, G, H, K, L, M, N, O");
+println(allpoints());
+println("alllines, should be a,b,c");
+println(alllines());
+println("allsegments, should be a,c");
+println(allsegments());
+println("allconics, should be C1, C2, C0");
+println(allconics());
+println("allcircles, should be C0, C2");
+println(allcircles());
+println("allmasses, should be M, O");
+println(allmasses());
+println("allsprings, should be c");
+println(allsprings());
+
+println("allelements");
+println(allelements());
+
+
+// INCIDENCES
+
+println("allpoints incident to b, should be K");
+println(allpoints(b));
+println("allpoints incident to C1, should be A B C D E");
+println(allpoints(C1));
+println("alllines incident to K, should be b");
+println(alllines(K));
+
+println("allsegments incident to L, should be c");
+println(allsegments(L));
+
+println("allmasses incident to Segment N-O, should be O");
+println(allmasses(no));
+println("allsprings incident to L, should be c");
+println(allsprings(L));
+println("allconics incident to A, should be C1");
+println(allconics(A));
+
+println("allcircles incident to G, should be C2");
+println(allcircles(G));
+
+println("allelements incident to K, should be b");
+println(allelements(K));
+
+
+//faulty case
+println("faulty case");
+A = [1,2,3];
+println(alllines(A));
+
+</script>
+
+    <script type="text/javascript">
+            var physics=[
+                       {behavior:{type:"Environment",gravity:-.2}},
+                       {name:"M", behavior:{type:"Mass",friction:0.1}},
+                       {name:"O", behavior:{type:"Mass",friction:0.1}},
+                       {name:"c", behavior:{type:"Spring"}}
+                       ];
+
+var cdy = createCindy({ 
+	scripts: "cs*", 
+	defaultAppearance: {}, 
+	behavior:physics,
+	geometry: [ 
+		{ name: "A", type: "Free", pos: [ 2.6486486486486487, -4.0, -0.6756756756756757 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "B", type: "Free", pos: [ 4.0, -3.183673469387755, -1.0204081632653061 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "C", type: "Free", pos: [ 2.8888888888888893, -4.0, -2.7777777777777777 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "D", type: "Free", pos: [ -2.795180722891566, -4.0, -1.2048192771084338 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "E", type: "Free", pos: [ 0.20382165605095542, -4.0, -0.6369426751592356 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "F", type: "Free", pos: [ 4.0, 0.8928571428571428, 0.4464285714285714 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "C0", type: "CircleByRadius", color: [ 0.0, 0.0, 1.0 ], radius: 3.807571404451924, args: [ "F" ], size: 1, printname: "$C_{0}$" }, 
+		{ name: "C1", type: "ConicBy5", color: [ 0.0, 0.0, 1.0 ], args: [ "A", "B", "C", "D", "E" ], size: 1, printname: "$C_{1}$" }, 
+		{ name: "G", type: "Free", pos: [ -1.6536312849162011, -4.0, -0.5586592178770949 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
+		{ name: "H", type: "Free", pos: [ -3.4254143646408846, -4.0, -0.5524861878453039 ], color: [ 1.0, 0.0, 0.0 ], labeled: true }, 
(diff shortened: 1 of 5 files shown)


strobelm:
Just pushed a rebased version on current master.

ref/Lists_of_Geometric_Elements.md
```diff
@@ -18,8 +18,10 @@ This is done with operators that return lists of elements of a particular kind.
 
 #### All springs of the construction: `allsprings()`
 
-#### All elements incident to a given one: `allsegments(‹geo›)`
+#### All elements incident to a given one: `all*(‹geo›)`
```
gagern: Can you please spell these out explicitely? We may want to use the reference manual to generate a list of all supported functions, to be used in e.g. an upcoming CindyJS editor. For that we need the list to be machine-parseable.

strobelm:
Hi, done!

gagern:
There are two aspects to this. The nullary functions `allcircles`, `allconics`, `allelements`, `alllines`, `allmasses`, `allpoints`, `allsegments` and `allsprings` are available in Cinderella, so we should have them for the sake of compatibility. No argument there.

The unary functions, on the other hand, are a new addition. Cinderella has `incidences` and the `is*` predicates, which can be combined in a `filter` to achieve this effect. So I'm not convinced here.

    allpoints(geo) := filter(incidences(geo), ispoint(#));

* Do we really want to provide these as separate functions, given that a simple combination of existing functions can achieve the same effect?
* Aren't names starting in `all` a bit confusing when you *don't* get *all* objects of a given line? Shouldn't we be using `incidentpoints` and so on instead?
* Should we really introduce an alias (`allelements` or `incidentelements`) for `indidences`? On the one hand it would be consistent with the systematic of the other functions, but on the other hand having two functions doing the same thing might be confusing to some users.

I'd like to see some input from @richter-gebert and/or @kortenkamp on these before merging.

strobelm:
It's a bit tedious to write documentation if the character delimiting the arguments is not on a standard keyboard. Isn't there a better solution than copying the arguments every time?

gagern:
@strobelm Your're on Linux, right? In that case I suggest you configure e.g. one of the Windows keys as [compose key](https://en.wikipedia.org/wiki/Compose_key) and edit your [`~/.XCompose` file](http://linux.die.net/man/3/xcompose) to provide you with suitable key bindings. Or go with the default; for a standard libX setup, [`/usr/share/X11/locale/en_US.UTF-8/Compose`](https://github.com/mirror/libX11/blob/7eb724dc24505f1591ef32620fa63f079b540646/nls/en_US.UTF-8/Compose.pre#L109-L110) should contain these settings:

    <Multi_key> <period> <greater>   	: "›"   U203a # SINGLE RIGHT-POINTING ANGLE QUOTATION MARK
    <Multi_key> <period> <less>      	: "‹"   U2039 # SINGLE LEFT-POINTING ANGLE QUOTATION MARK

kranich:
@gagern I generally agree with @strobel that these symbols are a pain. I 
would appreciate if we could come with a different alternative. Why 
can't we just use '<' and '>'? Embedded in backticks there should be no 
confusion with HTML tags for Markdown.

--
Stefan Kranich

gagern:
This is running off-topic, so I moved the ref manual formatting discussion to #211.

gagern:
First off, @richter-gebert approves of both having these functions and the names you chose. He suggested that we might use modifiers, to allow filtering by properties other than incidences. But I'd leave that for later.

I just noticed that indeed [Cinderella already documents an `allsegments(‹geo›)` function](http://doc.cinderella.de/tiki-index.php?page=Lists+of+Geometric+Elements) even though it doesn't seem to actually implement it. I had missed that fact when reading your diff. I filed [Cinderella bug 120](https://gitlab.cinderella.de:8082/cinderella/driftin/issues/120) for this discrepancy between documentation and implementation there, suggesting support for all types.

kranich:
@strobel It looks like you accidentally removed `ref/Lists_of_Geometric_Elements.md` in [a47fb2a4](https://github.com/strobelm/CindyJS/commit/a47fb2a45d6eef47cd60a88bdf53bd549e6270f1).

strobelm:
Thanks @kranich, i updated the branch.

Am 24.02.2016 um 10:03 schrieb Stefan Kranich:
>
> @strobel <https://github.com/strobel> It looks like you accidentally
> removed |ref/Lists_of_Geometric_Elements.md| in a47fb2a4
> <https://github.com/strobelm/CindyJS/commit/a47fb2a45d6eef47cd60a88bdf53bd549e6270f1>.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/209#issuecomment-188148399>.
>

-- 
Michael Strobel, M.Sc.

ref/Lists_of_Geometric_Elements.md
```diff
@@ -6,20 +6,35 @@ This is done with operators that return lists of elements of a particular kind.
 
 #### All points of the construction: `allpoints()`
 
+#### All points incident to an geometric object: `allpoints(‹geo›)`
```
kranich: @strobelm Sorry to bother you again. It should be 'a geometric object', not 'an', in all instances.

strobelm: Since when is "g" a consonant ;-) thanks again!

Am 24.02.2016 um 10:32 schrieb Stefan Kranich:
>
> In ref/Lists_of_Geometric_Elements.md
> <https://github.com/CindyJS/CindyJS/pull/209#discussion_r53913109>:
>
> > @@ -6,20 +6,35 @@ This is done with operators that return lists of elements of a particular kind.
> >  
> >  #### All points of the construction: `allpoints()`
> >  
> > +#### All points incident to an geometric object: `allpoints(‹geo›)`
>
> @strobelm <https://github.com/strobelm> Sorry to bother you again. It
> should be 'a geometric object', not 'an', in all instances.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/209/files#r53913109>.
>

-- 
Michael Strobel, M.Sc.

src/js/libcs/Operators.js
```diff
@@ -3592,14 +3592,98 @@ evaluator.halfplane$2 = function(args, modifs) {
 //   Geometric elements      //
 ///////////////////////////////
 
-evaluator.allpoints$0 = function(args, modifs) {
+
+// helper for all*(<geoobject>)
+eval_helper.allhelper1arg = function(args, kind, restrict) {
+    if (args.length > 1) return nada;
```
gagern: Can't happen: you are invoking this from a `all*$1` method so the compiler guarantees that there will be exactly one argument. Unless someone introduced a broken code-internal call, but in that case anything at all may happen.

gagern:
Will propose my own version building on this shortly. Please stand by…

src/js/libcs/Operators.js
```diff
@@ -3592,14 +3592,98 @@ evaluator.halfplane$2 = function(args, modifs) {
...
+    var geoEl = csgeo.csnames[args[0].name];
```
gagern: Here you are expecting the argument to be a variable, and take its name to access the geometric primitive. That's not the way to go. We need to evaluate the argument, and then deal with the result of that.

src/js/libcs/Operators.js
```diff
@@ -3592,14 +3592,98 @@ evaluator.halfplane$2 = function(args, modifs) {
...
+                });
+            }
+        } else { // no restrictions case
+            inci.forEach(function(el) {
+                cgeo = csgeo.csnames[el];
+                if (cgeo.kind === kind) {
+                    erg.push({
+                        ctype: "geo",
+                        value: cgeo,
+                        type: cgeo.kind
```
gagern: I know [`mover` sets the `type`](https://github.com/strobelm/CindyJS/blob/7882bf076a7baba4a810b62938b83610096fcdb1/src/js/libcs/Operators.js#L2877) (unconditionally to `P`), but [Namespace access to geo elements](https://github.com/strobelm/CindyJS/blob/7882bf076a7baba4a810b62938b83610096fcdb1/src/js/libcs/Namespace.js#L142-L145) only uses `ctype` and `value`, so that's all we should use anywhere, in my opinion.

src/js/libcs/Operators.js
```diff
@@ -3592,14 +3592,98 @@ evaluator.halfplane$2 = function(args, modifs) {
...
+    } // end if (geoEl)
+    return {
+        ctype: "list",
+        value: erg
+    };
+
+};
+
+// helper for all*()
+eval_helper.allhelper0arg = function(args, kind, array, restrict) {
```
gagern: Why pass `args` if it will be the empty array in any case?

src/js/libcs/Operators.js
```diff
@@ -3607,30 +3691,143 @@ evaluator.allpoints$0 = function(args, modifs) {
...
+    };
+
 };
 
 evaluator.alllines$0 = function(args, modifs) {
+    return eval_helper.allhelper0arg(args, "L", csgeo.lines);
+};
+
+evaluator.alllines$1 = function(args, modifs) {
+    return eval_helper.allhelper1arg(args, "L");
```
gagern: In most places, CindyJS treats segments as a special case of lines. So `alllines` should return segments, too, and there should be a test case to verify this.

examples/114_allops.html
```diff
@@ -0,0 +1,126 @@
...
+
+println("allpoints incident to b, should be K");
+println(allpoints(b));
+println("allpoints incident to C1, should be A B C D E");
+println(allpoints(C1));
+println("alllines incident to K, should be b");
+println(alllines(K));
+
+println("allsegments incident to L, should be c");
+println(allsegments(L));
```
gagern: `alllines()` and `alsegments()` correctly list `no` even though you don't include it in the expected result.

src/js/libcs/Operators.js
```diff
@@ -3607,30 +3691,143 @@ evaluator.allpoints$0 = function(args, modifs) {
...
-    for (var i = 0; i < csgeo.lines.length; i++) {
-        erg[i] = {
+
+    var points = evaluator.allpoints$0(null, null);
+    var masses = evaluator.allmasses$0(null, null);
+    var springs = evaluator.allsprings$0(null, null);
+
+    // no segments or circles since we consider them as lines 
+    var lines = evaluator.alllines$0(null, null);
+    var conics = evaluator.allsegments$0(null, null);
```
gagern: Why so complicated? `csgeo.gslp` has all the elements we want. Including texts and so on.

gagern:
Superseded by #294.

PR #210, opened 2016-02-04
Check for misuse of < and > to quote parameters in ref manual

gagern:
Motivation: https://github.com/CindyJS/CindyJS/pull/209#commitcomment-15883546


strobelm:
Works fine with my brackets, i think i'll write a script to replace them instead of introducing key  combinations i can't remember (EscapeMetaAltControlShift anyone?)

PR #212, opened 2016-02-04
Arrow heads empty and jet

### LINKED ISSUE #152: Additional arrow heads
Cinderella supports two more arrow shapes which CindyJS does not support at the moment. One is a hollow arrow, like `full` but only stroked not filled. The other is a delta-shaped head somewhere between `default` and `full`, and similar in appearance to arrow heads commonly used in TeX.

I propose the names `hollow` and `delta` for these new types. An implementation would have to extend [`draw_arrowhead`](https://github.com/CindyJS/CindyJS/blob/d39fb3b6d637581ef771e84ec27c8948c24cb4d0/src/js/libcs/Render2D.js#L397-L418), but also find the appropriate [base](https://github.com/CindyJS/CindyJS/blob/d39fb3b6d637581ef771e84ec27c8948c24cb4d0/src/js/libcs/Render2D.js#L365) [positions](https://github.com/CindyJS/CindyJS/blob/d39fb3b6d637581ef771e84ec27c8948c24cb4d0/src/js/libcs/Render2D.js#L374) for ending the line short of the arrow head.

gagern:
This fixes #152. Together with Cinderella merge request 65 it will ensure that arrow shapes exported from Cinderella can be matched by CindyJS. There is a *very* minor backwards incompatibility since the shape of existing arrow heads is slightly modified to match that used by Cinderella. I doubtr any content will have a problem with that.

### DIFF
--- a/examples/07_Feldlinien_newarrows.html
+++ b/examples/07_Feldlinien_newarrows.html
@@ -33,8 +33,8 @@
                           col=(c1,c2,0);
                           col=col/abs(col)*.7;
                           linesize(sqrt(s)*3);
-			  //draw(p-f*1.5,p+f*.5, arrow->true, arrowshape->"default", arrowsides->"==>", arrowsize->0.7, color->(0,0,1),color->col);
-			  draw(p-f*1.5,p+f*.5,arrow->true, arrowshape->"default" ,color->col);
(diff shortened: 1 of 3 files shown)


strobelm:
Looks fine to me, thanks @gagern!

PR #213, opened 2016-02-04
Implement the version function

gagern:
This implements a new nullary function called `version()`. At the moment, the version function will return a list

    ["CindyJS", 0, 0, 0, -1, "g‹commit-id›"]

which can be used to distinguish CindyJS from Cinderella, and to identify the commit in question.  The commit id will be terminated by an exclamation mark if the working directory is different from the named commit.

Once we start tagging releases, using annotated tags of the form `v‹major›.‹minor›.‹patch›`, then the second through fourth component will be the version components, and the fifth component the number of commits since that version (and therefore always 0 for official releases). The machinery for this is already in place in the build system.

To test this machinery, try `git tag -a v7.15.3 -m 'Just a test' master~12` or some such, but don't forget to do a matching `git tag -d` right after the test, in order to not accidentially push that tag to GitHub.

I'm not sure where to document this function. Since we want a matching function for Cinderella, I'll probably document it in the Cinderella handbook first, and let [docscrub](https://github.com/CindyJS/CindyJS/tree/docscrub) carry the changes from there to us.

### DIFF
--- a/make/README.md
+++ b/make/README.md
@@ -275,6 +275,17 @@ Here is a list of important commands.
   In that case, or if `files` is omitted altogether,
   `dst` should name a directory.
 
+### Version determination
+
+The file [`getversion.js`](getversion.js) provides a function
+which launces `git` to build a file called `Version.js`.
+However, unlike a regular command, this operation is not tied to a single task.
+Instead there are multiple tasks which need an up-to-date `Version.js`.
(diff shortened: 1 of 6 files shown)


strobelm:
As far as i can tell the code looks good. Thanks @gagern!

PR #214, opened 2016-02-05
Segments

gagern:
This is the result of my attempt to get the example from 79d611a8ee9390e1945b98d01752229eb696a828 working, with the possible exception of measured distance labels. There are some drive-by fixes and improvements, before 674e1b71d5bdcaac4a7aeb1d3c4ce7a59f0bf41d fixes a definite bug. Most of the issues have to do with segments, one way or another.

### DIFF
--- a/examples/114_ParallelSegments.html
+++ b/examples/114_ParallelSegments.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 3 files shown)


strobelm:
Good point, i wasn't aware of the segment problem. Nice to know! The fixes look fine.

PR #225, opened 2016-02-08
Text

gagern:
This is enough to get things like distance measurements going in the most common cases. Export support on the Cinderella side is in [!70](https://gitlab.cinderella.de:8082/cinderella/driftin/merge_requests/70).

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -2338,6 +2338,33 @@ geoOps.Dist.updatePosition = function(el) {
     el.value = List.abs(List.sub(List.normalizeZ(a), List.normalizeZ(b)));
 };
 
+geoOps.Angle = {};
+geoOps.Angle.kind = "V";
+geoOps.Angle.signature = ["L", "L", "P"];
+geoOps.Angle.initialize = function(el) {
+    if (el.angle === undefined)
+        el.angle = 0.5 * Math.PI;
+    putStateComplexNumber(CSNumber._helper.input(el.angle));
+};
+geoOps.Angle.updatePosition = function(el) {
+    var a = csgeo.csnames[el.args[0]].homog;
+    var b = csgeo.csnames[el.args[1]].homog;
+    var p = csgeo.csnames[el.args[2]].homog;
+    var ap = List.cross(a, List.linfty);
+    var bp = List.cross(b, List.linfty);
+    var cr = List.crossratio3(ap, bp, List.ii, List.jj, p);
+    var ang = CSNumber.mult(CSNumber.complex(0, 0.5), CSNumber.log(cr));
+    var prev = getStateComplexNumber();
+    var diff = (prev.value.real - ang.value.real) / Math.PI;
+    var winding = Math.round(diff);
+    if (!tracingInitial && Math.abs(winding - diff) > 1e-2)
+        requestRefinement();
+    ang = CSNumber.complex(winding * Math.PI + ang.value.real, ang.value.imag);
+    putStateComplexNumber(ang);
+    el.value = General.withUsage(ang, "Angle");
+};
+geoOps.Angle.stateSize = 2;
+
 geoOps.TextImpl = {};
 geoOps.TextImpl.kind = "Text";
 geoOps.TextImpl.signature = [];
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -10,6 +10,8 @@ geoOps._helper = {};
  * Tr - Projective transformation
  * Mt - Moebius transformation
  * Rc - Reflection in a circle
+ * V  - (numeric) value
+ * Text - Text
  */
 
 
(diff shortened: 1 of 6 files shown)


gagern:
Docking info from [!70](https://gitlab.cinderella.de:8082/cinderella/driftin/merge_requests/70) has been merged. Angle export is in [!72](https://gitlab.cinderella.de:8082/cinderella/driftin/merge_requests/72). I think this is ready for merging now.

elkins0:
@ gagern:Similar to #194, [32639c5](https://github.com/elkins0/CindyJS/commit/32639c5b3ad4b8775f016bc49766b731c84fb679) geoOps.Angle.kind = "V" instead of "A".

gagern:
@elkins0 Indeed, I had forgotten that we had Angle in there. More details there.

src/js/libgeo/GeoOps.js
```diff
@@ -2349,6 +2351,60 @@ geoOps._helper.conicOtherIntersection = function(conic, a, b) {
...
+    if (el.pos) el.pos = geoOps._helper.initializePoint(el);
+    if (el.dock) {
+        if (el.dock.offset && el.dock.offset.length === 2)
+            el.dock.offset = List.realVector([+el.dock.offset[0], +el.dock.offset[1]]);
+        else
+            el.dock.offset = List.realVector([0, 0]);
+    }
+};
+
+function noop() {}
```
strobelm: noop?

gagern: [No Operation](https://en.wikipedia.org/wiki/NOP). Would you prefer `nop`, or spelling it out as `noOperation`?

It's used for `geoOps.TextImpl.updatePosition` since text position isn't dealt with during tracing, only when rendering. But once we have a function of this name, we could use it in other locations as well \[[1](https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L377),[2](https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L461),[3](https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L715),[4](https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/Setup.js#L119),[5](https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/Head.js#L14)\].

strobelm: Oh i didn't see the call in updateposition, this makes perfectly sense.

Am 10.02.2016 um 13:24 schrieb Martin von Gagern:
>
> In src/js/libgeo/GeoOps.js
> <https://github.com/CindyJS/CindyJS/pull/225#discussion_r52450248>:
>
> > +geoOps.TextImpl.kind = "Text";
> > +geoOps.TextImpl.signature = [];
> > +geoOps.TextImpl.updatePosition = noop;
> > +geoOps.TextImpl.initialize = function(el) {
> > +    el.text = String(el.text);
> > +    el.size = CSNumber.real(el.size ? +el.size : defaultAppearance.textsize);
> > +    if (el.pos) el.pos = geoOps._helper.initializePoint(el);
> > +    if (el.dock) {
> > +        if (el.dock.offset && el.dock.offset.length === 2)
> > +            el.dock.offset = List.realVector([+el.dock.offset[0], +el.dock.offset[1]]);
> > +        else
> > +            el.dock.offset = List.realVector([0, 0]);
> > +    }
> > +};
> > +
> > +function noop() {}
>
> No Operation <https://en.wikipedia.org/wiki/NOP>. Would you prefer
> |nop|, or spelling it out as |noOperation|?
>
> It's used for |geoOps.TextImpl.updatePosition| since text position
> isn't dealt with during tracing, only when rendering. But once we have
> a function of this name, we could use it in other locations as well [1
> <https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L377>,2
> <https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L461>,3
> <https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/libgeo/Tracing.js#L715>,4
> <https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/Setup.js#L119>,5
> <https://github.com/CindyJS/CindyJS/blob/89219dec8d71a3df8fa1ee74b8e1821acf2a1ace/src/js/Head.js#L14>].
>
> —
> <https://github.com/CindyJS/CindyJS/pull/225/files#r52450248>.
>

-- 
Michael Strobel, M.Sc.

kranich: @gagern If you want to use function `noop` in many different places then maybe it should not live in GeoOps.js but in a more central file. Otherwise, why not simply assign `function () {}`?

gagern: @kranich I though that having a single function instead of several might save on resources like memory (since it's just a single function data structure) or computation time (due to easier caching and better branch prediction). But most probably the benefit is completely negligible, so if you consider `function() {}` clearer, I'd be happy to switch to that.

kranich: @gagern I do not strongly favour either of these choices, so just proceed as you see fit, please.

strobelm:
I just had a look on the example and observed some intersection between the labels and the docked text, is this intended?
text1

gagern:
> is this intended?

Yes. I probably should adjust the example to omit point labels. I just needed some points and texts docked to them, and was lazy about it.

Essentially what I did was draw some points in cinderella, look at the serialized representation of that (i.e. copy elements via clipboard to text editor), then try to match that with drawtext routines. That's why the example also has a draw script which I deactivated by renaming its id. I guess that should go as well.

PR #227, opened 2016-02-09
Arctan2

### LINKED ISSUE #226: Complex arctan2 implementation
The [current  `arctan2` implementation](https://github.com/CindyJS/CindyJS/blob/1e4083d09e6fbba3a22dde5ca9f2f2cd8bea941d/src/js/libcs/CSNumber.js#L343-L349) is only valid for real-valued input. We should replace that by a version which handles complex input correctly.

kranich:
This should fix #226.

### DIFF
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -251,7 +251,7 @@ The standard trigonometric functions are available through the following operato
 
 #### Inverse rigonometric tangent function: `arctan(‹expr1›)`
 
-#### Angle of a vector: `arctan2(‹real1,real2›)`
+#### Angle of a vector: `arctan2(‹expr1,expr2›)`
 
 #### Angle of a vector: `arctan2(‹vec›)`
 
@@ -266,6 +266,8 @@ However, the operator returns only one principal value, for which the real value
     < 45°
     > arctan2(-1,-1)
     < -135°
+    > arctan2(-i,i)
+    < 135°
 
 ------
 
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -255,6 +255,8 @@ The standard trigonometric functions are available through the following operato
 
 #### Angle of a vector: `arctan2(‹vec›)`
 
+#### Argument of a complex number: `arctan2(‹expr1›)`
+
 The `arc` operators are in principle multivalued.
 However, the operator returns only one principal value, for which the real value is between `+pi` and `-pi`.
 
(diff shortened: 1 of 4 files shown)


kranich:
If you do not like the idea of letting `arctan2` handle a single complex number as argument, please ask me to rebase, or cherry-pick the first commit.

gagern:
I think that adding support for `arctan2(‹complex›)` to the script function makes sense. I'm less certain about the internal library, mainly because I prefer to avoid case distinctions, but the cost of this one check is probably negligible compared to the rest of the function. So no veto there from me.

I am somewhat worried about the branch cuts, whether the real-valued computation has the same as the complex-valued one. Can you please add some unit tests which evaluate one real value pair from each quadrant, and then verify that adding a small complex error to that doesn't change the branch of the result?

If I understand you correctly, you deliberately want to propagate the effect of signed zeros in `Math.atan2` to the CindyJS implementation. If so, we *might* want to add four lines of test cases for that as well. Not certain it's important enough to warrant that, or even to support that in future implementations. What do you think, is this an official feature or just an implementation detail?

kranich:
>I think that adding support for arctan2(‹complex›) to the script function makes sense. I'm less certain about the internal library, mainly because I prefer to avoid case distinctions, but the cost of this one check is probably negligible compared to the rest of the function. So no veto there from me.

OK.

>I am somewhat worried about the branch cuts, whether the real-valued computation has the same as the complex-valued one. Can you please add some unit tests which evaluate one real value pair from each quadrant, and then verify that adding a small complex error to that doesn't change the branch of the result?

I think that won't work. I would rather test whether the results of `Math.atan2` coincide (approximately) with the results of our formula for real values in four different quadrants.
But this is only a hint that the branch cuts are similar, it does not say anything about the *exact* position of the branch cuts.

>If I understand you correctly, you deliberately want to propagate the effect of signed zeros in Math.atan2 to the CindyJS implementation.

Not necessarily.

>If so, we might want to add four lines of test cases for that as well. Not certain it's important enough to warrant that, or even to support that in future implementations. What do you think, is this an official feature or just an implementation detail?

If you ask me, its just an implementation detail. The case distinction for real arguments gives us automatic handling of `arctan2(0,0)`. That's why I added it.

It might make sense to read up on the issue of signed zeros in William Kahan's article 'Branch Cuts for Complex Elementary Functions, or Much Ado About Nothing's Sign Bit'. I found that quite illuminating when I read it some time ago.

kranich:
@gagern I added a compatibility test for the last two if-branches of `CSNumber.arctan2`. If you can come up with a better test, please let me know.

PR #231, opened 2016-02-09
Dim dependent points as in Cinderella

### LINKED ISSUE #217: Colour of dependent points
Dependent points are drawn in darker colours in Cinderella but not in CindyJS.

I am pretty sure this has been mentioned somewhere already (maybe on the mailing list) but I could not find a duplicate issue.

kranich:
This should fix #217.


gagern:
Please check with @richter-gebert whether the first hunk, the change of default, is really desired. It might well be that we should address this at the export from Cinderella instead. (We definitely should handle this property upon export, since the Cinderella user might have changed the setting. So it's just a question of whether we want to change the default in CindyJS *in addition to* handling it in export.

kranich:
@gagern I have just obtained @richter-gebert's OK, 'as long as it is globally or individually configurable', which it already is.

Issue #219, opened 2016-02-06
Add functions mentioned in documentation from Cinderella but not present in CindyJS


gagern:
The following functions (with arity) are mentioned in the documentation, are presumably supported by Cinderella, but at the time of this writing not ported to CindyJS:

* [ ] addforce$2
* [ ] adj$1
* [ ] algorithm$1
* [ ] allcircles$0
* [ ] allconics$0
* [ ] allelements$0
* [ ] allsegments$0
* [ ] allsegments$1
* [ ] allsprings$0
* [ ] amsdata$0
* [ ] assert$2
* [ ] attribute$2
* [ ] attribute$3
* [ ] autoclearlayer$2
* [ ] calibratedamsdata$0
* [ ] candisplay$1
* [ ] canvas$3
* [ ] check$0
* [ ] clear$0
* [ ] clear$1
* [ ] clearlayer$1
* [ ] closeconnection$1
* [ ] closefile$1
* [ ] colorplot$3
* [ ] convexhull3d$1
* [ ] createtool$3
* [ ] createtool$3
* [ ] crossratio$4
* [ ] d$2
* [ ] draw$1
* [ ] drawcurves$2
* [ ] drawfield$1
* [ ] drawfieldcomplex$1
* [ ] drawforces$0
* [ ] drawforces$1
* [ ] drawimage$5
* [ ] drawtable$2
* [ ] element$1
* [ ] elementsatmouse$0
* [ ] eval$4
* [ ] fillplot$1
* [ ] fillplot$2
* [ ] flush$1
* [ ] fontfamilies$0
* [ ] forall$2
* [ ] forall$3
* [ ] force$1
* [ ] geotype$1
* [ ] getmaxscore$0
* [ ] getscore$0
* [ ] greset$0
* [ ] grestore$0
* [ ] gsave$0
* [ ] guess$1
* [ ] guess$1
* [ ] hermiteanproduct$2
* [ ] imagergb$3
* [ ] import$1
* [ ] incidences$1
* [ ] inputs$1
* [ ] inspect$1
* [ ] inspect$2
* [ ] inspect$3
* [ ] instrument$1
* [ ] instrumentnames$0
* [ ] iskeydown$1
* [ ] isselected$1
* [ ] keydownlist$0
* [ ] keys$1
* [ ] length$1
* [ ] linearsolve$2
* [ ] load$1
* [ ] locusdata$1
* [ ] mapgrid$1
* [ ] mapimage$2
[...]

PR #233, opened 2016-02-09
Add take$2 operator

kranich:
Added `take$2` operator. Made all reftests for `take$2` and `infix_take` executable. This is one of the items on the #219 list.


kranich:
@gagern I think we should try to eliminate Code/Result tables from the reference documentation or else make them executable.

gagern:
> I think we should try to eliminate Code/Result tables from the reference documentation or else make them executable.

I'm very much in favor of the latter, and manually converted a number of them in 1275344974b3d9fab91adfbd8938403600a28405. I could try whether I can deal with this in some automated way.

PR #234, opened 2016-02-09
Reduce differences between CindyJS and Cinderella

gagern:
I've managed to export the CindyJS reftest suite to a JSON file, and hook that up to Cinderella so that it can run the same set of tests, ensuring compatibility for tested features or at least exposing incompatibilities so they can be discussed and dealt with. In the process, I found that

1. Cinderella prints angles with just a single decimal and
2. Cinderella uses `___`  to represent nada where CindyJS uses `_?_`.

The latter is likely an accident. I remember discussing compatibility with @richter-gebert at some point, and he was rather surprised that the string representations did *not* match. Both these points seem like unnecessary divergence, so we might unify behavior instead of catering for such differences in the test suite. I just rebased this onto #233 so it takes those cases of nada into account.

### DIFF
--- a/ref/Arithmetic_Operators.md
+++ b/ref/Arithmetic_Operators.md
@@ -306,10 +306,10 @@ for which the real part is between −90° = −π/2 and +90° = +π/2.
     > arcsin(-1)
     < -90°
     > arcsin(2) // the sign of the imaginary part is undefined here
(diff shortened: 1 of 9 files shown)


kranich:
@gagern Thank you.

PR #235, opened 2016-02-09
Improve tables and other things in ref docu

gagern:
Motivated by https://github.com/CindyJS/CindyJS/pull/233#issuecomment-181987883 I cleaned up several parts of the ref documentation, mostly related to tables.

There were and are only two remaining Code/Result tables, and since these describe drawing operations they can't be turned into executable code as easily as previous such tables.


kranich:
@gagern Thank you, another big improvement of the documentation.

PR #236, opened 2016-02-10
Mark functions not available in CindyJS

gagern:
This is based on a variation of the script from #219. I didn't process the sound parts yet.


strobelm:
Thanks @gagern that makes quite a nice task list. I think some functionality is already available internally but lacks a reference in operators.js.

Issue #229, opened 2016-02-09
default parameters for geoOps


strobelm:
While playing around with the angle bisector code it came to my mind that default arguments in geoOps would be nice to have. Therefor i quickly wrote some testing code:

https://github.com/strobelm/CindyJS/tree/defaultParam

The idea is to place some default arguments to the signature like this
```
geoOps.defaultArgChecker.signature = ["P", "P=[0,0,1]"];
```

and parse these in GeoBasics. Of course the code is only for discussion and not ready to be merged. Nevertheless any comments on the approach?

gagern:
I think that having a specific constant value as an optional argument is an unlikely case. The far more common case in my opinion would be an optional argument which when missing can be computed from the given arguments. I wouldn't want to do that in a text string encoding a CindyScript evaluation, but rather inside the implementation itself. (If that implementation were to be completely written in CindyScript, that would be a different story.) So I think we only need to tell the signature verification phase that it is OK to have `undefined` (and/or `null`?) elements in the `args` array. Then the implementation can deal with that in a way suitable to the situation at hand.

PR #237, opened 2016-02-10
Angle bisector take 3

gagern:
After #172 and #183 this is yet another candidate for angle bisectors. It is mostly a rebased version of #183, but with those changes that originated in #172 cherry-picked from there instead. The macro changes were not picked, but left for the very end to see whether aa5728897ce1c210d7d8e71cc51da0cadcc386b2 avoids the need for #229 and similar tricks.

Also omitted were the examples from Cinderella, since I haven't really looked at those good enough, and won't today because it's *really* late now. So I'm not requesting a review yet, since I'll have to add at least one Cinderella-generated example.

As a *huge* benefit, this pull request contains a new kind of unit tests, where you plug in a geometry description and can then verify the coordinates of the final object of that. There is a whole test suite in there verifying the angle bisector against the criteria from https://github.com/CindyJS/CindyJS/pull/183#issuecomment-175115285. This implementation satisfies all requirements.

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -137,7 +137,7 @@ module.exports = function build(settings, task) {
     // Run separate unit tests to test various interna
     //////////////////////////////////////////////////////////////////////
 
-    task("unittests", ["exposed"], function() {
+    task("unittests", ["exposed", "plain"], function() {
         this.cmdscript("mocha", "tests");
     });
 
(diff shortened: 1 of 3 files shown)


strobelm:
Hi, i checked the code and i really like it. The bisector code is, a big thanks to @elkins0, very neat and clean. @gagern thanks for tidying up and for the geometric tests.

So all good things are worth waiting for (and this one merged).

strobelm:
I just forgot to mention: the Cinderella export works fine (i didn't add a test file, though).

PR #239, opened 2016-02-11
fixed circle shape smoothness

### LINKED ISSUE #238: Rendering of circles is too coarse
I just tried to export a file from Cinderella – it worked fine, but the rendering of circles is below our standards... 

bildschirmfoto 2016-02-11 um 13 51 25

This is how it looks in Cinderella itself:
bildschirmfoto 2016-02-11 um 13 48 47

Browser used is Safari, latest version.

strobelm:
The number of segments approximating circles in eval_helper.shapeconvert was too low such that the human eye could see the polygon shape. This increases the number of segments to a reasonable (smoothness / performance) number and should resolve  #238.


src/js/libcs/Operators.js
```diff
@@ -2775,8 +2775,8 @@ eval_helper.shapeconvert = function(a) {
         var mx = aa.value[0].value.real;
         var my = aa.value[1].value.real;
         var r = a.value.value[1].value.real;
-        li = [];
-        var n = 36;
+        var n = 125;
+        li = new Array(n);
```
kranich: Have you tested the performance of `li = new Array(n)` vs. `li = []`?

kranich: OK, `new Array(n)` seems to beat `[]` at least in Chrome and Firefox, see http://jsperf.com/cindyjs-new-array-n-vs/3

kranich:
This should fix #238.

PR #248, opened 2016-02-15
Parse scripts more like Cinderella does, document language

gagern:
So far, Cinderella and CindyJS were parsing their input differently, particularly when it came to what are valid variable names or numeric literals. This change attempts to unify things, taking Cinderella as reference and adjusting CindyJS to match. It also documents the current state of affairs, in a way which is still far from formal, but nevertheless way more precise than what we had before.

If any deviation from this is intended, now or in the future, that intent should be made explicit and be addressed in both systems and the documentation.


kranich:
Could you please fix the build failure,
`libcs/Namespace.js: line 34, col 24, Expected '!==' and instead saw '!='.`

Otherwise: nice work!

gagern:
@kortenkamp Please have a look at the `Language.md` file here, see whether I have missed an important aspect of the language, whether you would like to add to that. No need to block the merge, I think, since we can always improve documentation.

kranich:
@gagern I won't be able to handle this before Friday. I am currently away and Internet connection is lousy. If you need this merged sooner, please assign @strobelm or merge it yourself.

ref/Boolean_Operators.md
```diff
@@ -181,7 +181,14 @@ This operator is similar to **&gt;** but tests for **less than or equal to**.
 
 ------
 
-#####  Fuzzy comparisons: `~=`, `~!=`, `~<`, `~>`, `~>=`, `~<=`
+####  Fuzzy comparisons:
+
+##### Approximately equal: `‹expr1› ~= ‹expr2›`
+##### Not approximately equal: `‹expr1› ~!= ‹expr2›`
```
kranich: This should be 'Approximately not equal' as this corresponds to the order of the symbols in the operator and would be consistent with Cinderella documentation.

src/js/libcs/Essentials.js
```diff
@@ -3,82 +3,98 @@
...
+operators['~~'] = 370; // list or shape intersection
+operators[':>'] = 370; // append element to list
+operators['<:'] = 370; // prepend element to list
+operators['='] = 400; // assignment
+operators[':='] = 400; // function definition
+operators[':=_'] = 400; // function undefinition
+operators['::='] = 400; // binding function definition
+operators['->'] = 400; // modifier
+operators[';'] = 500; // sequence of statements
+//operators[','] = 510; // list element separator, handled in parselist
```
kranich: At first, I wondered if this was deliberately commented out.

I would turn this into a mere comment, e.g.
```javascript
// The list element separator ',' (operator precedence 510) is handled in parselist.
```

kranich: Or
```javascript
// operators[','] = 510, the list element separator, is handled in parselist
```

tools/unicodeCat.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const request = require('request');
+
+create(
+    'bmpLetters',
+    'https://raw.githubusercontent.com/mathiasbynens/unicode-6.2.0/' +
```
kranich: Not sure I like the external reference here. Can't we include this in our codebase in the `tools` directory?

kranich:
@gagern I cannot comment on `Language.md`. Otherwise this should be fine to merge, up to the minor issues I pointed out above. Thank you!

ref/Boolean_Operators.md
gagern: That was a deliberate modification. To me, “approximately equal” is a common term which has a well-defined meaning: equal up to some fuzz. “Approximately not equal”, on the other hand, is a rather unusual term, so one has to start thinking. Is it just *approximately* unequal, but not *really exactly* unequal? How so? Looks unequal at first glance but turns out to be equal on closer inspection? Yes, you could rule out such interpretations pretty quickly, but the fact that you could start thinking along these lines indicates potential confusion. On the other hand, negating an expression is common enough, so negating “approximately equal” we obtain “not approximately equal”.

@kranich: The same for the other approximate comparisons: I rephrased them to be combinations of an exact comparison and approximate equality. For the strict fuzzy comparisons (strange combination of words!), I realize that the placement of the negation allows for a second alternative: instead of my “not greater or approximately equal” one could write “less and not approximately equal”, if you think that's an improvement.

Your statement about the order of symbols is of course quite valid; I hadn't considered that aspect. I find it difficult to judge between these goals. If you uphold your request even after hearing about my motivation, and noone else agrees with me soon enough, I'll concede to that.

By the way, what documentation are you referring to? The printed book? I thought that was generated from the web site, as were our documentation files. And I couldn't find any occurrence of “approximately” in our files. So where does the documentation have a spelled out name for these operators?

tools/unicodeCat.js
gagern: We could make the `unicode-6.2.0` package a dependency of ours, but that package is pretty big, and we only need a tiny portion of that, plus we only need that tiny portion if we decide to twiddle with the regexp generated from it. We could of course download the file and add it to our codebase, but then we'd have to include licensing and attribution details as well. And the file is still over 600kB. The most likely reason one would want to run my script is in order to switch to e.g. Unicode 8, in which case changing a URL is probably easier than replacing a file in our repo, particularly with respect to the resulting diff.

In general I'd say the repository is for things unique to our project, everything external should be fetched externally. That's why we removed the Closure Compiler from our repository when we refactored it before going public. That's why we don't include GWT or Node or any of the things we need to compile.

I just read [the unicode-data readme](https://github.com/mathiasbynens/unicode-data#readme), from where this data presumably originated. It mentions a [HTTP API](https://github.com/mathiasbynens/unicode-data#http-api) which we might use instead of this GitHub download. I could also follow https://github.com/mathiasbynens/node-unicode-data/issues/14 and create a node package just for the `L` category codepoint list. I don't feel like maintaining all categories for all Unicode versions, though. Or we could depend on `node-unicode-data` and generate the data file ourselves.

Since we should always be able to get the list of codepoints for a given category from some other source if one source fails us, I see no urgent need to add that file to the repo. Have I convinced you? Do you prefer any of the other alternatives I mentioned?

src/js/libcs/Essentials.js
gagern: Agreed, will change this.

gagern:
Assignig to @kortenkamp for `Language.md`, but I'll still resolve open questions with @kranich while waiting on that.

ref/Boolean_Operators.md
kranich: >So where does the documentation have a spelled out name for these operators?

http://doc.cinderella.de/tiki-index.php?page=Operator+Index under 'Infix operators'

kranich: And I would really stick with 'approximately not equal', etc. The picture from p. 254 of the printed book might be helpful as well.

tools/unicodeCat.js
kranich: Convinced, leave it as is. I erroneously assumed the tool would be run by the build process. Since it is only some sort of executable documentation for the unicode part of the parser, this is fine.

ref/Boolean_Operators.md
gagern: > http://doc.cinderella.de/tiki-index.php?page=Operator+Index under 'Infix operators'

Ah, thanks. I didn't fetch that since I intend to auto-generate the index eventually.

> And I would really stick with 'approximately not equal', etc.

I had a look at Unicode. It calls ≉ [“not almost equal to”](http://www.charbase.com/2249-unicode-not-almost-equal-to), ⪅ [“less-than or approximate”](http://www.charbase.com/2a85-unicode-less-than-or-approximate) and ⪉ [“less-than and not approximate”](http://www.charbase.com/2a89-unicode-less-than-and-not-approximate). I think these terms are closer to my formulation than the original one. I guess someone invested some effort into these names, but I'm not sure. I also [asked about this on Math SE](http://math.stackexchange.com/q/1662241/35416) and would like to wait a bit to see what answers come in there.

> The picture from p. 254 of the printed book might be helpful as well.

I don't have the book at hand just now, but I imagine you're referring to [this figure](https://github.com/CindyJS/CindyJS/blob/6662cc2e0a902d3be2ea590a3ef23545a32d494c/ref/img/Comparisons.png) which is already included in that document. It's indeed massively useful, and up until recently I had to refer to that in order to get the distinction between `~<` and `~<=` right. Now with my new names and the association to ⪉ and ⪅ set in my mind, I think I finally will remember them. Our discussion here surely helps in that respect, too.

tools/unicodeCat.js
gagern: Right, the build process should not need external resources at every run. Come to think of it, I should probably design some way of offline mode, where it won't run things like `npm install`, so that the build can be used when not hooked up to the internet at all even if the `package.json` changes due to git branch hopping. But that's another story.

ref/Boolean_Operators.md
kranich: Maybe @kortenkamp or @richter-gebert can give their opinion on the naming, too?

kortenkamp: For me, "a not approximately equal b" is `not(a ~= b)`, and "a approximately not equal b" differs from that a bit, that is at +-epsilon. But from the diagram it is not visible what happens there, and it must not be important. So, I prefer "approximately not equal" as it matches the symbol order `~`, `!`, `= ` in `~!=`.

gagern:
I changed the descriptions of the fuzzy comparison operators as @kranich requested, but I'm still not happy with that. I discussed them with some colleagues during lunch today. While “approximately not equal” apparently was easy enough to understand, the approximate inequalities led to different interpretations, indicating a considerable risk of mistakes and confusion.

One suggestion was to use a different adjective for the whole “not approximately equal” family of operators. One suggestion would be “noticeably”, so you'd have “noticeably different” (or “noticeably not equal”?), “noticeably greater” and “noticeably less”. One might also consider “significantly” as an alternative, but that term has some very well-defined meanings to statisticians which don't agree with our application.

There was some agreement that the phrase “approximately equal” should be seen as a unit, so that “less or approximately equal” would be acceptable, as would “approximately equal or less”, while “approximately less or equal” felt wrong and unclear. Of course, this would mean diverging from the symbol order.

What do you think of these?

@kortenkamp wrote:
> For me, "a not approximately equal b" is `not(a ~= b)`, and "a approximately not equal b" differs from that a bit, that is at +-epsilon. But from the diagram it is not visible what happens there, and it must not be important.

Firstly, I agree that the behavior at the ±ε boundaries must not matter for a single comparison, since these operators are meant to account for numeric imprecision. What *might* be important is whether these cases partition the set of all possible cases. In other words, I think it should hold that exactly one of `a ~> b` and `a ~<= b` should hold, exactly one of `a ~= b` and `a ~!= b`, and exactly one of `a ~= b`, `a ~< b` and `a ~> b`. Some users might want to consider these cases independently, and might assume that such a case distinction is exhaustive.

The above is barring any NaN values. To be consistent with other operators, NaN should compare false to any number, no matter the operation. I guess I should write some unit tests to ascertain that behavior.

In this light, the NaN behavior would be the only real difference between `not(a ~= b)` and `a ~!= b`: the former would be true if either number is NaN, the latter false. Personally I'd pronounce the former as “not *a* approximately equal to *b*” and the latter as “*a* not approximately equal to *b*”, but perhaps that point is moot if you agree to “*a* noticeably different from *b*”.

(One way to achieve consistent partitioning for all cases is by deciding whether ε itself is approximately equal to zero. The documentation claims ε = 10<sup>−10</sup>, which is indeed the threshold used in Cinderella. The floating point representation of this is 7737125245533627∙2<sup>−86</sup> which is slightly more than 10<sup>−10</sup>. So if we take the spec *very* literally here, we should consider ε ≉ 0. Looking at this made me notice #276.)

ref/Language.md
```diff
@@ -0,0 +1,361 @@
...
+These always denote a list, even if they contain exactly one expression.
+
+    > 7 * [1 + 2]
+    < [21]
+    > 7 * [1, 2]
+    < [7, 14]
+    > 7 * []
+    < []
+
+#### Curly braces `{…}`
```
kortenkamp: Shouldn't they be deprecated currently?

gagern:
@kortenkamp said he's happy with `Language.md`. I've now renamed the headlines for the fuzzy comparison to something which should be clearly understandable by a very wide audience, and added some doctests for these as well, noticing *slight* deviations from specified behavior in the process. I think this should be ready to land now, so that we can build on it with the new parser, and with a set of functions to control the ε variable for fuzzy comparisons.

kranich:
@gagern No objections from me. Feel free to merge this yourself.

PR #257, opened 2016-02-16
Reimplemented parser

gagern:
This is a change which will have a huge impact. So I don't expect this to be merged right now, but instead to be **discussed** and brought into a shape where it can be merged without causing pain for anyone. Essentially, this is a **rewrite of the parser**.

The old parser operated on strings, and used to repeatedly iterate over the same strings, causing something like an O(*n*<sup>2</sup>) time complexity. The new parser turns the input text into a token stream, and then builds expressions from these tokens in what looks like linear time to me. There is a **factor 5 performance boost**, reducing the time needed to parse all the scripts from all the examples from 2.5s to 0.5s, which is quite considerable in my opinion.

The parser is also more **modular**. The single file `Parser.js` can be used by other projects without depending on the rest of CindyJS. This is useful for applications like using CIndyScript at compile time, or building a CindyScript editor with syntax checking and highlighting. The CindyJS internals `infixmap` and `namespace` are still coupled via a shared scope, but there might be a more elegant way to deal with these in the long run.

The modularity can also be put to use in **unit tests**, comparing text input against expected parse results. The parser comes with a bunch of such unit tests, and may wel be extended to explicitely cover tricky cases of the language specification in the future.

While the parser keeps the CindyScript tradition of **ignoring whitespace**, it does so during tokenization, **without actually removing whitespace** from the input. That allows much greater flexibility in translating back and forth between parsed syntax and input text. This feature is essential for syntax highlighting. The parsed tokens already come with row and column source code position information. With moderate amounts of work it should be possible to extend that information to all nodes of the abstract syntax tree.

For convenience, the new parser was written in EcmaScript 2015 a.k.a. **ES6**. The build system has Closure compile that to ES5 so it will continue to run on most browsers. But internal tools like unit tests now may and will use ES6 syntax, as suppoered by node 4 and later. So we bump the node requirement from 0.12 to 4.0. **If desired, I could rewrite the parser in ES5.** There is nothing about it that couldn't be expressed in ES5 as well, with probably a few more lines of code.

The language parsed by the parser is ***not* the same language** Cinderella or the previous CindyJS parser recognizes.

* Operator precedence has been modified with the aim of more consistent behavior.
Compare the new `operatorLevels` data structure with the old `operators` dictionary.
* Some operations like `^` and `=` are right-associative.
Look for the `rassoc` key in `operatorLevels`.
* The dot `.` may no longer be used to denote a zero.
* The dot now becomes part of numeric literal tokens, which means `7.x` reports a lacking operator between `7.` and `x`, where Cinderella treats this as the field `x` of the object `7`, which apparently is `7`.
* Multiple dots are parsed as Cinderella does: `1..3` is the range, not the literal `1.` followed by the literal `.3`. `1...3` would be `1 .. .3`.
* Numeric literals may now be written in scientific notation, e.g. `1e-3`.
* Many operators now come with unicode variants of the symbol, for better readability in some scenarios.
* The parser is more strict; it will issue a compile-time error where other implementations used to just return nada at runtime. This is particularly true for fixity checks in `applyOperator`.
* `|…|` may now be nested, as long as it is enclosed in a bracket of a different type. This is thanks to the recursive parsing, which allows `|x, f(|y|)|` to work as expected.
* A name may not start with a numeric literal. While Cinderella allows `4by4` to be used as a variable name, this new parser will tokenize this as the numeric literal `4` followed by the identifier `by4`.
* Names of fields and modifiers have to be identifiers.
* Modifiers may not be enclosed in parentheses.
* Like Cinderella but unlike current CindyJS, function names now follow the same rules as variable names, which in particular means that the builtin function `#drawgrid` has to be renamed.

This pull request is **closely related to #248**. I guess this one here is much more controversial than that one, so I expect that one to be merged before this here is ready, and will then merge it into my branch here as well.

I haven't cleaned out the old parser yet. Once we reach consensus on the new parser, I intend to **remove the old parser** before this whole thing gets merged.

**Please discuss** the changes pointed out above, voice any objections you have to them, and any additional features you'd like to see included. Please also have a look at the code and let me know if I should do anything to make it easier to understand and maintain.

If we reach consensus here, I'll try to write a **similar parser for Cinderella**, to keep the two implementations compatible. Most likely Cinderella will then have two independent parsers, one for backwards compatibility and one for newer code, with the option to switch between them at will.


gagern:
@richter-gebert expressed his general consent with this whole approach (via email). Will do some tweaks and some cleaning up, then I'll request a review with the specific state of the implementation we have by then.

kortenkamp:
And I also expressed my appreciation (but only orally, so I just state it here).

I think it is a good idea to test a lot of examples with the new parser, maybe we can just quietly do this upgrade, even if it introduces incompatibilities.

gagern:
@strobelm I think this is ready for code review now. Please point out all code portions which are not readily understandable, so that I can add comments to make things clearer.

strobelm:
Hi, building on the branch currently fails:

```
required modules installed, resuming command
[Cindy.js    ] Forcing run of Cindy.js since setting 'build' changed value
[cindy3d     ] Forcing run of cindy3d since setting 'c3d_closure_warnings' changed value
[cindygl     ] Forcing run of cindygl since setting 'cgl_closure_warnings' changed value
[quickhull3d ] Forcing run of quickhull3d since setting 'gwt_args' changed value
TypeError: undefined is not a function
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:75:36)
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:64:30)
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:57:29)
    at Visitor.statement (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:41:20)
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:56:22)
    at Visitor.statement (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:41:20)
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:56:22)
    at Visitor.statement (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:41:20)
    at Visitor.visit (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:56:22)
    at Visitor.statement (/home/michael/Dokumente/promotion/cindyjs/tools/cs2js.js:41:20)
From previous event:
    at /home/michael/Dokumente/promotion/cindyjs/make/build.js:49:18
From previous event:
    at Array.reduce (native)
    at Task.run (/home/michael/Dokumente/promotion/cindyjs/make/Task.js:237:22)
From previous event:
    at /home/michael/Dokumente/promotion/cindyjs/make/Task.js:205:18
Makefile:2: recipe for target 'all' failed
make: *** [all] Error 2
```

furthermore we have merge conflicts in the current state.

gagern:
@strobelm: Rebased, and resolved errors.

kranich:
@gagern b99e652 introduced a build error:
```
[nodetest    ] node ref/js/runtests.js
[nodetest    ] Location:  /home/travis/build/CindyJS/CindyJS/ref/Arithmetic_Operators.md:112
[nodetest    ] Input:     > [[2,4],[4,8]] / (-2*i)
[nodetest    ] Expected:  < [[1*i, 2*i], [2*i, 4*i]]
[nodetest    ] Actual:    < [[0 + i*1, 0 + i*2], [0 + i*2, 0 + i*4]]
[nodetest    ]
[nodetest    ] 1 of 510 tests failed
```

gagern:
@strobelm I think this is ready for review and merging. So when you find the time…

strobelm:
On 03/23/2016 08:28 PM, Martin von Gagern wrote:
> @strobelm I think this is ready for review and merging. So when you find the time…
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/CindyJS/CindyJS/pull/257#issuecomment-200507270

It's on top of my stack and i'm looking forward to review the code.

-- 
Michael Strobel, M.Sc.

make/build.js
```diff
@@ -39,17 +40,29 @@ module.exports = function build(settings, task) {
...
+        });
+    });
+
+    task("plain", ["cs2js"], function() {
         version(this);
         this.concat(src.srcs, "build/js/Cindy.plain.js");
     });
 
-    task("ours", [], function() {
+    task("ours", ["cs2js"], function() {
```
strobelm: Perhaps we could make 'plain' the default behavior (of course not merge critical here)

make/sources.js
```diff
@@ -1,16 +1,18 @@
...
 exports.libcs = [
-    "src/js/libcs/Namespace.js",
-    "src/js/libcs/Accessors.js",
     "src/js/libcs/CSNumber.js",
     "src/js/libcs/List.js",
-    "src/js/libcs/Essentials.js",
     "src/js/libcs/General.js",
+    "src/js/libcs/Essentials.js",
+    "src/js/libcs/Namespace.js",
+    "build/js/Compiled.js",
```
strobelm: I'm wondering why something from build is included here.

make/sources.js
```diff
@@ -36,6 +38,10 @@ exports.lib = [
     "lib/clipper/clipper.js",
 ];
 
+exports.cssrc = [
+    "src/cs/drawgrid.cs",
```
strobelm: If i got that right, i can write my algorithms in cindyscript now? amazing!

package.json
```diff
@@ -7,6 +7,7 @@
     "test": "tests"
   },
   "devDependencies": {
+    "chai": "^3.5.0",
```
strobelm: By the way: it look like were not affected by the pullback of node packaged which occurred today:
http://www.heise.de/newsticker/meldung/JavaScript-Paket-aus-NPM-entfernt-Node-Babel-und-Co-scheiterten-beim-Build-3148796.html

src/js/libcs/Parser.js
```diff
@@ -1,674 +1,693 @@
...
+var flexfix = [';', ','];
+
+var operatorSymbols = [];
+var operators = {};
+
+(function initializeOperators() {
+    var precedence = 0;
+    operatorLevels.forEach(function(level) {
+        precedence += 2;
+        var rassoc = !!level.rassoc;
```
strobelm: double exclamation mark?

tools/unicodeCat.js
```diff
@@ -3,10 +3,9 @@
 const request = require('request');
```
strobelm: sorry, i have no idea what's going on in this file.

strobelm:
Some examples of CindyGL are not working with the new parser, for example cindygl/11_tilings.html. 

The rest of the examples (including Cindy3D) are looking fine (we got quite a stockpile already!). Frankly speaking: this is a beautiful piece of code, thank you very much @gagern!
As soon as CindyGL is working i'll merge the code.

make/build.js
gagern: Already did that in 7fdc737671c772bfa39ce65e88d1407ab6b4dc0c as discussed in https://github.com/CindyJS/CindyJS/pull/154#issuecomment-173406936.

make/sources.js
gagern: Most control flow operations aren't properly supported yet, so chances are you'll have to tweak the transpiler a bit, but in theory yes, it should be fairly easy.

src/js/libcs/Parser.js
gagern: Cast to bool. Turns `undefined` into `false`.

make/sources.js
gagern: That's JavaScript code transpiled from CindyScript. Perhaps I should choose a different name here. Any suggestions?

tools/unicodeCat.js
gagern: In eb648f4b06dc2ce84e028ba139878439e2933322 I added a comment to explain the code here, both purpose and algorithm.

gagern:
I have some trouble pinpointing the CindyGL issue. Looks like the type inference has some problems, but since I don't know how it's supposed to work, I can't see where it goes wrong. @montaga, do you want to have a look at this? Maybe it's due to 34dbb853c75263c5d2438270705bc63af855f521 or fb2d552b0c031c45e87f281f2311dfe606ec8341 somehow.

gagern:
@montaga, I believe I found the problem: I was representing a field access using the `.` operator as a binary operation with a variable as the second argument, and your code apparently iterates over the expression tree no matter the operation, collecting variables from the arguments. Fixed in 1ef3062ea1bc6e88e639c1d4a511cfc30f9083fc at least for `.`. I did the same for `->` but these don't occur in the AST so I think we should be on the safe side.

strobelm:
On 03/25/2016 06:01 PM, Martin von Gagern wrote:
> @montaga, I believe I found the problem: I was representing a field access using the `.` operator as a binary operation with a variable as the second argument, and your code apparently iterates over the expression tree no matter the operation, collecting variables from the arguments. Fixed in 1ef3062ea1bc6e88e639c1d4a511cfc30f9083fc at least for `.`. I did the same for `->` but these don't occur in the AST so I think we should be on the safe side.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/CindyJS/CindyJS/pull/257#issuecomment-201364103

Great! Any show-stopper left?

-- 
Michael Strobel, M.Sc.

Issue #258, opened 2016-02-16
Incorrect results in Lie circle geometry computation


gagern:
In [this example](http://martin.von-gagern.net/se/MX1657718) I've used Lie circle geometry to sole an instance of Apollonius' problem. In Cinderella this works nicely: the cyan circles touch the blue circle and two of the lines. In CindyJS, however, the result looks just wrong.

I haven't found time yet to investigate this problem and see where the implementations diverge. So right now this ticket here is mostly a reminder to do some further investigation. If I find the actual cause, I might edit it with a more detailed description of what's actually going on.

strobelm:
On 02/17/2016 12:07 AM, Martin von Gagern wrote:
>
> In this example <http://martin.von-gagern.net/se/MX1657718> I've used
> Lie circle geometry to sole an instance of Apollonius' problem. In
> Cinderella this works nicely: the cyan circles touch the blue circle
> and two of the lines. In CindyJS, however, the result looks just wrong.
>
> I haven't found time yet to investigate this problem and see where the
> implementations diverge. So right now this ticket here is mostly a
> reminder to do some further investigation. If I find the actual cause,
> I might edit it with a more detailed description of what's actually
> going on.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CindyJS/CindyJS/issues/258>.
>

You are refereeing to the jumping circles, aren't you? Additionally
there are drawing bugs caused by canvas.arc, i think we want to use the
conic drawing routine (although it's expensive) since it's doing the
drawing at least not as bad as canvas.arc. The Bezier curves which
@richter proposed cause problems rendering infinite radii (at least what
i can tell here).

-- 
Michael Strobel, M.Sc.

Technische Universität München -- Zentrum Mathematik
Lehrstuhl für Geometrie und Visualisierung (M10)
Boltzmannstr. 3, 85747 Garching bei München 
https://www-m10.ma.tum.de/bin/view/Lehrstuhl/MichaelStrobel

gagern:
@kranich: I'm referring mainly to the fact that the circles are not tangent to the other things, but the jumping probably is a consequence thereof. I'll have the script print out the raw circle drawing data, in case it's just rendering bugs, but I doubt it.

PR #259, opened 2016-02-17
draw circles with conic routine

strobelm:
Hi, the canvas.arc routine fails to draw circles with large radius. 

wrong:
arc1
correct:
arc2
This code will switch the drawing to the conic routine. I haven't removed the old code since it's more efficient and can be reused as soon the drawing bugs are fixed in chromium. 

This will NOT resolve the bug in #258.

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -281,53 +281,49 @@ eval_helper.drawcircle = function(args, modifs, df) {
     var v0 = evaluateAndVal(args[0]);
     var v1 = evaluateAndVal(args[1]);
 
-    function magic_circle(ctx, x, y, r) {
-        m = 0.551784;
-
-        ctx.save();
-        ctx.translate(x, y);
-        ctx.scale(r, r);
-
-        ctx.beginPath();
-        ctx.moveTo(1, 0);
-        ctx.bezierCurveTo(1, -m, m, -1, 0, -1);
-        ctx.bezierCurveTo(-m, -1, -1, -m, -1, 0);
-        ctx.bezierCurveTo(-1, m, -m, 1, 0, 1);
-        ctx.bezierCurveTo(m, 1, 1, m, 1, 0);
-        ctx.closePath();
-        ctx.restore();
-    }
-
-
     var pt = eval_helper.extractPoint(v0);
 
 
-    if (!pt.ok || v1.ctype !== 'number') {
+    if (!pt.ok || v1.ctype !== 'number' || !CSNumber._helper.isAlmostReal(v1)) {
         return nada;
     }
-    var m = csport.drawingstate.matrix;
 
-    var xx = pt.x * m.a - pt.y * m.b + m.tx;
-    var yy = pt.x * m.c - pt.y * m.d - m.ty;
+    /*
+       var m = csport.drawingstate.matrix;
 
-    Render2D.handleModifs(modifs, Render2D.conicModifs);
-    Render2D.preDrawCurve();
+       var xx = pt.x * m.a - pt.y * m.b + m.tx;
+       var yy = pt.x * m.c - pt.y * m.d - m.ty;
+       Render2D.handleModifs(modifs, Render2D.conicModifs);
+       Render2D.preDrawCurve();
 
-    csctx.beginPath();
-    csctx.arc(xx, yy, v1.value.real * m.sdet, 0, 2 * Math.PI);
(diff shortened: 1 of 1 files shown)


src/js/libcs/OpDrawing.js
```diff
@@ -281,53 +281,49 @@ eval_helper.drawcircle = function(args, modifs, df) {
...
-        ctx.closePath();
-        ctx.restore();
-    }
-
-
     var pt = eval_helper.extractPoint(v0);
 
 
-    if (!pt.ok || v1.ctype !== 'number') {
+    if (!pt.ok || v1.ctype !== 'number' || !CSNumber._helper.isAlmostReal(v1) || !List._helper.isAlmostReal(v0)) {
```
gagern: I think the `List._helper.isAlmostReal(v0)` check is incorrect. Consider `drawcircle(i*[3,2,1], 4)` where the point is a complex representative of a real point. `extractPoint` should be able to deal with this, and report its success in the `pt.ok` flag. Checking whether the homogeneous coordinate vector is real as well causes unneccessary restrictions.

strobelm: On 02/17/2016 10:41 AM, Martin von Gagern wrote:
>
> In src/js/libcs/OpDrawing.js
> <https://github.com/CindyJS/CindyJS/pull/259#discussion_r53140238>:
>
> >      var pt = eval_helper.extractPoint(v0);
> >  
> >  
> > -    if (!pt.ok || v1.ctype !== 'number') {
> > +    if (!pt.ok || v1.ctype !== 'number' || !CSNumber._helper.isAlmostReal(v1) || !List._helper.isAlmostReal(v0)) {
>
> I think the |List._helper.isAlmostReal(v0)| check is incorrect.
> Consider |drawcircle(i*[3,2,1], 4)| where the point is a complex
> representative of a real point. |extractPoint| should be able to deal
> with this, and report its success in the |pt.ok| flag. Checking
> whether the homogeneous coordinate vector is real as well causes
> unneccessary restrictions.
>
> —
> <https://github.com/CindyJS/CindyJS/pull/259/files#r53140238>.
>

Oh yes you're right!

-- 
Michael Strobel, M.Sc.

strobelm:
So here is the update. The routine ignores the complex number issue now (since extractPoint throws away the complex part). Perhaps we should discuss this sometime.

gagern:
Merged in 634f6120ecccee6c16c433f4ba24165bf71cc42e. I changed the indentation so that #260 still applies without modifications, as will other changes to that piece of code if we ever need them.

gagern:
I made a mistake in merging this: while `drawconic` is a suitable alternative when it comes to *drawing* circles, it fails at *filling* them. So we actually loose functionality in those cases where filled circles are needed. I switch ed back to the old code path in a5d0f6796095d8571d671e91ab208bc0c45016e3. Feel free to reopen if you have a code change to suggest. Otherwise we should probably open an issue instead, to keep track of this problem.

gagern:
I've managed to reproduce one instance of this issue, and reported it as [Chrome feedback](https://support.google.com/chrome/answer/95315). I used [this file](https://gist.github.com/gagern/fb8e28a8ff0bb34deb02) to reproduce the problem. Surprisingly, the issue only occurs in HiDPI mode for me. If I artificially set the device pixel ratio to 1 in CindyJS, the problem disappears.

Even more confusing: If I decide to close the path after drawing the arc, the circle becomes smooth. But this effect is apparently restricted to the reproducing example. If I add a `csctx.closePath();` after the `arc` command in `drawcircle` in CindyJS, the problem remains. This is *very* strange! If we can find a way to reproduce the effect this `closePath` has in the reproducing example, we might make use of that.

If you can reproduce any of the above with a reasonably recent Chromium, you may want to [post a bug report there](http://dev.chromium.org/for-testers/bug-reporting-guidelines). I may be able to test this on the Gentoo later tonight, but I make no promises.

gagern:
Note: The state saving from #208 has been immensely useful in making this problem reproducible. It allows us to interactively navigate to a problematic configuration, and then do

```js
JSON.stringify(createCindy.instances[0].saveState())
```

in the Web Developer console to store that configuration in a format that can be pasted into the original file. Then it can be run through various combinations of settings and debug output generation.

gagern:
> If we can find a way to reproduce the effect this `closePath` has in the reproducing example, we might make use of that.

Found the missing link and proposed #262 to deal with this.

PR #260, opened 2016-02-17
Negative radius

gagern:
I noticed that canvas implementations reject an arc command with negative radius. Cinderella however has no trouble drawing circles with negative radii. So we should take the absolute value of the radius when drawing. I also fixed some capitalization issues in the documentation which I noticed in passing.

I'd have liked to add a reftest for this, but don't know how to deal with the fact that all the other examples in the documentation on Elementary Drawing Functions yield over 4000 lines of drawing instructions. Adding all that makes the documenation alsmost unreadable. Should we ignore them, or write them to separate files? I guess I'd rather ignore them, and leave rendering tests to the planned recording facility.

### DIFF
--- a/ref/Elementary_Drawing_Functions.md
+++ b/ref/Elementary_Drawing_Functions.md
@@ -223,7 +223,7 @@ This operator can handle the following modifiers:
 
 **Description:**
 Draws a circle at `‹point›` with radius given by a number `‹radius›`.
-The point may be given either in euclidean or in homogeneous coordinates.
+The point may be given either in Euclidean or in homogeneous coordinates.
 
 **Modifiers:**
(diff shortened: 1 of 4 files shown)


strobelm:
Hi, the canvas.arc routine is quite buggy (at least in chromium). I replaced it with the one we use for conics (see https://github.com/CindyJS/CindyJS/pull/259) which also resolves the negative radius issue (unintentionally). If you consent to the changes i made we could cherry pick the documentation updates. 

On the reftest issue: i would ignore them.

gagern:
We should have the fix mentioned here in the commented-out code from #259 in my opinion. So that if we switch back, we get the fix included as well.

PR #262, opened 2016-02-17
Work around a bug in Chrome's arc drawing

gagern:
This aims to address #259 in a different way.

Through a lot of experimentation, I found out that the missing factor to make circle rendering work for me was the line join type. With closed paths and miter join, the circles are smooth as they should be.

@strobelm Please verify that this fixes the issue for you as well, both for the added example and for the original example where you first observed the problem.

### DIFF
--- a/examples/bugs/0259a_ChromeArc.html
+++ b/examples/bugs/0259a_ChromeArc.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../../build/js/Cindy.js"></script>
+<script id="csdraw" type="text/x-cindyscript">
+  drawcircle([3.4652946678, 0.9609442573], 4.0714703598);
+  drawcircle([-8.6989520431, -179.1999199497], 178.949860468);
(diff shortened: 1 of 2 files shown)


strobelm:
Unfortunately this does't fix the bug for me. I was playing around with the line width and found even more strange bugs:

linewidth = 1.169
arc1
linewidth = 1.17
arc2
linewidth = 1.171
arc3

Looks like 1.17 is the magic crash number or something. Will somebody look for a working conspiracy theory? Illuminati, Freemason?

gagern:
The longer I look at this, the stranger it is. I reported this issue as [Chromium issue 587582](https://code.google.com/p/chromium/issues/detail?id=587582), after reporducing it on Gentoo Linux. One really strange thing is that although the original problem went away if I removed the HiDPI adaptation, the weird rendering at line width 1.17 remains. At 1.1705 it spans both parts of the circle.

Despite all of this, I think we should merge this change here.
* It does improve the situation in Chrome on OS X.
* Closing the path is semantically the right thing to do, and might lead to visually more correct results if one looks closely at the endpoint, depending on the line cap style.
* The line join should not affect smooth nodes, and the circle should be completely smooth, so changing the line join should have no visible effect and therefore can't harm.

Issue #245, opened 2016-02-11
CindyGL and nested if


kranich:
CindyGL seems to have problems with nested if statements.

When I tried to implement my own hue function

```
hueToRGB(hue) := (
    hue = mod(hue / pi * 180, 360) / 60;
    h = floor(hue);
    f = hue - h;
    if(h == 0,
        [1,f,0]
    , if(h == 1,
        [1-f,1,0]
    , if(h == 2,
        [0,1,f]
    , if(h == 3,
        [0,1-f,1],
    , if(h == 4,
        [f,0,1],
    , if(h == 5,
        [1,0,1-f]
    ))))));
);
```

the following errors were thrown:

```
float[3] is no subtype of voidt (trying to cast the term vec3(float(1),float(0),(float(1)-f))) CindyGL.js:23:595

Could not find a signature for if(bool, float[3], voidt, voidt).
Returning empty code CindyGL.js:34:33

Could not find a signature for if(bool, float[3], voidt, voidt).
Returning empty code CindyGL.js:34:33

voidt is no subtype of float[3] (trying to cast the term )
```

The following workaround did not throw any errors:

```
hueToRGB(hue) := (
    hue = mod(hue / pi * 180, 360) / 60;
    h = floor(hue);
    f = hue - h;
    col = [1,f,0];
    if(h == 1, col = [1-f,1,0]);
    if(h == 2, col = [0,1,f]);
    if(h == 3, col = [0,1-f,1]);
    if(h == 4, col = [f,0,1]);
    if(h == 5, col = [1,0,1-f]);
    col;
);
```

Originally, I wanted to write

```
hueToRGB(hue) := (
    regional(h, f);
    hue = mod(hue / pi * 180, 360) / 60;
    h = floor(hue);
    f = hue - h;
    [[1,f,0],[1-f,1,0],[0,1,f],[0,1-f,1],[f,0,1],[1,0,1-f]]_(h+1);
);
```

but this did not work due to #243 and #244.

montaga:
@kranich, thank you for your detailed bug report.
Your last if did not have any else branch and CindyGL does not know which type it should assert here. If we change your Code to
```
      if(h == 0,
        [1,f,0]
    , if(h == 1,
        [1-f,1,0]
    , if(h == 2,
        [0,1,f]
    , if(h == 3,
        [0,1-f,1] //no , here
    , if(h == 4,
        [f,0,1] //no , here
    , if(h == 5,
        [1,0,1-f]
[...]

PR #263, opened 2016-02-17
always return either value of second argument of if$2 or glsl-default…

montaga:
… value of same type instead of trying to cast to void-type, which probably results in errors, as described in #245.

I cannot see where this will cause unwanted side effects.


kranich:
@montaga Thank you!

PR #274, opened 2016-02-23
CindyGL: Added `colorplot` and `imagergb[a]` function with no coordinate arguments

montaga:
Now it is also possible to call `colorplot("canvas", term)` instead of `colorplot(A, B, "canvasname", term)`. In this case we will simply assume that "canvas" has the same coordinates at the main-canvas, i.e. A and B will become the lower left and upper right corner.

In the same manner now one can read textures with `imagergb("canvas", coordinate)` instead of  `imagergb(A, B, "canvas", coordinate)`

This should make programs using feedback loops without coordinate change - which happens according to my experience in the many cases - more easy to write and read.

### DIFF
--- a/examples/cindygl/08_julia.html
+++ b/examples/cindygl/08_julia.html
@@ -28,12 +28,12 @@
       color(z) := (
         z = f(z, c);
         if(abs(z)<1.5,
-          imagergb([-1.5,-1.5], [1.5,-1.5], "julia", z) + 0.01*(1,2,3),
+          imagergb("julia", z) + 0.01*(1,2,3),
           (0,0,0)
         )
       );
       
-      colorplot([-1.5,-1.5], [1.5,-1.5], "julia", (color(complex(#))));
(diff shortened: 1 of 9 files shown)


kortenkamp:
Please check whether this makes sense - what if we have several views?

kranich:
@kortenkamp Then you can still specify points A and B for lower left respectively upper right corner, of course. This is just another option with reasonable defaults.

montaga:
In the same spirit of optional modifiers, I would consider `colorplot$2` and `imagergb[a]$2` as a additional feature that can be used in most cases while more special applications - as if one wants to handle several views or another coordinate system - are still possible with the old `colorplot$4`/`imagergb[a]$4`.

One might argue that one should consequently implement an `createimage$1` and  `drawimage$1`, which  creates an image with the resolution of the main-canvas and draws an image over the whole canvas resp.

kranich:
In today's CindyJS meeting, @richter-gebert agreed to merge this.

PR #275, opened 2016-02-24
center of conic (again)

strobelm:
Hi,

i was thinking about the code of #261 which i merged today and i think it's (mathematically) imprecise. 

The center of a conic isn't an euclidean object: consider a degenerate conic. The center of this is the intersection of the two lines the conic collapses into. This will be, of course, a far point iff the lines are parallel. 

So Cinderella is wrong in this case and we should discuss whether we change its (her?) behavior.

### DIFF
--- a/examples/109_CenterOfConic.html
+++ b/examples/109_CenterOfConic.html
@@ -5,6 +5,11 @@
 <title>Cindy JS Example</title>
 <meta charset="UTF-8">
 <script type="text/javascript" src="../build/js/Cindy.js"></script>
+
+<script id="csdraw" type="text/x-cindyscript">
+
+println(X.homog);
+</script>
 <script type="text/javascript">
 
 var cdy = createCindy({ // See ref/createCindy documentation for details.
--- a/src/js/libcs/Accessors.js
+++ b/src/js/libcs/Accessors.js
@@ -33,21 +33,14 @@ Accessor.setGeoField = function(geoname, field, value) {
     return nada;
 };
 
-function dehom(v) {
-    v = v.value.slice();
-    var n = v.length - 1;
-    var d = CSNumber.inv(v[n]);
-    v.length = n;
-    for (var i = 0; i < n; ++i)
-        v[i] = CSNumber.mult(d, v[i]);
-    return List.turnIntoCSList(v);
-}
 
 Accessor.getField = function(geo, field) {
     var erg;
     if (geo.kind === "P") {
         if (field === "xy") {
-            erg = dehom(geo.homog);
+            var xx = CSNumber.div(geo.homog.value[0], geo.homog.value[2]);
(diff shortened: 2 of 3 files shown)


gagern:
:-1:

While I agree from an idealistic point of view that having the center be given in Euclidean coordinates is suboptimal, from a pragmatic point of view I have to observe that there may well be *a lot* of Cinderella code out there which makes use of `.center`, and which would break by such a change. Therefore compatibility forces us to support that, I think. There are also the majority of users to consider which neither know nor care about projective geometry. (Currently this perhaps only applies to CInderella, due to the still small CindyJS user base, but things will change.) To them, the Euclidean coordinates are the right thing, while the homogeneous ones are confusing. Remember that in most situation where a function expects a list but the user provides a point, that point is converted to its Euclidean coordinates, not its homogeneous ones.

We *might* introduce the homogeneous center using a different name. I'm open for suggestions. `homcenter`? `hcenter`? I also was wondering whether we could make the syntax `C0.center.homog` work. Not only work technically, but also semantically so that we can clearly define what kind of object `C0.center`actually is, to look like a two-element list but still provide such a `homog` field. While we are at it, we might also want to make the dual matrix accessible. Do you prefer `dmatrix, `dual` or something else?

I also have strong objections to 1e803129f1ed08fa1ecc77675446eeca02b3ea15. What makes you define the center of the conic as the intersection of the lines in the degenerate situation? If you consider the special case of circles, then the computation of the center can be simplified by canceling a factor. (Perhaps we should actually do this, depending on the usage of the matrix.) In the degenerate case, the circle factors into a (probably finite) line and the line at infinity. The center then becomes the point at infinity orthogonal to the first line. Which is different from the point of intersection. In my opinion, any special case for degenerate conics must capture this scenario the same way before I start believing in it.

kortenkamp:
I also think that `.center` must not be changed. And for circles (the usual application for `.center`) the center rarely is at infinity (the usual application for homogeneous coordinates).

gagern:
> What makes you define the center of the conic as the intersection of the lines in the degenerate situation?

I realized that this is indeed what happens if the point of intersection is finite. I still maintain that it is incorrect to generalize this to intersections at infinity. For example, consider the family of conics *ax*<sup>2</sup> + *y*<sup>2</sup> = 1. For *a* ≠ 0 these have a center at the origin, (0,0,1). For *a* = 0 you get a pair of parallel lines. Making the “center” jump to (1, 0, 0) is unmotivated and will break tracing. Barring more detailed knowledge, I think it is correct to return the null vector if the two lines of a degenerate conic intersect at infinity.

Considering non-degenerate conics, those with center at infinity are exactly the parabolas, unless I made a mistake. So the question is whether we want a new field to access the centers of parabolas and of of degenerate circles. I see no urgent need for this; I think we might want to add this one day but have more urgent issues right now. If you agree, I'd close this merge request as wontfix, with the express note that this does not preclude adding a new field instead.

PR #277, opened 2016-02-26
Use global cindyjs.externs for plugins

kranich:
There should be only one cindyjs.externs file that defines the plugin API for all plugins.


montaga:
Sounds reasonable :-)

CindyGL also needs
```
getImage: function(string, boolean),
getMyfunction: function(string)
```
(compare `plugins/cindygl/src/js/cindyjs.externs`)

kranich:
@montaga This is why I made `plugins/cindygl/src/js/cindyjs.externs` the default version in my pull request.

montaga:
Ah, now I see :smile: Did not know how to read the diff.

Issue #207, opened 2016-02-03
normalizeMax real or complex


gagern:
Currently we are using `normalizeMax` for two effects:

1. Avoid overflow. This is done after most elementary operations, so it's a very common operation.
2. Make all coordinates real if there is a real representative. This property is required when after the tracing phase [an almost real number is made perfectly real](https://github.com/CindyJS/CindyJS/blob/f7a64a1e237956e478b52ee43fbac233116f83a5/src/js/libgeo/Tracing.js#L56-L65).

Since the former of these approaches is used so much more often, I wonder whether we should make an effort to make this computationally cheaper. For example, we could treat all complex numbers as pairs of real numbers, and just take the maximal component there. That would avoid a complex inversion and many complex multiplications, making the whole normalization step just operate on real values. Doing so should be cheaper and may even reduce numeric errors.

If we do this, I suggest introducing names `normalizeMaxR` and `normalizeMaxC` to distinguish whether the entries are seen as just a bunch of real numbers, or as complex numbers. The former would be called in most places, the latter only at the very end of a trace.

strobelm:
A further problem of normalizeMax is a change of signs (or phases) which
causes unnecessary discontinuities in our operations. For example take
sqrt of some determinant, if the normalization changes the sign of the
determinant from + to -  we introduce and additional i as factor which
has to be captured by tracing. This yields additional difficulties
tracing roots only instead of whole vectors.

It would be nice to normalize e.g. by absolute value in order to
preserve the phase.

Am 03.02.2016 um 10:48 schrieb Martin von Gagern:
>
> Currently we are using |normalizeMax| for two effects:
>
> 1 Avoid overflow This is done after most elementary operations, so
> it's a very common operation
> 2 Make all coordinates real if there is a real representative This
> property is required when after the tracing phase an almost real
> number is made perfectly real
> <https://githubcom/CindyJS/CindyJS/blob/f7a64a1e237956e478b52ee43fbac233116f83a5/src/js/libgeo/Tracingjs#L56-L65>
>
> Since the former of these approaches is used so much more often, I
> wonder whether we should make an effort to make this computationally
> cheaper For example, we could treat all complex numbers as pairs of
> real numbers, and just take the maximal component there That would
> avoid a complex inversion and many complex multiplications, making the
> whole normalization step just operate on real values Doing so should
> be cheaper and may even reduce numeric errors
>
> If we do this, I suggest introducing names |normalizeMaxR| and
[...]

PR #280, opened 2016-02-29
Reimplement conic drawing

gagern:
This is a rebased version of [my conicDrawing branch](https://github.com/CindyJS/CindyJS/compare/a5d0f6796095d8571d671e91ab208bc0c45016e3...gagern:conicDrawing?expand=0). Some aspects have already been discussed by mail. in general it should be an improvement over what we have so far.


src/js/libcs/OpDrawing.js
```diff
@@ -393,385 +393,339 @@ evaluator.drawconic$1 = function(args, modifs) {
...
-        for (var j = 0; j < 2; j++) {
-            if (Math.abs(mat.value[i].value[j].value.imag) > CSNumber.eps) return;
+    var tmat = csport.toMat();
+    mat = List.mult(List.transpose(tmat), mat);
+    mat = List.mult(mat, tmat);
+    mat = List.normalizeMax(mat);
+
+    // Using polynomial coefficients instead of matrix
+    // since it generalizes to higher degrees more easily.
+    // cij is the coefficient of the monomial x^i * y^j.
```
strobelm: We should add some checking for complex values here, otherwise 

```
C0 = [-0.31, -0.27, -0.13, 2.8, 0.09, 1];
C3 = [-0.31+i, -0.27, -0.13+i, 2.8, 0.09, 1];
``` 

will be drawn as the same conic. Choosing the epsilon here should not be too strict, otherwise we run into a blinking Christmas tree of disappearing and reappearing circles.

src/js/libcs/OpDrawing.js
```diff
@@ -393,385 +393,339 @@ evaluator.drawconic$1 = function(args, modifs) {
...
+    var c01 = mat.value[1].value[2].value.real * 2;
+    var c00 = mat.value[2].value[2].value.real;
+
+    var det = Math.abs(c20 * c02 * c00 * 4 + c11 * c01 * c10 -
+        c00 * c11 * c11 - c20 * c01 * c01 - c02 * c10 * c10);
+    if (det < eps) {
+        var discr = c20 * c02 * 4 - c11 * c11;
+        if (discr < eps) { // pair of lines
+            var lines = geoOps._helper.splitDegenConic(conicMatrix);
+            if (lines !== nada) {
```
strobelm: I think splitDegenConic does not return nada explicitly if something goes wrong. The lines themselves could be nada, though.

src/js/libcs/OpDrawing.js
```diff
@@ -393,385 +393,339 @@ evaluator.drawconic$1 = function(args, modifs) {
...
+    var c11 = mat.value[0].value[1].value.real * 2;
+    var c10 = mat.value[0].value[2].value.real * 2;
+    var c02 = mat.value[1].value[1].value.real;
+    var c01 = mat.value[1].value[2].value.real * 2;
+    var c00 = mat.value[2].value[2].value.real;
+
+    var det = Math.abs(c20 * c02 * c00 * 4 + c11 * c01 * c10 -
+        c00 * c11 * c11 - c20 * c01 * c01 - c02 * c10 * c10);
+    if (det < eps) {
+        var discr = c20 * c02 * 4 - c11 * c11;
```
elkins0: 'discr' occurs as a subexpression in several places, including in 'det' above. Moving 'discr' out and up, one can write::
```javascript
    var discr = c20 * c02 * 4 - c11 * c11;
    var det = Math.abs(discr * c00 + c11 * c01 * c10 - c20 * c01 * c01 - c02 * c10 * c10);
```

src/js/libcs/OpDrawing.js
```diff
@@ -393,385 +393,339 @@ evaluator.drawconic$1 = function(args, modifs) {
...
+        x = -c01 / c11;
+        pt = mkpg(x, Infinity, 1, 0);
+        specialPoints.push([pt, pt]);
+        pt = mkpg(x, Infinity, 1, 0);
+        specialPoints.push([pt, pt]); // This is a double double point
+    } else {
+        // Compute the roots of the y discriminant
+        // for points with vertical tangents
+        sol = solveRealQuadratic(
+            c11 * c11 - 4 * c02 * c20,
```
elkins0: Here again the expression 'discr' occurs as its negative:
```javascript
-            c11 * c11 - 4 * c02 * c20,
+           -discr,
```

src/js/libcs/OpDrawing.js
```diff
@@ -393,385 +393,339 @@ evaluator.drawconic$1 = function(args, modifs) {
...
-                y1 = r1.value[1].value.real;
-                y2 = r2.value[1].value.real;
-            } else {
-                y1 = y;
-                y2 = y;
+    if (Math.abs(c20) >= eps) { // no horizontal asymptote
+        // Compute the roots of the x discriminant
+        // for points with horizontal tangents
+        sol = solveRealQuadratic(
+            c11 * c11 - 4 * c02 * c20,
```
elkins0: Add again the expression 'discr' occurs as its negative:
```javascript
-            c11 * c11 - 4 * c02 * c20,
+           -discr,
```

examples/63_Conic.html
```diff
@@ -36,12 +37,45 @@
...
                       ];
-            createCindy({canvasname:"CSCanvas",
+            var cdy=createCindy({canvasname:"CSCanvas",
                         movescript:"csmove",
                         grid:1,
                         snap:true,
                         geometry:gslp});
 
+            var cases = [
+                ["parabola1", [0,0,1], [1,1,1], [2,4,1], [-2,4,1], [-1,1,1]],
```
elkins0: When instrumented, clicking on the 'parabola1' button shows a simple stroke sequence:
```javascript
moveTo(157.02540452080043,0)
quadraticCurveTo(250.49999999999997,698.9999999999992,343.97459547919953,0)
```
Adding positioning for the rotation of 'parabola1' by 45&deg; clockwise:
```javascript
 +                ["parabola2", [0, 0, 1], [0, 2, 1], [6, 2, 1], [2, 6, 1], [2, 0, 1]],
```
and then clicking on the 'parabola2' however yields a sequence consisting of 80 quadraticCurveTo calls. It would seem that this could be a simple sequence containing just one quadraticCurveTo call.

Adding positioning for a circle:
```javascript
 +                ["circle1", [6,1,1], [10,3,1], [10,9,1], [2,9,1], [2,3,1]],
```
and then clicking on the 'circle1' yields a sequence consisting of 128 quadraticCurveTo calls interspersed with two lineTo calls.
```javascript
moveTo(275.5,199.5)
quadraticCurveTo(275.5,196.43142223638486,275.65056797435346,193.3665407090727)
...
quadraticCurveTo(397.4314222363835,74.50000000000011,400.5,74.50000000000011)
lineTo(400.5,74.50000000000001)
quadraticCurveTo(403.56857776361465,74.50000000000001,406.6334592909267,74.65056797435348)
...
quadraticCurveTo(403.56857776361323,324.5,400.5,324.5)
lineTo(400.5,324.4999999999999)
quadraticCurveTo(397.43142223638347,324.4999999999999,394.36654070907263,324.3494320256465)
...
quadraticCurveTo(275.5,202.5685777636156,275.5,199.5)
closePath()
```
The use of 'lineTo' appears to be a bit over zealous.

It would seem that this should be a simple sequence containing just four quadraticCurveTo calls (or perhaps eight for very large circles) as is typically done in PDF files to no perceptible visual difference.

elkins0: Also something strange I noticed (perhaps something inadvertently was changed by me), clicking the buttons 'circle1',  'parabola2' and 'parabola1' in that order ends up drawing only the left portion of the 'parabola1'.

src/js/libcs/OpDrawing.js
gagern: You are right. Thanks for spotting this. For some reason I had assumed that the caller would take care of this, but that's wrong, it only checks that it's a numeric matrix. `CSNumber.isAlmostReal` uses `CSNumber.epsbig = 1e-6` instead of `CSNumber.eps = 1e-10`. Perhaps we should do the same here?

src/js/libcs/OpDrawing.js
gagern: Good catch, thank you very much!

src/js/libcs/OpDrawing.js
gagern: No, the lines can't be nada either, only zero vectors. See 127ce6b8b9824e2a11a4a7afb1255f7c0b240ba2.

examples/63_Conic.html
gagern: I don't see `parabola1` getting drawin in a single stroke. At what level did you instrumemt this? Inside the CindyJS code, or in the Browser somehow? In any case, the termination criterion is checking the area of the convex hull of the two endpoints and the control point. That's likely overestimating the approximation error, particularly in case of a parabola. But I can think of no better error measure which could be computed in reasonable time. The distance between the center point and the conic was something I tried before, and that failed to detect lack of fidelity in almost-degenerate hyperbola situations in particular. Can you think of some other error measure, @elkins0?

The `lineTo` calls I could reproduce. The distance spanned is *very* short, so the tangent lines are almost identical, triggering the `Math.abs(denom) > 1e12` check. I doubt that refinement led to such close points. Instead I assume that due to numeric errors, the two horizontal tangents are seen at slightly different *x* values. We might want to detect such cases, and merge them into one, but on the other hand the current behavior shouldn't have much of a performance impact either, I think.

We might do special handling of circles and circular arcs one day, using cubic approximations with even the endpoints slightly off the curve to reduce overall error. But that's a different matter, to be handled in a separate pull request.

gagern:
@strobelm: Incorporated review comments, and made Travis happy again. Is it ready now?

examples/63_Conic.html
elkins0: @gagern:The instrumentation was added entirely within `eval_helper.drawconic = function(conicMatrix, modifs)` next to any `csctx` calls:
```javascript
            console.log("lineTo(" + [pt2.px, pt2.py] + ")");
            return csctx.lineTo(pt2.px, pt2.py);
...
            console.log("lineTo(" + [pt2.px, pt2.py] + ")");
            return csctx.lineTo(pt2.px, pt2.py);
...
        console.log("quadraticCurveTo(" + [cx, cy, pt2.px, pt2.py] + ")");
        csctx.quadraticCurveTo(cx, cy, pt2.px, pt2.py);
...
    csctx.beginPath();
    console.log("beginPath(" + starts.length + ", " + specialPoints.length + ")");
...
        csctx.moveTo(pt.px, pt.py);
        console.log("moveTo(" + [pt.px, pt.py] + ")");
...
                csctx.closePath();
                console.log("closePath()");
...
    csctx.stroke();
    console.log("stroke(" + starts.length + ", " + specialPoints.length + ")");
```

The console log shows:
```
beginPath(13, 8)
moveTo(157.0254045208004,0)
quadraticCurveTo(250.50000000000003,699.0000000000002,343.9745954791996,0)
stroke(13, 8)
```
Applying a 45&deg; rotation matrix to the those control points would give the parabola2 (although not clipped correctly).

elkins0:
@gagern:For Reference
1. parabola1's equation is `x^2==y`.
2. parabola2's equation is `(x-y)^2==2*(x+y)`.
3. parabola2 is parabola1 rotated 45&deg; and (x|y) scaled by sqrt(2).
4. The control points are the two intersection points at the boundary and the intersection point of their tangent lines to the curve.

So parabola2 here can be rendered exactly using only:
```
moveTo(436.25413911823125,0)
quadraticCurveTo(-18.127069559115228,618.1270695591152,600,163.74586088176875)
```
CindyJS versions of [parabola1.html](https://gist.github.com/elkins0/a47faff1972972e54e10) and [parabola2.html](https://gist.github.com/elkins0/6533e5e9079e95eed7bc) along with [parabolaPL.html](https://gist.github.com/elkins0/963a2a876a7ff7a34b47), a version of [parabola2.html]() using `ConicParabolaPL` instead of `ConicBy5` each show the three control points needed to exactly render the curve. They also use `drawbezier([L,N,M],color->(1,0,0));
` to render the curve ([B,D,C] in [parabolaPL.html](https://gist.github.com/elkins0/963a2a876a7ff7a34b47)).

It is a shame that there was no Bézier curve support in CindyScript so I hacked OpDrawing.js to remedy that :smile:
```javascript
evaluator.drawbezier$1 = function(args, modifs) {
    return eval_helper.drawpolygon(args, modifs, "B", false);
};

eval_helper.drawpolygon = function(args, modifs, df, cycle) {
...
    function drawpoly() {
        var i;
        for (i = 0; i < v0.value.length; i++) {
            var pt = eval_helper.extractPoint(v0.value[i]);
            if (!pt.ok) {
                return;
            }
            var xx = pt.x * m.a - pt.y * m.b + m.tx;
            var yy = pt.x * m.c - pt.y * m.d - m.ty;
            if (i === 0)
                csctx.moveTo(xx, yy);
            else if (df === "B") { // Handle Bézier mode
                i++;
                var pt2 = eval_helper.extractPoint(v0.value[i]);
                if (!pt2.ok) {
                    return;
                }
                var xx2 = pt2.x * m.a - pt2.y * m.b + m.tx;
                var yy2 = pt2.x * m.c - pt2.y * m.d - m.ty;
                csctx.quadraticCurveTo(xx, yy, xx2, yy2);
            } else
                csctx.lineTo(xx, yy);
        }
        if (cycle)
            csctx.closePath();
    }
...
    if (df === "D" || df === "B") {
        csctx.stroke();
    }
```
In [parabolaPL.html](https://gist.github.com/elkins0/963a2a876a7ff7a34b47), moving just point A to near (6|-6) shows `eval_helper.drawconic` failing, drawing the curve as two intersecting lines.
parabolaplfail

elkins0:
The example [ConicBy2Foci1P-drawconic.html](https://gist.github.com/elkins0/63bf14b6bdb50db6ab46) using `ConicBy2Foci1P` demonstrates a failing shown in the images below where the hyperbola switches between being drawn correctly and then incorrectly as a set of parallel lines as point B is jiggled.
conicby2foci1p-drawconic ok
conicby2foci1p-drawconic error

elkins0:
[49-TangentParabola2linesAt2points.html](https://gist.github.com/elkins0/003c975cc7e6ef27ed4a) constructs a parabola tangent to two lines a and b at points A and B. The Bézier curve control point C is the intersection of the lines a and b. Conversely, [ParabolaOfBézierControlPoints.html](https://gist.github.com/elkins0/d46d593d80ac25984bbf) constructs the Bézier curve control point C first and then the intersecting lines a and b through it and points A and B respectively. That construction is more amenable to finding problems with `eval_helper.drawconic` since the Bézier curve control points A, B and C can be directly manipulated. The buttons show various cases where `eval_helper.drawconic` fails:

parabolaofbeziercontrolpoints-6
*is not a single line*

parabolaofbeziercontrolpoints-4
*is not a pair of lines*

parabolaofbeziercontrolpoints-5
*incomplete curve*

gagern:
@elkins0 Thanks for these great examples! I've had some hope that all our problems were due to the fact that we switched to “pair of lines” too soon, so changing a single threshold would fix everything. But that last example appears to be of a different nature, and I don't yet understand what's happening there. Needs to be addressed, though.

elkins0:
[10 days ago](https://github.com/CindyJS/CindyJS/pull/280/#discussion_r55433297) @gagern wrote:
> In any case, the termination criterion is checking the area of the convex hull of the two endpoints and the control point. That's likely overestimating the approximation error, particularly in case of a parabola. But I can think of no better error measure which could be computed in reasonable time. The distance between the center point and the conic was something I tried before, and that failed to detect lack of fidelity in almost-degenerate hyperbola situations in particular. Can you think of some other error measure, @elkins0?

The following code works fine as a termination criterion for all the cases presented in 63_Conic.html.
Within `eval_helper.drawconic`, inside `refine`, remove the area check and add a check that the new point on the actual curve is more than half a pixel away from the point on the Bézier curve:
```javascript
     function refine(pt1, pt2, sign, depth) {
...
-        var area = Math.abs(
-            pt1.px * cy + cx * pt2.py + pt2.px * pt1.py -
-            pt2.px * cy - cx * pt1.py - pt1.px * pt2.py);
-        if (area > 1 && depth < 10) {
+        if (depth < 10) {
+            // The midpoint of the control point (mx|my) and the midpoint of pt1 and pt2 lies on the Bézier curve
             var mx = 0.5 * cx + 0.25 * (pt1.px + pt2.px);
             var my = 0.5 * cy + 0.25 * (pt1.py + pt2.py);
...
                 if (pt3) {
-                    refine(pt1, pt3, sign, depth + 1);
-                    refine(pt3, pt2, sign, depth + 1);
+                    var dx = pt3.px - mx;
+                    var dy = pt3.py - my;
+                    if (dx * dx + dy * dy > 0.25) { // compare to the distance in pixels squared
+                        refine(pt1, pt3, sign, depth + 1);
+                        refine(pt3, pt2, sign, depth + 1);
+                        return;
+                    }
                 }
```
Feel free to use [ConicTangentAtTwoLinesThruOnePoint.html](https://gist.github.com/elkins0/62019296ce9f2d0a067f) to experiment and find new test cases.

elkins0:
BTW @gagern, [ConicTangentAtTwoLinesThruOnePoint.html](https://gist.github.com/elkins0/62019296ce9f2d0a067f), which uses `geoOps.ConicBy3p2l` produces zeroed matrices for degenerate conics along with four identical solutions. To improve upon that we can use `geoOps._helper.conicFromTwoDegenerates` and implement `geoOps.ConicBy2t1p` shown below and use it instead to create the conic with:
```javascript
{ name: "C0", type: "ConicBy2t1p", color: [ 0, 0, 1 ], args: [ "A", "a", "B", "b", "C" ] }
```
Similarly, the conic in [ParabolaOfBézierControlPoints.html](https://gist.github.com/elkins0/d46d593d80ac25984bbf) can be produced by implementing `geoOps.ParabolaBy2t` also shown below and using it instead to create the conic with:
```javascript
{ name: "C0", type: "ParabolaBy2t", color: [ 0, 0, 1 ], args: [ "A", "a", "B", "b" ] }
```
Given `["A", "a", "B", "b", "P"]` where `a` is a line thru `A`, `b` line is a thru `B`, 
and `P` is a point on the conic then `geoOps.ConicBy2t1p` produces the conic given the two tangencies (a tangent point on a line) and the point on the conic:
```javascript
geoOps.ConicBy2t1p = {};
geoOps.ConicBy2t1p.kind = "C";
geoOps.ConicBy2t1p.signature = ["P", "L", "P", "L", "P"];
geoOps.ConicBy2t1p.updatePosition = function(el) {
    var A = csgeo.csnames[(el.args[0])].homog;
    var a = csgeo.csnames[(el.args[1])].homog;
    var B = csgeo.csnames[(el.args[2])].homog;
    var b = csgeo.csnames[(el.args[3])].homog;
    var P = csgeo.csnames[(el.args[4])].homog;

    var c = List.cross(A, B);
    var v23 = List.turnIntoCSList([a]);
    var v14 = List.turnIntoCSList([b]);
    var v1234 = List.turnIntoCSList([c]);

    var matrix = geoOps._helper.conicFromTwoDegenerates(v23, v14, v1234, v1234, P);
    el.matrix = General.withUsage(List.normalizeMax(matrix), "Conic");
};
```
Given `["A", "a", "B", "b", "P"]` where `a` is a line thru `A` and `b` line is a thru `B`, 
then `geoOps.ParabolaBy2t` produces the parabola given the two tangencies (a tangent point on a line):
```javascript
geoOps.ParabolaBy2t = {};
geoOps.ParabolaBy2t.kind = "C";
geoOps.ParabolaBy2t.signature = ["P", "L", "P", "L"];
geoOps.ParabolaBy2t.updatePosition = function(el) {
    var A = csgeo.csnames[(el.args[0])].homog;
    var a = csgeo.csnames[(el.args[1])].homog;
    var B = csgeo.csnames[(el.args[2])].homog;
    var b = csgeo.csnames[(el.args[3])].homog;
    var mid = geoOps._helper.midpoint;
    var P = mid(mid(A, B), List.cross(a, b));

    var c = List.cross(A, B);
    var v23 = List.turnIntoCSList([a]);
    var v14 = List.turnIntoCSList([b]);
    var v1234 = List.turnIntoCSList([c]);

    var matrix = geoOps._helper.conicFromTwoDegenerates(v23, v14, v1234, v1234, P);
    el.matrix = General.withUsage(List.normalizeMax(matrix), "Conic");
};
```

gagern:
I've changed the threshold for switching from non-degenerate conic to pair-of-lines from 10<sup>−14</sup> to 10<sup>−24</sup>, replacing one arbitrary choice with another one. I also fixed two bugs related to rendering of parabolas in particular. Now things look better; all the examples @elkins0 provided render correctly. I've made no effort to render parabolas with as few strokes as possible, since I don't see much benefit in doing so.

elkins0:
@gagern:Here are some additional test cases for `63_Conic.html` which reveal problems with rendering degenerate conic curves:
```javascript
["degen4", [-2, 8, 1],      [2.2, 4, 1], [6, 2, 1], [2, 2, 1], [1, 2, 1]],
["degen5", [-1.72, 7.88, 1],  [2, 5, 1], [6, 2, 1], [3, 2, 1], [2, 2, 1]],
["degen6", [-5.24, 10.68, 1], [2, 5, 1], [6, 2, 1], [3, 2, 1], [2, 2, 1]],
["degen7", [-0.92, 6.8, 1],   [2, 5, 1], [6, 2, 1], [3, 2, 1], [2, 2, 1]],
["degen8", [-0.6800000000000002, 6.68, 1], [2, 5, 1], [6, 2, 1], [3, 2, 1], [2, 2, 1]],
```
Note that most of these are only different in the position of the point `A`. Moving point `A` alone produces considerable chaotic behavior. For example, `degen7` renders as:
63_conic-failed-degen7

elkins0:
@gagern wrote:
> I've made no effort to render parabolas with as few strokes as possible, since I don't see much benefit in doing so.

It is not just parabolas that could benefit from being rendered with fewer strokes. Using `63_Conic.html` here is a table of the count of the `quadraticCurveTo` calls made by my alternative loop termination code versus the the current implementation for various curves:

```
parabola1    2     64
hyperbola1   8    128
ellipse1    14    184
ellipse2    10    144
corners      6     96
```
As you can see the current implementation makes an order of magnitude greater in calls to `quadraticCurveTo`. Perhaps you might consider using my alternative loop termination instead...

gagern:
@elkins0 Honestly I don't trust your termination criterion – yet. I guess I'll implement it and then see whether or not I can find an example where mine works but yours fails. I guess I should also change the (pretty arbitrary) threshold in my code to produce approximately the same number of drawing operations as yours, and then see whether there is a difference in fidelity, i.e. which of the two approaches is more likely to spend work where it's actually needed.

gagern:
I fear that the new test cases you suggest are the flipside of the threshold change from gagern@7198d9d0f76080528c91006a9d05022badf6408a. Before we'd get conics rendered as lines, now we get lines rendered as conics, and incorrectly as well. I guess I could try tweaking the threshold to see whether I find a setting which satisfies both kinds of cases. But I'm beginning to think that perhaps I should concentrate on ensuring that the code for non-degenerate cases handles degenerate cases as well. In any case, thank you very much for the new cases, @elkins0!

strobelm:
@gagern whats the status pull request? May i help finding appropriate epsilons?

gagern:
I just had another look at this. I added the extra test cases from https://github.com/CindyJS/CindyJS/pull/280#issuecomment-209184409 but failed to reproduce the behavior demonstrated in the screen shot there. @elkins0, can you still demonstrate that behavior, if you enter the examples as you gave them into the file, and using the current CindyJS file from this branch? If so, what browser is that?

elkins0:
The reason for the reproducibility difficulties are that I use power-of-two divisors in normalizeMax (see https://github.com/CindyJS/CindyJS/issues/207#issuecomment-181631444) which helps with idempotency. I built gagern:conicDrawing2 to create `gagern/CindyJS/build/js/Cindy.js` and used `gagern/CindyJS/examples/63_Conic.html` to rework the settings for only `degen4` and `degen7` by jiggling the mouse and then using the result logged to the console by `Write to console`. The others I listed before gave very similar results with my setup. You'll need to add only these lines to `gagern/CindyJS/examples/63_Conic.html` to get what is shown below:
```javascript
                ["degen4", [-0.25, 1, 0.125], [0.4953271028037384, 1, 0.23364485981308416], [1, 0.33333333333333326, 0.16666666666666663], [1, 1, 0.5], [0.5, 1, 0.5]],
                ["degen7", [-0.09883720930232559, 1, 0.1453488372093023], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
```
Note that if you click on point C (nearest the intersection) after clicking `degen4`, you get two lines, the expected result. Here is the result of clicking `degen4`:
degen4g
Here is the result of clicking `degen7`:
degen7g
The results are for Internet Explorer 11 and are the same in FireFox 46.

elkins0:
@gagern, @strobelm: I have further expanded upon the setting choices (`degen4` and `degen7` listed above are included here):
```javascript
                ["degen4", [-0.25, 1, 0.125], [0.4953271028037384, 1, 0.23364485981308416], [1, 0.33333333333333326, 0.16666666666666663], [1, 1, 0.5], [0.5, 1, 0.5]],
                ["degen5", [-0.25, 1, 0.125], [0.5643564356435643, 1, 0.24752475247524752], [1, 0.3333333333333333, 0.16666666666666666], [1, 1, 0.5], [0.5, 1, 0.5]],
                ["degen6", [-0.4962406015037595, 1, 0.09398496240601504], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
                ["degen7", [-0.09883720930232559, 1, 0.1453488372093023], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
                ["degen8", [-0.5111111111111112, 1, 0.0925925925925926], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
```
Clicking the settings in order produces what I have seen. Interestingly, if upon first loading `gagern/CindyJS/examples/63_Conic.html ` one clicks `degen8` everything seems fine. Yet if `degen8` is clicked a second time this is the result (note the tiny blemish in the curve above the intersection):
degen8g
`degen6` also has a behavior like `degen8`. `degen7`, if clicked first after loading, is wrong in a different way until clicked a second time. Clicking `parabola1` and then clicking `degen5` is almost perfect except for a tiny blemish.
The results are the same for both *Internet Explorer 11* and *FireFox 46.0*.

strobelm:
Still some major glitches in here:

con_bug2


With additionally Segment clipping #343 merged:
con_bug

I really like the look of the drawing (if one does not consider the bugs ;-)), would be great to have!

strobelm:
I'll deassign myself until the branch is ready for review again. It would be much appreciated if we could have this at some time in the future.

elkins0:
@gagern : Here is my fix for the problem positions I gave above. I replaced most of the refine function and used this solveRealQuadraticX function within the refine function only. The testing I did used the gagern:conicDrawing2 branch code.

```javascript
// See also eval_helper.quadratic_roots for the complex case
// Returns either null (if solutions would be complex or NaN)
// or two solutions satisfying ax^2 + bx + c == 0
// or one solution satisfying bx + c == 0 when a === 0
function solveRealQuadraticX(a, b, c) {
    if (Math.abs(a) <= CSNumber.eps) { // also a is not NaN
    /*jshint -W018 */
        if (!(Math.abs(b) > CSNumber.eps)) return null; // also return null if b is NaN
    /*jshint +W018 */
        return [-c / b];
    }
    // We use this form of solution so that d is always in units of pixels squared
    // Note that given CSNumber.eps === 10^-10, sqrt(CSNumber.eps) is 10^-5 pixels
    var b2a = -b / (2 * a);
    var d = b2a * b2a - c / a;
    /*jshint -W018 */
    if (!(d >= -CSNumber.eps)) return null; // also return null if d is NaN
    /*jshint +W018 */
    var r = d <= CSNumber.eps ? 0 : Math.sqrt(d);
    return [b2a - r, b2a + r];
}
```

```javascript
    // Find the control points of a quadratic Bézier which at the
    // endpoints agrees with the conic in position and tangent direction.
    // Unused: sign is -1 for the lower branch and +1 for the upper.
    function refine(pt1, pt2, sign, depth) {
        // (x - pt1.px) * pt1.gx + (y - pt1.py) * pt1.gy = 0
        // x * pt1.gx + y * pt1.gy = pt1.dot
        // x * pt2.gx + y * pt2.gy = pt2.dot
        // Solve using Cramer's rule
        var denom = 1 / (pt1.gx * pt2.gy - pt2.gx * pt1.gy);
        if (Math.abs(denom) > 1e12)
            return csctx.lineTo(pt2.px, pt2.py);
        var cx = (pt1.dot * pt2.gy - pt2.dot * pt1.gy) * denom;
        var cy = (pt1.gx * pt2.dot - pt2.gx * pt1.dot) * denom;
        if (!(isFinite(cx) && isFinite(cy))) // Probably already linear
            return csctx.lineTo(pt2.px, pt2.py);
        if (depth < 10) {
            // Compute pt3 as the intersection of the segment (hx|hy) & (cx|cy) and the conic
            var hx = 0.5 * (pt1.px + pt2.px);
            var hy = 0.5 * (pt1.py + pt2.py);
            var rdx = cx - hx;
            var rdy = cy - hy;
            if (rdx * rdx + rdy * rdy > 0.25) { // compare to the distance in pixels squared
                var ua = c20 * rdx * rdx + c11 * rdx * rdy + c02 * rdy * rdy;
                var hrxy = hy * rdx - hx * rdy;
                var sol, pt3 = null;
                if (Math.abs(rdy) < Math.abs(rdx)) { // Avoids divide by zero
                    sol = solveRealQuadraticX(ua,
                        rdx * (c10 * rdx + c01 * rdy) + (c11 * rdx + 2 * c02 * rdy) * hrxy,
                        c00 * rdx * rdx + (c01 * rdx + c02 * hrxy) * hrxy);
                    if (sol) {
                        var x3 = sol.length > 1 && inRange(sol[1], hx, cx) ? sol[1] : sol[0];
                        pt3 = mkp(x3, hy + rdy * (x3 - hx) / rdx);
                    }
                } else {
                    sol = solveRealQuadraticX(ua,
                        rdy * (c10 * rdx + c01 * rdy) - (c11 * rdy + 2 * c20 * rdx) * hrxy,
                        c00 * rdy * rdy - (c10 * rdy - c20 * hrxy) * hrxy);
                    if (sol) {
                        var y3 = sol.length > 1 && inRange(sol[1], hy, cy) ? sol[1] : sol[0];
                        pt3 = mkp(hx + rdx * (y3 - hy) / rdy, y3);
                    }
                }
                if (pt3) {
                    // The midpoint (mx|my) of the control point (cx|cy) and the
                    // midpoint of pt1 and pt2 (hx|hy) lies on the Bézier curve
                    var mx = 0.5 * (cx + hx);
                    var my = 0.5 * (cy + hy);
                    var dx = pt3.px - mx;
                    var dy = pt3.py - my;
                    if (dx * dx + dy * dy > 0.25) { // compare to the distance in pixels squared
                        refine(pt1, pt3, sign, depth + 1);
                        refine(pt3, pt2, sign, depth + 1);
                        return;
                    }
                }
            }
        }
        csctx.quadraticCurveTo(cx, cy, pt2.px, pt2.py);
    }
```

elkins0:
@gagern: Here is a version of `solveRealQuadratic` which has been reconciled to work with my version of `refine` as shown above as, replacing `solveRealQuadraticX`, well as throughout the rest of `eval_helper.drawconic`. The `sol.length > 1 && ` in my version of `refine` is no longer needed.
```javascript
function solveRealQuadratic(a, b, c) {
    var b2a = b / (2 * a);
    var u = 4 * a * c / (b * b);
    var d = 1 - u;
    /*jshint -W018 */
    if (!(d >= -CSNumber.eps)) return null; // also return null if d is NaN
    /*jshint +W018 */
    var r = d <= CSNumber.eps ? 0 : Math.sqrt(d);
    return [b2a * (-1 - r), Math.abs(u) > CSNumber.eps ? b2a * (-1 + r) : -c / b];
}
```
Also the following code in `eval_helper.drawconic`:
```javascript
        if (Math.abs(discr) > eps) {
            // Compute the roots of the x discriminant
            // for points with horizontal tangents
            sol = solveRealQuadratic(
                discr,
                4 * c01 * c20 - 2 * c10 * c11,
                4 * c00 * c20 - c10 * c10);
        } else {
            // Parabola, only a single horizontal asymptote
            sol = [
                (c10 * c10 - 4 * c00 * c20) / (4 * c01 * c20 - 2 * c10 * c11)
            ];
        }
```
can become just
```javascript
        sol = solveRealQuadratic(
            discr,
            4 * c01 * c20 - 2 * c10 * c11,
            4 * c00 * c20 - c10 * c10);
```
This code was tested on the gagern:conicDrawing2 branch code.
The `solveRealQuadratic` formulation is based on:
sqrtseries
Expanding the series to show the first few terms as a finished result:
sqrtseriesexpansion
When  4 * a * c / (b * b) is very small, only the first term -c / b is needed.

gagern:
Status update: I've had another look at this today. On the symptomatic level, one problem is that in the “almost pair of lines” situation with an “almost horizontal asymptote”, we don't get a special point for the horizontal tangent, which breaks rendering later on as the curve moves slightly outside the expected ranges of values.

The main problem however in my opinion is the various arbitrary thresholds I use different places. I think that the proper solution here would be making sure that these choices simply don't matter, visually speaking. One way to achieve this would be performing the computation in a coordinate system where the coordinate axes are far away from the asymptotic directions of the conic. That way, special cases like “(almost) horizontal / vertical“ would no longer matter. That would remove a lot of special cases: pair of lines, vertical asymptote a.k.a. `containsYFP`, parabola. So the resulting code should be simpler at the control flow level. I'm not sure how much more complicated it would get at the coordinate level.

I also had another look at the code @elkins0 proposed. One main problem here is that I don't yet understand *why* it does solve the issues it solves. So I'm unsure whther it just moves the breakage somewhere else. Is the proposed `solveRealQuadratic` which snaps the discriminant to zero really an improvement, or will it cause hyperbolas to meet when they should not? Isn't the less symmetric formulation wrt. `a` vs. `c` a possible cause for numeric instability? I think @kranich pointed this out in some personal conversation. So again instead of introducing more references to some epsilon, I'd rather try to avoid all those epsilons by moving away from special situations.

gagern:
I think I finally understood the motivation behind your termination criterion, @elkins0, and the way it helps reduce the number of segments. And while it aparently did solve the issues observed with my code, I think it is an improvement best considered independently. So I've now included the change of coordinate system I suggested, while keeping my termination criterion. Based on that we can compare termination criteria more easily, I think. The change of coordinate system indeed allowed dropping a large number of special cases, which I consider a very good thing.

It's not all good yet: here is a problematic instance of `examples/65a_Conic_3P_2l.html`:

```js
{"geometry":[{"alpha":1,"color":[1,0,0],"name":"A","size":5,"type":"Free","pos":[1,1,0.25]},{"alpha":1,"color":[1,0,0],"name":"B","size":5,"type":"Free","pos":[1,0.4188034188034188,0.2136752136752137]},{"alpha":1,"color":[1,0,0],"name":"C","size":5,"type":"Free","pos":[1,0.6666666666666666,0.3333333333333333]},{"alpha":1,"color":[1,0,0],"name":"X","size":5,"type":"Free","pos":[1,-0.3333333333333333,0.16666666666666666]},{"alpha":1,"color":[1,0,0],"name":"Y","size":5,"type":"Free","pos":[1,0.8333333333333333,0.16666666666666666]},{"alpha":1,"color":[1,0,0],"name":"Z","size":5,"type":"Free","pos":[-0.6046511627906977,1,-0.5813953488372093]},{"alpha":1,"args":["X","Y"],"clip":"none","color":[0,0,1],"name":"g","overhang":1,"size":1,"type":"Join"},{"alpha":1,"args":["X","Z"],"clip":"none","color":[0,0,1],"name":"h","overhang":1,"size":1,"type":"Join"},{"args":["A","B","C","g","h"],"name":"CoT","type":"ConicBy3p2l"},{"alpha":1,"args":["CoT"],"clip":"none","color":[1,0,0],"name":"Co1","overhang":1,"size":1,"type":"SelectConic","pos":{"xx":7.64764e-319,"yy":9.158e-320,"zz":1.5820273e-317,"xy":5.0284e-319,"xz":-7.01414e-318,"yz":-2.3777e-318}},{"alpha":1,"args":["CoT"],"clip":"none","color":[1,0,1],"name":"Co2","overhang":1,"size":1,"type":"SelectConic","pos":{"xx":5.355346e-318,"yy":1.275584e-318,"zz":5.221736e-317,"xy":-4.159983e-318,"xz":-3.4475743e-317,"yz":1.153761e-317}},{"alpha":1,"args":["CoT"],"clip":"none","color":[0,1,1],"name":"Co3","overhang":1,"size":1,"type":"SelectConic","pos":{"xx":2.450592e-317,"yy":9.158e-320,"zz":4.52906173e-316,"xy":-2.1155e-318,"xz":-2.11250217e-316,"yz":8.09566e-318}},{"alpha":1,"args":["CoT"],"clip":"none","color":[0,1,0],"name":"Co4","overhang":1,"size":1,"type":"SelectConic","pos":{"xx":4.862876e-318,"yy":1.275584e-318,"zz":1.01565356e-316,"xy":5.611785e-318,"xz":-4.484285e-317,"yz":-2.7549515e-317}}]}
```

`examples/69_Intersection_Conic_Conic.html` does flicker occasionally, too. I'll have to figure out whether this is a drawing artefact or already present in the underlying representation of the conic, and investigate further.

elkins0:
The causes of drawing failures as far as I have seen are these:
1. The input conic matrix is actually wrong due to numerical precision problems. I have seen at least one such instance which is why I keep displaying the matrix in the drawing. That however is something to be addressed elsewhere.
2. The quadratic solutions have numerical problems. The term expansion I showed above is just a red herring and not the right solution to the numerical precision problem. The way the original function solveRealQuadraticHomog deals with the `huge-sqrt(huge*huge+tiny)` numerical precision problem is almost right. The only useful change I have been able to make to `solveRealQuadraticHomog` is to deal with near imaginary numbers near zero that would otherwise result in NaNs:
```JavaScript
-    if (!(d >= 0)) return null; // also return null if d is NaN
+    if (!(d >= 0)) if (d >= -CSNumber.eps * (b * b)) d = 0; else return null; // also return null if d is NaN
```
After that the main problem is two fold:
1. The curve as it crosses the drawing area needs to be separated into continuous curves which start and end on the drawing area boundaries or in the case of ovals are fully within the drawing area. Only ovals need the special treatment of being further split into up to four pieces. Arbitrarily slicing these otherwise is not helpful. Also a provision needs to be made to allow for filling the curves. To that end this function is useful:
```JavaScript
    // True if the polarLine(point(x,y,1))'s intersections with the conic are complex.
    function pointInsideConic(x, y) {
        return ((c20 * x + c10) * x + (c02 * y + c01) * y + c11 * x * y + c00) * det > 0;
    }
```
2. The drawing of curves can then proceed by only paying attention to the starting and ending points.
A Bézier curve with control points can exactly follow a parabola and also a straight segment.
csctx.quadraticCurveTo is quite capable of drawing straight segments so no special handling is required there. If we could provide spline weights to the Bézier curve it could follow any conic section. *Mathematica* has such a capability `BSplineCurve[pts, SplineWeights -> {1, w, 1}]`. See [Rational Bézier Curves: Conic Sections](http://www.cs.mtu.edu/~shene/COURSES/cs3621/NOTES/spline/NURBS/RB-conics.html) for detail on the mathematics involved. Since JavaScript has no such capability, separate Bézier curves are instead used to approximate the conic section. To facilitate filling up to four corners of the drawing area need to be included in the path.

gagern:
Thanks @elkins0 for your analysis.

> Arbitrarily slicing these otherwise is not helpful.

The main reason I had for slicing my conics was to help me correctly connect the points of intersection I found. That's what happens as I iterate over `specialPoints` and set the `prev` and `next` links along the way. The assumption of monotonic coordinate changes is also used in the `inRange` checks inside `refine`, which you kept in your version of `refine`. If you have a simpler suggestion on how to achieve this, I'd be very interested to hear about it.

Let me think aloud for a moment. Suppose I have only the intersection with the boundary of the canvas, how can I tell which of these boundary points to connect? We have up to 8 points of intersection with the boundary, as its 4 lines and each may intersect the conic in 2 points. Disregarding the case of coniciding points, the total number of intersections with the bounding rectangle has to be even. Each boundary point should get connected to one adjacent point. So there are essentially two possible patterns: one where point 2<em>k</em> connects to 2<em>k</em> + 1 and one where 2<em>k</em> connects to 2<em>k</em> − 1.

This approach models a pair of crossing lines as a degenerate hyperbola, with one arc spanning one part of each line, and the other arc spanning the other part of them. I think this is more robust against poorly chosen thresholds than actually having straight lines crossing. The choice which halves to consider one arc is arbitrary here. But in an *almost* degnerate hyperbola, the choice matters and is very hard to make by looking at the boundary crossings only. At least in this case, it might be best to look at the center of the hyperbola. That center is always “outside” for a hyperbola. And boundary points should be connected in such a way that the boundary between the connected points is “inside”.

This is not neccessarily true if the center of the hyperbola lies outside the bounding box. And it doesn't translate well to other conics, particularly to parabolas with their center at infinity *on* the conic. So I guess the center isn't a sufficiently generic concept here. Is there any better rule to tell us whether we want to connect points in such a way that the boundary between them is “outside” or “inside”?

> … deal with near imaginary numbers near zero that would otherwise result in NaNs

Can we actually get NaNs here? As I check that `d >= 0` I would assume that `Math.sqrt(d)` will *always* be a well-defined and non-negative number. Possibly positive infinity, if `d` itself were infinite.

I realized that a slightly negative discriminant in this quadratic equation solving might be part of the cause for some of the flickering I observed: if I have a pair of lines, the points where the horizontal or vertical tangents are will all coincide, leading to such solutions. But I would rather try make my code sufficiently robust that this choice does not matter, that it will draw any pair of hyperbola arcs nonetheless.

> Only ovals need the special treatment of being further split into up to four pieces.

Do you think this also applies to ellipses which do already intersect the boundary? Because if not, then in my opinion the end points by themselves would not be enough to distinguish which part of the ellipse is to be drawn. Perhaps it is enough to split those ellipses whose center is inside the bounding box.

On the whole I wouldn't mind changing all of this to something which does fewer splits, but I'm unsure how to do that without too many case distinctions.

gagern:
@elkins0: I've implemented some of the ideas you suggested above, and also taken some of the code you pasted here, and wrote #555 from this. As I'm not 100% sure that approach will be more robust than this one here, I'd like to keep both of them open for comparison. But so far I think things look pretty good, except for one problem for which I'm still looking for a solution. See there.

elkins0:
@gagern: Here is an implementation of the `drawConicArk` and `refine` functions for this version of your code.

Note that the `c##` are shown here unchanged so that the placement of the added `mat = List.normalizeMax(mat);` is shown. I already had this working before #555 appeared :smile: See #555 for further details.

```JavaScript
    mat = List.normalizeMax(mat);

    // Using polynomial coefficients instead of matrix
    // since it generalizes to higher degrees more easily.
    // cij is the coefficient of the monomial x^i * y^j.
    var c20 = mat.value[0].value[0].value.real;
    var c11 = mat.value[0].value[1].value.real * 2;
    var c10 = mat.value[0].value[2].value.real * 2;
    var c02 = mat.value[1].value[1].value.real;
    var c01 = mat.value[1].value[2].value.real * 2;
    var c00 = mat.value[2].value[2].value.real;
    //-var discr = c20 * c02 * 4 - c11 * c11;

    // The adjoint matrix k## values
    var k20 = 4 * c00 * c02 - c01 * c01;
    var k11 = c01 * c10 - 2 * c00 * c11;
    var k10 = c01 * c11 - 2 * c02 * c10;
    var k02 = 4 * c00 * c20 - c10 * c10;
    var k01 = c10 * c11 - 2 * c01 * c20;
    var k00 = 4 * c02 * c20 - c11 * c11; // discr -4(b * b - a * c)
    var discr = k00;
    var det = c02 * k02 + c11 * k11 + c20 * k20 - c00 * k00;
    // conic center [ccx / ccz, ccy / ccz]
    var ccx = k10 / k00;
    var ccy = k01 / k00;

    // Draw a conic arc from A to C, with control point B & a weight of k/(1-k)
    function drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, n) {
        var j = 1 - k;
        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);
        var Fx = 0.5 * (Bx + Mx);
        var Fy = 0.5 * (By + My);
        var kBx = k * Bx;
        var kBy = k * By;
        var Gx = kBx + j * Mx;
        var Gy = kBy + j * My;
        if ((Fx - Gx) * (Fx - Gx) + (Fy - Gy) * (Fy - Gy) < 0.25 || n > 10) {
            csctx.quadraticCurveTo(Bx, By, Cx, Cy);
        } else {
            var kr = 1 / (Math.sqrt(2 * j) + 1);
            ++n;
            drawConicArk(Ax, Ay, kBx + j * Ax, kBy + j * Ay, Gx, Gy, kr, n);
            drawConicArk(Gx, Gy, kBx + j * Cx, kBy + j * Cy, Cx, Cy, kr, n);
        }
    }

    // Find the control points of a quadratic Bézier which at the
    // endpoints agrees with the conic in position and tangent direction.
    // sign is -1 for the lower branch and +1 for the upper.
    function refine(pt1, pt2, sign, depth) {
        var Ax = pt1.px;
        var Ay = pt1.py;
        var Cx = pt2.px;
        var Cy = pt2.py;

        var ux = Ay - Cy;
        var uy = Cx - Ax;
        var uz = Ax * Cy - Cx * Ay;

        //pole = {ux, uy, uz}.{{k20, k11, k10}, {k11, k02, k01}, {k10, k01, k00}};
        var Bz = (k10 * ux + k01 * uy + k00 * uz);
        if (Math.abs(Bz) < CSNumber.eps) {
            return csctx.lineTo(
                dirCos * Cx - dirSin * Cy,
                dirCos * Cy + dirSin * Cx);
        }
        var Bx = (k20 * ux + k11 * uy + k10 * uz) / Bz;
        var By = (k11 * ux + k02 * uy + k01 * uz) / Bz;

        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);
        // Note: The line thru M and B also passes thru the conic center [ccx/ccz,ccy/ccz]
        var rdx = Bx - Mx;
        var rdy = By - My;

        var k = 0.5;
        if (k00 !== 0) {
            var r = Math.sqrt(-2 * det / (
                My * k00 * (c01 + c11 * Mx + 2 * c02 * My) +
                Mx * k00 * (c10 + 2 * c20 * Mx + c11 * My) -
                (k10 - Mx * k00) * c10 - (k01 - My * k00) * c01));
            if (Math.abs(rdy) < Math.abs(rdx)) {
                k = ((ccx - Mx) * (1 - r) / rdx);
            } else {
                k = ((ccy - My) * (1 - r) / rdy);
            }
        }

        // De-rotate
        var tt;
        tt = dirCos * Ax - dirSin * Ay;
        Ay = dirCos * Ay + dirSin * Ax;
        Ax = tt;
        tt = dirCos * Bx - dirSin * By;
        By = dirCos * By + dirSin * Bx,
        Bx = tt;
        tt = dirCos * Cx - dirSin * Cy;
        Cy = dirCos * Cy + dirSin * Cx;
        Cx = tt;
        if (!(k >= 0 && k <= 1)) {
            console.log("out-of-range: k="+k+" r="+r+" det="+det+" ccx="+ccx+" ccy="+ccy);
            k = 0.5;
        }
        drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, 0);
    }
```

elkins0:
In terms of adjoint matrix `k##` values these changes are possible:
```JavaScript
         // Compute the roots of the y discriminant
         // for points with vertical tangents
-        sol = solveRealQuadratic(
-            discr,
-            4 * c10 * c02 - 2 * c01 * c11,
-            4 * c00 * c02 - c01 * c01);
+        sol = solveRealQuadratic(k00, -2 * k10, k20);
```
and
```JavaScript
         // Compute the roots of the x discriminant
         // for points with horizontal tangents
-        sol = solveRealQuadratic(
-            discr,
-            4 * c01 * c20 - 2 * c10 * c11,
-            4 * c00 * c20 - c10 * c10);
+        sol = solveRealQuadratic(k00, -2 * k01, k02);
```

gagern:
Superseded by #555 or something along those lines, I'd say.

PR #281, opened 2016-02-29
Handle per-element dimDependent settings

gagern:
When switching an element to a custom color in Cinderella, by a scripted assignment of the form `el.color = [r, g, b]`, the dimming of dependent points won't apply to such an element any more.  So we need to exclude certain elements from the dimming rule.  I've already posted a Cinderella merge request for matching export code.

This is a follow-up to #231.

### DIFF
--- a/src/js/libgeo/GeoBasics.js
+++ b/src/js/libgeo/GeoBasics.js
@@ -37,6 +37,7 @@ function csinit(gslp) {
 
     //Relevant fields for appearance:
     //color
+    //dimDependent
     //size
     //alpha
     //overhang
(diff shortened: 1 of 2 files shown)


kranich:
@gagern I think what is still missing in your PR is code to set `el.dimDependent = 1.0` upon `el.color` assignment.

gagern:
@kranich I'm not sure. The `pointDefault` already includes the `dimDependent` setting into the color when loading the construction. Further modifications shouldn't be affected by this. Things may look differently when saving the state. But I just realized that we have a problem there anyway: we save the already dimmed color, but when loading that state again, we will dim a second time. So we probably should indeed set `el.dimDependent = 1` upon assignment, but mainly so that we will save that with the state, and can undo the dimming when saving the colors. Will update my branch soon.

kranich:
>Will update my branch soon.

@gagern Please let me know when you have updated the branch.

strobelm:

PR #282, opened 2016-03-01
Allow setting homogeneous coordinates of FreeLine

kranich:


strobelm:
We should add tracing to this @kranich promised to create at least a bug report

PR #289, opened 2016-03-02
Fix drawsegcore

kranich:
This fixes spurious line rendering bugs in Firefox (see also https://bugzilla.mozilla.org/show_bug.cgi?id=1183702).

I think we once had an issue for that, though I cannot currently find it.

This is also related to #56, and maybe also to #97.

I have split the change into two commits to make it easier to understand.

Issue #97, opened 2015-11-24
Join rendered incorrect in degenerate situation


gagern:
In the following example, making the two points coincide has a good chance of rendering the line at an incorrect position, i.e. a position where it passes through neither.

```js
createCindy({
	geometry: [
		{name: "A", type: "Free", pos: [5,1]},
		{name: "B", type: "Free", pos: [5,-1]},
		{name: "a", type: "Join", args: ["A", "B"]}
	],
	ports: [{ id: "CSCanvas", width: 600, height: 400 }]
});
```

It would be better to detect such undefined situations and not render the elements in question. This seems related to #57.

PR #292, opened 2016-03-03
Do not draw lines beyond boundary of canvas

kranich:
This is a follow up of #289. I realized that it is actually `Render2D.drawline` that needs be fixed, not `Render2D.drawsegcore`.

This fixes line rendering bugs in Firefox (see also https://bugzilla.mozilla.org/show_bug.cgi?id=1183702 and [CindyJS GitLab issue #6](https://gitlab.cinderella.de:8082/cindyjs/cindyjs/issues/6)).

This is also related to #56, and maybe also to #97.

### DIFF
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -523,9 +523,9 @@ Render2D.drawline = function(homog) {
     // do not draw beyond canvas boundary
     var m = csport.drawingstate.matrix;
     var b1 = [m.a, m.b, m.tx]; // left boundary
-    var b2 = [m.a, m.b, m.tx - csctx.canvas.width]; // right boundary
+    var b2 = [m.a, m.b, m.tx - csw]; // right boundary
     var b3 = [m.c, m.d, m.ty]; // upper boundary
-    var b4 = [m.c, m.d, csctx.canvas.height - m.ty]; // lower boundary
+    var b4 = [m.c, m.d, csh - m.ty]; // lower boundary
     if (l[0] === 0) { // horizontal line
         Render2D.drawsegcore(crossxy(l, b1), crossxy(l, b2));
         return;
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -547,16 +547,18 @@ Render2D.drawline = function(homog) {
         y: (-c - a * xMin) / b
     });
 
-    var toUserSpace = function (p) {
-        var q = List.productMV(mInverse, List.realVector([p.x, p.y, 1]));
-        return {
-            x: q.value[0].value.real / q.value[2].value.real,
-            y: q.value[1].value.real / q.value[2].value.real
-        };
-    };
+    if (erg.length === 2 && Render2D.lsize >= 0.01) {
+        csctx.lineWidth = Render2D.lsize;
+        csctx.lineCap = Render2D.lineCap;
+        csctx.lineJoin = Render2D.lineJoin;
+        csctx.miterLimit = Render2D.miterLimit;
+        csctx.strokeStyle = Render2D.lineColor;
 
-    if (erg.length === 2)
-        Render2D.drawsegcore(toUserSpace(erg[0]), toUserSpace(erg[1]));
+        csctx.beginPath();
+        csctx.moveTo(erg[0].x, erg[0].y);
+        csctx.lineTo(erg[1].x, erg[1].y);
+        csctx.stroke();
+    }
 };
 
 Render2D.dashTypes = {
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -500,8 +500,11 @@ Render2D.drawline = function(homog) {
 
     // transformation to canvas coordinates
     var m = csport.drawingstate.matrix;
-    var mInverse = List.inverse(List.realMatrix([[m.a, -m.b, m.tx],
-        [m.c, -m.d, -m.ty], [0, 0, 1]]));
+    var mInverse = List.inverse(List.realMatrix([
+        [m.a, -m.b, m.tx],
+        [m.c, -m.d, -m.ty],
+        [0, 0, 1]
+    ]));
     var n = List.normalizeMax(List.productMV(List.transpose(mInverse), homog));
     var a = n.value[0].value.real;
     var b = n.value[1].value.real;
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -505,11 +505,11 @@ Render2D.drawline = function(homog) {
     var l = List.normalizeMax(homog);
     l = [l.value[0].value.real, l.value[1].value.real, l.value[2].value.real];
 
-    function crossxy(u, v) {
-        var zInverse = 1 / (u[0] * v[1] - u[1] * v[0]);
+    function lcrossxy(v) {
+        var zInverse = 1 / (l[0] * v[1] - l[1] * v[0]);
         return {
-            x: (u[1] * v[2] - u[2] * v[1]) * zInverse,
-            y: (u[2] * v[0] - u[0] * v[2]) * zInverse
+            x: (l[1] * v[2] - l[2] * v[1]) * zInverse,
+            y: (l[2] * v[0] - l[0] * v[2]) * zInverse
         };
     }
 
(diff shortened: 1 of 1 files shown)


kranich:
The second and third commit are not directly related to not drawing beyond canvas boundary. They are general improvements of `Render2D.drawline`.

kranich:
@gagern I think this is ready for review now.

src/js/libcs/Render2D.js
```diff
@@ -500,59 +500,41 @@ Render2D.drawpoint = function(pt) {
...
-    } else {
-        b1 = [0, 1, 30];
-        b2 = [0, -1, 30];
+
+    // do not draw beyond canvas boundary
+    var m = csport.drawingstate.matrix;
+    var b1 = [m.a, m.b, m.tx]; // left boundary
+    var b2 = [m.a, m.b, m.tx - csctx.canvas.width]; // right boundary
+    var b3 = [m.c, m.d, m.ty]; // upper boundary
+    var b4 = [m.c, m.d, csctx.canvas.height - m.ty]; // lower boundary
```
gagern: I suggest you use `csw` and `csh` here. Rationale: we may one day offer other backends, like SVG, in which case `csctx.canvas` would not be defined the way it is now.

src/js/libcs/Render2D.js
```diff
@@ -500,59 +500,41 @@ Render2D.drawpoint = function(pt) {
...
+    if (!List._helper.isAlmostReal(homog))
+        return;
+    var l = List.normalizeMax(homog);
+    l = [l.value[0].value.real, l.value[1].value.real, l.value[2].value.real];
+
+    function crossxy(u, v) {
+        var z = (u[0] * v[1] - u[1] * v[0]);
+        return {
+            x: (u[1] * v[2] - u[2] * v[1]) / z,
+            y: (u[2] * v[0] - u[0] * v[2]) / z
```
gagern: Since multiplication is faster than division, we could consider computing `1/z` and then scaling with that. Not sure it's worth the resulting loss in readability, though.

src/js/libcs/Render2D.js
```diff
@@ -500,59 +500,41 @@ Render2D.drawpoint = function(pt) {
...
-        y: erg1[1] / erg1[2]
-    };
-    var pt2 = {
-        x: erg2[0] / erg2[2],
-        y: erg2[1] / erg2[2]
-
-    };
-
-    Render2D.drawsegcore(pt1, pt2);
+    var intersection = [b1, b2, b3, b4].map(crossxy.bind(null, l)).sort(sortxy);
```
gagern: As far as I can see, all calls to `crossxy` involve `l` as one of the arguments. And since `crossxy` is in the same scope as `l`, you could just let that function refer to `l` and a single other argument, and then avoid the binding.

gagern: I'm slightly worried about the `sortxy` here: an almost vertical line could conceivably lead to points with exactly the same *x* coordinate but different *y* coordinates, right? So in that case you'd have a strange and inconsistent mix between “sort by *x*” and “sort by *y*”. Come to think of it, an almost vertical line could also lead to infinite *y* values, right? I'm wondering whether #103, adjusted to take screen bounds into account, would be a better way to deal with this. As far as I remember it avoids special cases for horizontal or vertical. What do you think?

kranich: >I'm slightly worried about the sortxy here: an almost vertical line could conceivably lead to points with exactly the same x coordinate but different y coordinates, right? So in that case you'd have a strange and inconsistent mix between “sort by x” and “sort by y”. Come to think of it, an almost vertical line could also lead to infinite y values, right? I'm wondering whether #103, adjusted to take screen bounds into account, would be a better way to deal with this. As far as I remember it avoids special cases for horizontal or vertical. What do you think?

I could not manually produce any misrendered lines using my approach. But if you prefer #103, feel free to use that or cherry-pick.

kranich:
I will soon unify this with #103, as discussed in today's meeting.

kranich:
@gagern Please review my attempt at unification with #103.

src/js/libcs/Render2D.js
```diff
@@ -500,59 +500,49 @@ Render2D.drawpoint = function(pt) {
...
+        return x * a + y * b + c < 0;
     };
 
-    Render2D.drawsegcore(pt1, pt2);
+    // do not draw beyond canvas boundary
+    var m = csport.drawingstate.matrix;
+    var xMin = (m.b * m.ty - m.d * m.tx) / m.det;
+    var xMax = (m.b * (m.ty - csh) - m.d * (m.tx - csw)) / m.det;
+    var yMax = (m.c * m.tx - m.a * m.ty) / m.det;
+    var yMin = (m.c * (m.tx - csw) - m.a * (m.ty + csh)) / m.det;
```
gagern: I'mm slighly worried about the situation where a `rotate` call rotated the coordinate system to something no longer axis-aligned. In that case, delimiting by constant *x* resp. *y*  coordinate would no longer be appropriate. Perhaps we should transform the line itself to canvas coordinates, instead of transforming the canvas boundaries back into user space?

Or, as an alternative, we should at least feed canvas coordinates to `distNeg` and convert them to user space before plugging them into the line equation. That way we'd have twice as many conversions (two independent coordinates for every corner), and computing the actual intersections would be a bit more tricky as well, so I think using canvas coordinates should be easier all around.

Maybe we should have a test case for that scenario.

kranich: >I'mm slighly worried about the situation where a rotate call rotated the coordinate system to something no longer axis-aligned.

@gagern You are not alone with your worries! ;-)

>Perhaps we should transform the line itself to canvas coordinates, instead of transforming the canvas boundaries back into user space?

@gagern Yes, I think that would be sensible. But in the end we need to transform the arguments of `Render2D.drawsegcore` back to user space. I will give that a try.

kranich: >Maybe we should have a test case for that scenario.

@gagern Can you think of a way to test this? We could start with the test case included in #103. Then we would at least see that the lines extend at least to the canvas boundary. But how do we ensure that we do not draw beyond the canvas boundary?

kortenkamp: I wonder why you don't clip by canvas boundaries when drawing. It is impossible to determine the correct boundaries in advance, and when you have thick lines it will be impossible to draw correctly without clipping anyway.

kranich: @kortenkamp Because the whole drawing process is a mess! ;-) `Render2D.drawline` uses `Render2D.drawsegcore`, i.e. lines are drawn as segments whose end points we need determine. We do not want these to lie in the interior of the canvas, since lines should extend to the boundary of the canvas. On the other hand, we should of course also clip segments. So we will have some duplicate code in any case, as far as I see it.

kranich:
@gagern, @kortenkamp Maybe it makes sense to sort out the optimal approach in the next CindyJS meeting?

src/js/libcs/Render2D.js
gagern: > But in the end we need to transform the arguments of `Render2D.drawsegcore` back to user space.

I see no reason why the arguments to `drawsegcore` *have* to be user space coordinates. We could make them canvas space, and let the caller of `drawsegcore` handle conversion. Or move the bulk of `drawsegcore` to some new function, and let `drawsegcore` do the conversion and then call the new function.

> On the other hand, we should of course also clip segments. So we will have some duplicate code in any case, as far as I see it.

When doing segment clipping, we have to be particularly aware of the fact that a segment between two (almost) identical points should be supported, even though the line connecting those is not uniquely defined. So I'd probably build segment clipping on the notion of whether the endpoints of a segment are separated by a given canvas border. So that you don't compute points of intersection unneccessarily. I'd say this probably should go into some function that calls `drawsegcore`, not into `drawsegcore` itself as it is now. If you make those two functions mentioned above, perhaps the outer one of them is the right place for segment clipping?

Not sure what to do about rays, though. Do we need a third code path, or does it make sense to somehow treat them as a special case of lines or segments? We will encounter rays e.g. in the case of segments that pass through infinity.

kranich:
In today's CindyJS meeting, we agreed that

1. line and segment drawing should be separated,
2. segments must also be clipped, and
3. clipping must take line width into account.

I will prepare the next iteration.

kranich:
@gagern Can you please explain the meaning of overhang for lines to me? Doesn't this only make sense for `Join`? If so, then it is probably broken, currently.

kranich:
@gagern I have separated line drawing from segment drawing and have taken line width into account for line clipping. Please review. I will open another issue for segment clipping.

kranich:
Sort of superseded by #343.

(#343 includes this pull request, so merging #343 makes #292 obsolete. However, it might be desirable to sort things out for #292 first, then merge #292 and rebase #343 on #292).

PR #294, opened 2016-03-08
all operations and guess more incidences take 2

gagern:
This supersedes #209.

@strobelm, I'm happy with what you did in `guessIncidences`, but had serious concerns about the operator implementations, so I rewote them for the most part. Please have a look at my code there, and give it a review.


strobelm:
Hi @gagern thanks for the improvements, the code looks much clearer now.

PR #295, opened 2016-03-08
Formulate incidence checks as conjectures

gagern:
This builds on #294, so either review that first before dealing with this here, or review this one only since merging it will close that one as well.

This lays the groundwork for randomized theorem proving, since it allows us (in a future commit) to check those conjectures before turning them into actual incidences. One benefit already obtained from this change in implementation is that dynamically added elements will be checked for incidences, too.

### DIFF
--- a/examples/114_allops.html
+++ b/examples/114_allops.html
@@ -1,0 +1,126 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Konstruktion.cdy</title>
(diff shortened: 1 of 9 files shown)


strobelm:
Hi, i checked the rebased version of it. Looks also fine to me. 

I'm really eager to start implementing the randomized proving! I'll open an issue to coordinate the efforts.

gagern:
Thanks for the lightning-quick review, @strobelm!

PR #297, opened 2016-03-08
element$1 operator and some drive-by fixes

gagern:
This adds the `element(‹string›)` operator. Writing the testcase for this, I found that `==` failed to handle geometric objects, and incorrectly turned points into their `xy` coordinates. And thinking about corner cases, I found errors using JavaScript objects as dictionaries of geometric elements or variables, since inherited properties would cause confusion. That last change will neccessitate a rebase of #257 but that's currently in conflict anyway.


strobelm:
Looks good to me, thanks @gagern.

PR #299, opened 2016-03-11
randomized proving

strobelm:
Hi, i think the randomized prover is ready to be reviewed. 

I implemented random moves (in a small environment) of all geoOps which are movable. There are example files which are named 141_\* where you can see actually nothing, which is not a bad thing ;-). If you enable debugging in Prover.js checkConjectures() you can observe the points moving. Only real moves are currently enabled so you can actually see the points moving. I will enable complex movements as soon as the review is done and no severe bugs are found. 

There are also unit tests implemented since the example files are not really useful when debugging is disabled.

Currently drawing is not disabled while proving, i would really like input on that. I suppose it's enough to disable all calls despite traceMouseAndScript in updateCindy but i'm not sure.


gagern:
Travis had some concerns, but I fixed this in 6ae82ef2a3499d2cb8a81294d2d78686f6ca39ce. Without going through the usual code review mechanism, since I think this uncontroversial and urgently needed. I then triggered a re-run, and now Travis is happy, too. Will have a closer look at the code next week at the latest.

strobelm:
On our last meeting we had some points which could be improved, i'll notice you as soon these changes are done.

strobelm:
Hi,

here's the update: only move free elements which are involved in conjectures and watch out for failed traces. I think this is ready for review again.

src/js/libgeo/GeoOps.js
```diff
@@ -371,6 +444,22 @@ geoOps.Free.updatePosition = function(el) {
...
 };
+geoOps.Free.getRandomMove = function(el) {
+    var rcomplex = function() {
+        var fact = 0.1;
+        // debug only real remove later!
+        return CSNumber.complex((Math.random() - 0.5) * fact, (Math.random() - 0.5) * 0);
+    };
+    var oldpos = List.normalizeMax(el.homog);
+    var rvect = List.turnIntoCSList([rcomplex(), rcomplex(), CSNumber.real(0)]);
+    var move = List.scalmult(oldpos.value[2], rvect);
```
gagern: As discussed, this kind of movement will preserve points at infinity. What's even worse, since you scale by `oldpos.z`, points at infinity won't move at all. That can easily prevent us from disproving some spurious conjectures.

The whole setup puts a bit too much emphasis on the choice of coordinates, for my taste. A user using a coordinate system with thousand units per pixel may experience different effects compared to one using a thousand pixels per unit. Since I'm still advocating large movements, it's the former case that has me worried: Could it happen that at a very low zoom setting, some configuration just doesn't move enough to actually disprove invalid conjectures? Can we estimate the chances of this, for some simple conjectures (i.e. without too much of an artificial dampening construction)? I don't see this aspect as a showstopper, but a personal concern nonetheless.

src/js/libgeo/GeoOps.js
```diff
@@ -709,6 +801,20 @@ geoOps.CircleMr.updatePosition = function(el) {
     el.matrix = General.withUsage(matrix, "Circle");
     el.radius = r;
 };
+geoOps.CircleMr.getRandomMove = function(el) {
+    // radius 
+    var r;
+    var oldr = el.radius;
+    do {
+        r = CSNumber.add(oldr, CSNumber.real(0.1 * (Math.random() - 0.5)));
+    } while (r.value.real <= 0); // make sure r > 0;
```
gagern: How about scaling the radius by a random factor instead, to avoid both a change in sign and a complete change of magnitude? Might break for circles of radius zero, though.

Do we even have a problem with circles of negative radius? I think there *should* be no problems, but I'm not sure. Do we allow for complex radii eventually?

src/js/libgeo/Prover.js
```diff
@@ -45,6 +45,9 @@ function applyIncidence(a, b) {
 
 function incidentPL(p, l) {
     return {
+        getInvolved: function() {
+            return [p, l];
+        },
```
gagern: If I didn't know better, I'd read the name here as an invitation for possible future contributors: *Get involved with the CindyJS project!* :smile:

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,68 @@ function incidentPC(p, c) {
...
+
+    var ii, jj, emove;
+    var nummoves = 3;
+
+    // filter free objects which are involved in conjectures
+    var involved = [];
+    conjectures.forEach(function(con) {
+        var invs = con.getInvolved();
+        invs = invs.forEach(function(el) {
+            if (involved.indexOf(el) < 0 && csgeo.free.indexOf(el) >= 0) involved.push(el);
```
gagern: `el.movable` should be a shorter substitute for `csgeo.free.indexOf(el) >= 0`.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,68 @@ function incidentPC(p, c) {
...
+        });
+    });
+
+    // debug code remove later
+    var nconject = conjectures.length;
+    involved.forEach(function(el) {
+        ii = nummoves;
+        while (ii--) {
```
gagern: I don't have a good reason for this just now, but my gut feeling would be that it might be better to exchange the order of loops here: perform one move for each of the `involved` elements, and then a second one, and so on up to `nummoves` for each of them. That's what Cinderella does, according to @kortenkamp, and somehow that feels right to me. You *could* consider re-building the list of `involved` elements after each pass, particularly if you managed to disprove some conjecture. Not sure about the performance tradeoff here; I guess it depends on the application.

tests/Prover_tests.js
```diff
@@ -0,0 +1,212 @@
...
+    "lineTo",
+    "moveTo",
+    "restore",
+    "save",
+    "stroke",
+    "strokeText",
+    "fillText",
+].forEach(function(m) {
+    FakeCanvas.prototype[m] = dummy;
+});
```
gagern: I think we should check whether we actually need a canvas, and if so move this fake canvas to a module of its own to be shared between tests. But IIRC I managed to get some tests running without a canvas at all, thanks to the `isNode` setting.

src/js/libgeo/Prover.js
strobelm: Get involved and receive a pair of geometric objects for free! ;-)

Thanks for the annotations Martin, i'll look into it.

Am 01.04.2016 um 00:12 schrieb Martin von Gagern:
>
> In src/js/libgeo/Prover.js
> <https://github.com/CindyJS/CindyJS/pull/299#discussion_r58136437>:
>
> > @@ -45,6 +45,9 @@ function applyIncidence(a, b) {
> >  
> >  function incidentPL(p, l) {
> >      return {
> > +        getInvolved: function() {
> > +            return [p, l];
> > +        },
>
> If I didn't know better, I'd read the name here as an invitation for
> possible future contributors: /Get involved with the CindyJS project!/
> :smile:
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/CindyJS/CindyJS/pull/299/files/13207c592c99c82ab73a5bdda83d9586b1a9e53d#r58136437>
>

-- 
Michael Strobel, M.Sc.

strobelm:
I incorporated the changes accordingly. The loops were exchanged, radius is scaled (and also a check for small radii is done) and far points should now be fine. Concerning the test cases i'm not sure if there is a need to change them or not.

gagern:
I haven't looked at the test cases in detail yet. Are there some testing actual theorems, like Pappos' or Pascal's? It might make sense to have these, just in case we one day decide to do something more efficient about incidences which are obvious from the definitions of the objects involved. There is no surprise that a `Join` is incident to both its defining points, stuff like that.

kortenkamp:
By the way, Cinderella (Classic) has the concept of "expected theorems" (Conjectures that are expected to hold) for the cases you just described, @gagern

strobelm:
Emphasis within the tests lies on dropping conjectures, usually i start
with an coincidental incident situation and test if the we can prove
that it was coincidental.

I'll add some more advanced theorems.   

Am 01.04.2016 um 13:20 schrieb Martin von Gagern:
>
> I haven't looked at the test cases in detail yet. Are there some
> testing actual theorems, like Pappos' or Pascal's? It might make sense
> to have these, just in case we one day decide to do something more
> efficient about incidences which are obvious from the definitions of
> the objects involved. There is no surprise that a |Join| is incident
> to both its defining points, stuff like that.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/CindyJS/CindyJS/pull/299#issuecomment-204353243>
>

-- 
Michael Strobel, M.Sc.

strobelm:
Added Pappo's theorem and we already had the angle bisector theorem as test case.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,77 @@ function incidentPC(p, c) {
...
+    for (var kk = 0; kk < nummoves; kk++) {
+        for (var oo = 0; oo < involved.length; oo++) {
+            var el = involved[oo];
+            if (!el.pinned) {
+                if (debug) console.log("prover: moving element", el.name);
+                // get random move and move free element
+                emove = geoOps[el.type].getRandomMove(el);
+                movepointscr(el, emove.value, emove.type);
+                // if something bad happens
+                if (tracingFailed) stateIn.set(stateLastGood);
```
gagern: This looks broken. In general I'd assume that the prover operates while the mouse is not being used. So `movepointscr` would not use its `inMouseMove` branch. As a consequence it would call `stateContinueFromHere`, which does `stateLastGood.set(stateIn)`. Therefore the `stateIn.set(stateLastGood)` call is without effect. If the randomized prover were to enter a singularity, it would not leave it again. I wonder how we can properly test this situation.

src/js/libgeo/GeoOps.js
```diff
@@ -59,6 +59,26 @@ geoOps.FreeLine.updatePosition = function(el) {
...
+
+    var move = List.cross(perp, l);
+    move = List.normalizeMax(move);
+
+    return {
+        type: "mouse",
+        value: move
+    };
+
+};
```
gagern: I'm trying to understand this, and I think that some more verbose names and/or comments might help. `l` is the line as it is. `tt` is the point at infinity orthogonal to it. `p` is some random point near the origin. Now you connect `tt` and `p` to obtain a line `perp` which is perpendicular to the original `l`. Next you intersect `perp` with `l` and treat that intersection like a mouse move. The goal here is finding a small but definitely non-zero change, right? So the fact that you intersect with the line will ensure that the resulting point won't be too far away from the line. And the fact that the line `perp` passes close to the origin will make sure that the resulting point is again close to the origin, and thus to the point where the mouse would define the line in question. Did I understand this correctly?

If `l` is the line at infinity, then `tt`, `perp` and `move` will be zero. While this couldn't happen using the current mouse user interface, it could well happen using scripted movement, or perhaps some future input methods.

src/js/libgeo/GeoOps.js
```diff
@@ -215,6 +234,25 @@ geoOps.HorizontalLine.updatePosition = function(el) {
...
+    var rvec = List.turnIntoCSList([CSNumber.zero, rcomplex(), rcomplex()]);
+    rvec = List.scalmult(p.value[2], rvec);
+
+    var move = List.add(rvec, p);
+    var res = {
+        type: "mouse",
+        value: move
+    };
+    return res;
+};
```
gagern: This should get the job done, but I can't help wondering whether there is a simpler approach using `"homog"` instead of `"mouse"` as the movement type. Can't we just add `[0, rcomplex(), rcomplex()]` to `el.homog`? If not, under what circumstances would this be problematic?

The code has trouble if the line is the line at infinity, e.g. after an explicit assignment to the `homog` field. Then `p.value[2]` is zero, so `rvec` is null. Furthermore `move` equals `List.ex` which makes the resulting line undefined.

Both comments above apply to `VerticalLine` as well.

src/js/libgeo/GeoOps.js
```diff
@@ -371,6 +441,31 @@ geoOps.Free.updatePosition = function(el) {
     putStateComplexVector(param); // copy param
     el.homog = General.withUsage(param, "Point");
 };
+geoOps.Free.getRandomMove = function(el) {
+    var rcomplex = function() {
+        var fact = 0.1;
+        return CSNumber.complex((Math.random() - 0.5) * fact, (Math.random() - 0.5) * fact);
+    };
+    var oldpos = List.normalizeMax(el.homog);
```
gagern: This should be normalized already.

src/js/libgeo/GeoOps.js
```diff
@@ -371,6 +441,31 @@ geoOps.Free.updatePosition = function(el) {
...
+    var oz = oldpos.value[2];
+    var ozabs = CSNumber.abs(oz).value.real;
```
gagern: I'm not sure whether we should prefer `abs2` over `abs` in order to avoid the square root. From a performance standpoint this would make sense, since it's easy enough to adjust the threshold accordingly. But from a maintainance standpoint the version with `abs` may be easier to understand. And the cost of `sqrt` likely isn't excessive, compared to all the other things we do.

src/js/libgeo/GeoOps.js
```diff
@@ -371,6 +441,31 @@ geoOps.Free.updatePosition = function(el) {
...
+
+    var rvect = List.turnIntoCSList([rcomplex(), rcomplex(), CSNumber.real(0)]);
+
+    // far points 
+    if (ozabs < CSNumber.eps) {
+        oz = CSNumber.real(Math.random());
+        rvect.value[2] = CSNumber.real(Math.random());
```
gagern: This is “slightly evil”, since it modifies a CindyJS data structure and those are considered immutable after construction, in order to make copy-on-write work reliably. In this particular case things are safe, since we know that noone else has a copy of this `rvec.value` array. But nevertheless it would be cleaner to change the order of things: initialize some variable to zero, then change it to some random number in this `if` here, and only then construct the `rvec` taking that variable as the last coordinate.

What's the point in randomizing `oz` as well, as opposed to setting it to 1? If there is a good reason, a comment explaining it might be in order.

src/js/libgeo/GeoOps.js
```diff
@@ -709,6 +807,22 @@ geoOps.CircleMr.updatePosition = function(el) {
...
     el.radius = r;
 };
+geoOps.CircleMr.getRandomMove = function(el) {
+    // radius 
+    var r;
+    var oldr = el.radius;
+    var oabs = CSNumber.abs(oldr).value.real;
+
+    // if radius was small we want something larger and if not we scale the old one
+    r = oabs < CSNumber.eps ? CSNumber.real(Math.random() + 0.1) : CSNumber.mult(oldr, CSNumber.real(Math.random() + 0.5));
```
gagern: I think that a regular `if … else …` might be easier to read here, since the line is pretty long already.
This code chooses only real radii, where most other randomization is complex.
It also performs pretty large changes where most other randomizations tend to do small steps only.
Are there any particular reasons for these?
(Personally I'm all for making large changes, but there were concerns that these would make tracing take too long.)

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,77 @@ function incidentPC(p, c) {
...
 
 function checkConjectures() {
+    var debug = false;
+    if (debug) console.log("conjectures", conjectures.length);
+    //if (!debug)
     if (conjectures.length === 0) return;
-    // TODO: we need some randomized proving here:
-    // move free elements and check conjectures a number of times.
-    // For now assume that all conjectures could be verified.
+    backupGeo();
```
gagern: I don't think this is the appropriate function here. `backupGeo` and `restoreGeo` don't maintain a stack, but just a single backup for the sake of animations. So while I see no problem running the prover at widget load time, we will run into trouble if we use those functions while some animation is playing.

But we don't really need `backupGeo`. Much of its code is related to saving and restoring the state of behavior properties. We are only dealing with geometry, here, so copying the geometric state should be sufficient. Add a name like `"prover"` to `stateArrayNames`, then use `stateArrays.prover.set(stateIn);` to back up and `stateIn.set(stateArrays.prover); recalcAll();` to restore.

kortenkamp: Let me suggest something: Just introduce a stack for backupGeo… ;-)

gagern: I'd rather rename those to something containing the word *animation*, or move their content into `csplay` and `csstop`.

Having a fixed number of backup states for a variable number of objects works well. Having a variable number of backup states for a fixed number of objects would be fine as well. But since we have `createtool` the number of objects is no longer fixed. And allocating in several places to keep the states in sync is trouble I'd rather avoid unless I see good reason not to. And allocating for each individual backup will likely be too costly. Last I checked, allocating typed arrays was far from cheap.

.gitignore
```diff
@@ -8,3 +8,4 @@
 *.swp
 /Selenium/
 /private_examples/
+examples/bugs/*
```
gagern: Any particular reason not to use `private_examples` for this? I'd rather avoid introducing a custom excluded directory for each contributor.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,77 @@ function incidentPC(p, c) {
...
+
+    var ii, jj, emove;
```
gagern: `ii` and `jj` appear to be unused. `emove` could be declared where it's being used.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,77 @@ function incidentPC(p, c) {
...
+
+    // filter free objects which are involved in conjectures
+    var involved;
+
+    var recalcInvolved = function() {
+        involved = [];
+        conjectures.forEach(function(con) {
+            var invs = con.getInvolved();
+            invs = invs.forEach(function(el) {
+                if (involved.indexOf(el) < 0 && el.movable) involved.push(el);
```
gagern: I think `involved` should also cover non-movable elements, and should be computed transitively. It may well be that some conjecture directly involves two non-free elements, but in that case we need to consider all the free and semi-free ancestors up the `args` connections, to find which things we should move to affect the given conjecture.

src/js/libgeo/Prover.js
gagern: Do we want to set `stateLastGood` if tracing succeeded? Or in other words, if we entered a singularity, do we want to start from the original configuration or from the last state outside the singularity? The former has lower risk of entering the singularity again, but it may stay closer to the original (and possibly special) configuration. I'm not sure which to prefer. Currently you always start from the original, and that's fine for now. But I'd invite everyone to think about this distinction.

gagern:
I finally managed to have a closer look at the code here. First off thanks again for all that work, @strobelm! I've added a number of review comments to the diff, so below are more general concerns not specific to a single chunk of the diff.

* The `rcomplex` code could be shared. And perhaps moved to the `CSNumber` namespace.
* I dislike the idea of abusing the `"mouse"` type for this, since it's not really a mouse movement. Not sure whether this is worth worrying, though. We can easily change this on a per-operation basis if the need arises. Those types using `"homog"` feel more appropriate already.
* I wonder whether it would be better to return ready-to-use parameters, to avoid the extra detour via `getParamForInput`. In that case we could use `type="param"` throughout, and have `traceMover` skip the `getParamForInput` in this special case. Somehow this feels cleaner to me, but it's not urgent.
* Can we please have some unit test to check that all movable operations come with a `getRandomMove` method?
* I'm still concerned by how strongly this thing depends on the choice of coordinate system. There currently are several cases where objects at infinity won't move at all. And while fixing that specific case is easy enough, this still means that objects “far away” will likely move *very* little, perhaps even so little that the prover doesn't reject some false conjectures. I have no solution except to abandon the goal of staying close to the input configuration. Personally I think it would be elegant if we could just choose a completely random homogeneous coordinate vector for each object, without attempting to make things easier for the tracer by staying close to the original.

strobelm:
Thank you very much Martin! comments below

Am 18.05.2016 um 19:03 schrieb Martin von Gagern:
> 
>   * The |rcomplex| code could be shared. And perhaps moved to the
>     |CSNumber| namespace.

yes, definitely.

>   * I dislike the idea of abusing the |"mouse"| type for this, since
>     it's not really a mouse movement. Not sure whether this is worth
>     worrying, though. We can easily change this on a per-operation basis

The homog parameter wasn't available when i implemented the first parts
of the code, and i worked around quite a while (perhaps more than
implementing homog on my own).

>     if the need arises. Those types using |"homog"| feel more
>     appropriate already.
>   * I wonder whether it would be better to return ready-to-use
>     parameters, to avoid the extra detour via |getParamForInput|. In
>     that case we could use |type="param"| throughout, and have
>     |traceMover| skip the |getParamForInput| in this special case.
>     Somehow this feels cleaner to me, but it's not urgent.
>   * Can we please have some unit test to check that all movable
>     operations come with a |getRandomMove| method?

I'll look into that.

>   * I'm still concerned by how strongly this thing depends on the choice
>     of coordinate system. There currently are several cases where
>     objects at infinity won't move at all. And while fixing that
>     specific case is easy enough, this still means that objects “far
>     away” will likely move /very/ little, perhaps even so little that
>     the prover doesn't reject some false conjectures. I have no solution
>     except to abandon the goal of staying close to the input
>     configuration. Personally I think it would be elegant if we could
>     just choose a completely random homogeneous coordinate vector for
>     each object, without attempting to make things easier for the tracer
>     by staying close to the original.

Seen from a theoretical side we should be safe (if i implemented the
movements correctly) nevertheless we should evaluate if we observe
glitches in practice. If so we can switch to completely random behavior.

> 
> —
> <https://github.com/CindyJS/CindyJS/pull/299#issuecomment-220092595>
> 

-- 
Michael Strobel, M.Sc.

strobelm:
I'm little late for birthday presents Martin, nevertheless here is the updated prover. It would be great if you could have thorough look at the state handling and findAllIncis.

src/js/libcs/CSNumber.js
```diff
@@ -765,180 +765,14 @@ CSNumber._helper.solveCubicHelper = function(a, b, c, d) {
     ]);
 };
 
+CSNumber.getRandReal = function(min, max) {
+    var mmin = min || 0;
+    var mmax = max || 1;
+    return CSNumber.real(Math.random() * (mmax - mmin) + mmin);
+};
```
gagern: I'm not completely happy with the handling of default parameters here. Intuitively I'd expect `getRandReal(7)` to return something from the range [0, 7], or perhaps [-7, 7], but not [1, 7] as it does. Perhaps more serious: I'd expect `getRandReal(-1, 0)` to return a value from the range [-1, 0], but since zero is falsey, I'd actually get a result from the `[-1, 1]` range.

I know none of this affects how you are *currently* using the function. But it might still be confusing to future developers. So I suggest you either drop the substitution of default values (and force developers to always provide two arguments or to implement defaults *when the need arises*), or you do something like this:

```js
    var mmin, mmax;
    if (max !== undefined) {
        mmin = min;
        mmax = max;
    } else if (min !== undefined) {
        mmin = -min; // symmetric by default, or do we prefer zero here?
        mmax = min;
    } else {
        mmin = -1; // symmetric as well, for the sake of consistency
        mmax = 1;
    }
```

The same holds for the other random generating functions, although if you decide to allow default arguments, then in many cases you can just pass arguments along without bothering which of them are defined and which are not, letting the `getRandReal` (and perhaps also the `getRandComplex`) function handle the case distinctions.

src/js/libcs/List.js
```diff
@@ -2511,3 +2511,17 @@ List.ofGeos = function(geos) {
         };
     }));
 };
+
+List.getRandRealVec = function(min, max) {
+    var mmin = min || 0;
+    var mmax = max || 1;
+
+    return List.turnIntoCSList([CSNumber.getRandReal(mmin, mmax), CSNumber.getRandReal(mmin, mmax), CSNumber.getRandReal(mmin, mmax)]);
```
gagern: I'd prefer to see this wrapped to stay within 80 columns. One line per `getRandReal` invocation, plus one for the opening and one for the closing stuff around them. This also makes it more obvious that you are doing the exactly same invocation three times.

I wonder whether the length of the vector, i.e. 3, should be encoded in the name of the function. Analogous to `det3` and `adjoin3`. I think that would make sense.

src/js/libgeo/GeoOps.js
```diff
@@ -713,6 +755,26 @@ geoOps.CircleMr.updatePosition = function(el) {
...
+    var r;
+    var oldr = el.radius;
+    var oabs = CSNumber.abs(oldr).value.real;
+
+    // if radius was small we want something larger and if not we scale the old one
+    if (oabs < CSNumber.eps) {
+        r = CSNumber.getRandReal(0.05, 0.10);
+    } else {
+        r = CSNumber.mult(oldr, CSNumber.getRandReal(0.95, 1.05));
+    }
```
gagern: Is there a good reason why the radius has to be real? If there is, it should be commented upon. If not, choosing a complex radius would be more consistent.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,108 @@ function incidentPC(p, c) {
...
+
+    // filter free objects which are involved in conjectures
+    var involved;
+
+    var recalcInvolved = function() {
+        involved = [];
+        conjectures.forEach(function(con) {
+            var invs = con.getInvolved();
+            var incis;
+            invs = invs.forEach(function(el) {
```
gagern: `forEach` returns `undefined` so using the result is pointless and confusing here.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,108 @@ function incidentPC(p, c) {
...
+                    incis.forEach(function(i) {
+                        if (involved.indexOf(i) < 0) involved.push(i);
+                    });
+                }
+            });
+        });
+    };
+
+    // recursively find all incidences to an geo object
+    var findAllIncis = function(el, list) {
```
gagern: I don't think *incidences* is the correct concept here. What we need is to know which elements we should move in order to check a given conjecture. So I'd rather traverse the `args` properties of our elements, i.e. the chain of *defining* elements. One possible counterexample is a `CircleMP`, were one would want to move the midpoint even though it is not incident with the circle. So I'd say the name and implementation of that function should change.

Since we identify all objects by their name, we might speed things up by using an obect as a hash map, instead of those `list.indexOf` calls all over the place.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,108 @@ function incidentPC(p, c) {
...
+            var el = involved[oo];
+            if (!el.pinned && geoOps[el.type].isMovable) {
+                if (debug) console.log("prover: moving element", el.name);
+                // get random move and move free element
+                emove = geoOps[el.type].getRandomMove(el);
+                movepointscr(el, emove.value, emove.type);
+                // backup configuration if tracing succeeded 
+                if (!tracingFailed) stateArrays.proverTmp.set(stateLastGood);
+                // restore last good configuration if not
+                else stateIn.set(stateArrays.proverTmp);
```
gagern: I think this should not be neccessary. Let me think aloud for a moment, to see whether I'm right.

You are essentially using `proverTmp` as a synonym for `stateLastGood`, I think. The `!tracingFailed` check is already [in `traceMouseAndScripts`](https://github.com/strobelm/CindyJS/blob/d92c1e166205d9b05faeda2c647ec56372a4257c/src/js/libgeo/Tracing.js#L113), the `stateLastGood.set(stateIn)` is [in `stateContinueFromHere`](https://github.com/strobelm/CindyJS/blob/d92c1e166205d9b05faeda2c647ec56372a4257c/src/js/libgeo/Tracing.js#L54) and the `stateIn.set(stateLastGood)` is [in `traceMouseAndScripts`](https://github.com/strobelm/CindyJS/blob/d92c1e166205d9b05faeda2c647ec56372a4257c/src/js/libgeo/Tracing.js#L101) again.

But you are calling [`movepointscr`](https://github.com/strobelm/CindyJS/blob/d92c1e166205d9b05faeda2c647ec56372a4257c/src/js/libgeo/Tracing.js#L130-L139) which does a number of calls to `stateContinueFromHere` that break this use of state buffers. I think it would be better not to go through `movepointscr` but to call `traceMover` directly, so that all the individual moves combine to form a single grand move.

But there is a problem here: we are not moving a single element along a multi-segment path, but instead we are switching between objects after each segment. Which means the default machinery for tracing a multi-segment path is indeed not appropriate here, and we need some new code to throw away the last move before we switch mover.

But the code we have here doesn't do that right either: it does restore the previous state arrays, but we don't have all our state in those arrays. In particular, the `homog` properties of the elements are not updated. We need to recalc those for a suitably chosen set of elements to actually undo a move.

gagern:
I'm reassigning to the author for now, to get the issues I pointed out addressed. Please re-assign to me once that is done.

strobelm:
Hi, i think the code is ready. It would be great if @gagern could check especially the tracing component.

src/js/libcs/CSNumber.js
```diff
@@ -774,180 +774,17 @@ CSNumber._helper.solveCubicHelper = function(a, b, c, d) {
     ]);
 };
 
+CSNumber._helper.getRangeRand = function(min, max) {
+    return Math.random() * (max - min + 1) + min;
```
montaga: Why do we need this `+ 1` here? This function might return numbers greater than `max`.

src/js/libgeo/Prover.js
montaga: I agree, a `involved[el] = true` and `if(involved[el])` would be both more readable and faster.

src/js/libgeo/Prover.js
```diff
@@ -68,21 +74,111 @@ function incidentPC(p, c) {
...
+        if (nincis.length === 0) return list;
+        // add new incidences
+        nincis.forEach(function(ii) {
+            list.push(csgeo.csnames[ii]);
+        });
+        // recursive call
+        nincis.forEach(function(nel) {
+            return findAllIncis(csgeo.csnames[nel], list);
+        });
+        return list;
```
montaga: same here, `list` as dict would probably faster and readable (but change the name before)

PR #300, opened 2016-03-11
Upgrade outdated node packages

gagern:
This is based on the output of `npm outdated`.  It seems as though it should be safe for us to update all packages, even those with a bump in their major version:
* expect 1.15.2 fixes the incompatibility introduced by 1.15.0
* glob 7 checks that cwd is a directory, which it should be in our cases
* should 8 differs in how window.should is treated, which doesn't concern us
* source-map 0.5 no longer supports node 0.8, which we don't support either


strobelm:
thanks again @gagern. Merged.

PR #301, opened 2016-03-16
SVG and PDF support

gagern:
This adds two new commands, `exportSVG` and `exportPDF`, to the cdy instance object. These render the current construction in the corresponding format, and open the result in a new tab. CindyScript bindings are planned but not implemented yet. So right now the way to test this would be by launching any example, opening the browser's web developer console and typing one of these commands:

    createCindy.instances[0].exportSVG();
    createCindy.instances[0].exportPDF();

When drawing images you may have to serve the example via HTTP, e.g. using [st](https://github.com/isaacs/st) (i.e. `npm install st && node_modules/.bin/st`).


strobelm:
Wow amazing, thanks a lot @gagern. Since this is quite urgent and does not touch other core functionality i'll merge this directly. I tested a larger example which looked very nice.

gagern:
Thanks @strobelm.

PR #302, opened 2016-03-17
CSV support

strobelm:
This pull request will add CSV support using papaparse (MIT license www.papaparse.com). Strings will be converted to CindyScript strings and real Numbers to CSNumbers. I don't think that complex numbers are standardized in CVS and therefor we don't support them. I could add the Cindy complex number notation if desired. Thanks a lot @gagern for helping me out with the script loading process.

Next stop: computational statistics with CindyJS ;-).


gagern:
Note that in the near future we'll be making function names case-insensitive, so we'll have to rename the `evaluator` element accordingly. But we can do that as part of the pull request which does the lowercasing, and we can keep the docs in the readable camle case anyway. So no need for action now, just something for me to remember.

gagern:
Please rename [`CVS`](https://en.wikipedia.org/wiki/Concurrent_Versions_System) to [`CSV`](https://en.wikipedia.org/wiki/Comma-separated_values) consistently.

Please have a look at [RFC 4180](http://tools.ietf.org/html/rfc4180), and add doctests to check its claims. So for example I'd like to see a test involving double quotes, even if writing the input for this in CindyScript can be tedious due to the lack of escape sequences. Doubled quotation marks (as a means to escape them) should be documented either way.

In particular “Spaces are considered part of a field and should not be ignored” appears to be violated by the reftest. If such violation is deliberate, please document that fact and provide a reason. If it's due to the library, and can't be affected by configuration, perhaps we should ask its developers to address this, or choose a different one, or write one ourselves?

Please start every line of the reftest with four spaces, a `>` and another space. That part of it should be stripped by the test runner, leading to a test with multi-line input. If not, it's a bug in the runner.

Do you intend to expose some of papaparse's config via modifiers? For example using `;` as a delimiter, as is common in Germany? Or using tab as a delimiter? Perhaps also a way to suppress auto-conversion to types other than string, the way `tokenize` does?

What about conversion to boolean? You didn't mention that. It would be nice to have an example of that as well, and some reftest to check it works, e.g. by extracting the value from the result and applying the `!` operator to it.

strobelm:
Hi thanks for the feedback!

I added support for Bools. Whitespace, other delimiters and double quotes are fine. I also added modifiers for delimiters and newline characters. The doctests fail because the papaparse dependency isn't loaded properly.

gagern:
This needs to be rebased.

gagern:
Superseded by #433 or #434.

Issue #303, opened 2016-03-19
unit tests for extra modules


strobelm:
Currently it's not possible to run tests for extra modules like convexhull3d or the currently developed CSV parser.

kranich:
It might also be interesting for plugins like Cindy3D or CindyGL to run unit tests.

gagern:
@kranich: Cindy3D and CindyGL as a whole have mainly visual effects; checking their CindyScript API will probably not be much use since most functions return nada anyway, which doesn't tell you anything about whether the function performed as expected. It makes sense to test individual parts of Cindy3D and CindyGL in unit tests, but not using the ref test suite. convexhull3d and CVS parser are a different issue, though, and making these work would indeed be great. I'll look into it.

gagern:
Status update: I changed the reftest runner to execute its code in a [node vm sandbox](https://nodejs.org/api/vm.html). But loading quickhull in there fails, since it expects to perform excessive DOM manipulations to load the appropriate module asynchroneously. So I can think of several alternatives:

1. Don't support GWT-compiled and other asynchroneously-loading code in reftests
2. Pick the most appropriate GWT compilation and load that directly
3. Manually extend the DOM capabilities in a hackish way till GWT loads
4. Use [latest jsdom](https://github.com/tmpvar/jsdom#readme) to simulate the DOM, and update our node requirement to Node 4 as required by this
5. Use [jsdom 3](https://github.com/tmpvar/jsdom/tree/3.x#readme) to simulate the DOM
6. Try out [domino](https://github.com/fgnass/domino#readme), see whether that works for us

I'll have a closer look at these alternatives, and do hope that one of them ends up in a pull request soon.

gagern:
See [my domino branch](https://github.com/gagern/CindyJS/commits/domino) for a first stab at 6. Pretty hackish, and still a lot of overhead there. I think I should give latest jsdom a try next, see how that compares.

PR #304, opened 2016-03-19
Convex hull

nasay:
Hi everyone. My name is Natalia and I am working on reimplementation of the quick hull algorithm from Java to Javascript. 
It is not real pull request, it is just an intermediate step on which I want to get some feedback in order to understand what I am doing in the wrong way. 

What is done:

1. All related java-files are removed;
2. There are no Vector3d and Point3d files anymore, for them I used List;
3. Face and Half-Edge are partly implemented;
4. From quick hull algorithm only initial simplex is implemented.

By now I have tested it only for ```examples/50_ConvexHull3D.html``` and only for ```obj1``` .

To compare: <a href="https://github.com/CindyJS/CindyJS/tree/942607919c9cac901246aa4bda9261b4d4fe3002/src/java/cindyjs/quickhull3d">Java files</a>.


nasay:
:up:  
Comments are very welcome.

src/js/Setup.js
```diff
@@ -256,8 +256,8 @@ function createCindyNow() {
 }
 
 function loadExtraModules() {
```
nasay: I am not sure: are the functions ```loadExtraModules``` and ```loadExtraPlugin``` still needed?

strobelm: On 03/20/2016 12:35 PM, nasay wrote:
>
> In src/js/Setup.js
> <https://github.com/CindyJS/CindyJS/pull/304#discussion_r56762833>:
>
> > @@ -256,8 +256,8 @@ function createCindyNow() {
> >  }
> >  
> >  function loadExtraModules() {
>
> I am not sure: are the functions |loadExtraModules| and
> |loadExtraPlugin| still needed?
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/CindyJS/CindyJS/pull/304/files/e1e62973386fd0bd9ab761a199df911deb96d0f5#r56762833>
>
Hi,

for convexhull3d the functions will be unnecessary as soon we replaced
the plugin, nevertheless i'm planning to use them for CSV support
(https://github.com/strobelm/CindyJS/blob/csv/src/js/Setup.js#L262),
therefor removing them will cause merge conflicts later on (in the case
my branch will get merged).

-- 
Michael Strobel, M.Sc.

src/js/libcs/Operators.js
```diff
@@ -4339,3 +4339,26 @@ evaluator.setsimulationquality$1 = function(args, modifs) {
     }
     return nada;
 };
+
+/**
+ * @todo
+ */
+evaluator.convexhull3d$1 = function(args) {
+    var pointsList = args[0].args[0].stack[0];
```
strobelm: Here is something odd, i would have expected something like evaluateAndVal(args[0]).

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,271 @@
...
+
+            currentEdge = currentEdge.next;
+        } while (currentEdge !== this.halfEdge0);
+
+        headPoint = halfEdgeMax.head.point;
+        tailPoint = halfEdgeMax.tail().point;
+
+        lenMax = Math.sqrt(lenSqrMax);
+
+        u = List.scaldiv(lenMax, List.sub(headPoint, tailPoint));
```
strobelm: I think this is wrong, since List.scaldiv expects the scalar to be a CSNumber. CSNumber.real(lenSqrMax) could help.

strobelm:
Hi, thanks for the code!
First of all, i'm fine with using instantiated objects instead of literals. Perhaps @gagern, @kranich, @kortenkamp and @richter-gebert could comment on that since i wasn't involved in these decisions. 

After a brief look on the code, i'm not having major concerns despite the things i mentioned. Actually it looks quite good to me. 

I'm sure @gagern will also add comments on the code as soon he got some spare time.

src/js/Setup.js
nasay: Just deleted the line which were commented: https://github.com/CindyJS/CindyJS/pull/304/commits/ad3e024fd3f2bb4a37de79ac3a92032adefd94da

src/js/libcs/Operators.js
nasay: https://github.com/CindyJS/CindyJS/pull/304/commits/a9337732d16d64a689ee01d52a7f445d93fd5269

src/js/quick-hull-3d/face.js
nasay: https://github.com/CindyJS/CindyJS/pull/304/commits/bf340576cd45a23278de04d137fe07cf63572da3

gagern:
Hello Natalia, thanks for working on this! I just had a *very* quick look at the code. First off, I didn't attempt to understand the old code in detail, so I can't just look at the new implementation to tell whether it's correct. If everything works as it should, and a sufficient number of tests say it is, then that's good enough for me and you don't have to care about the rest of this comment.

If you want us to look at the code along the way, we may need some help. Reviewing the intermediate state of things by looking at the diffs is a bit hard. I'm just wondering whether I can artificially generate some intermediate step, which at some points already follows the new naming and indentation, but doesn't contain any real code changes yet. That *might* help spotting the differences in implementation more easily, or at least provide a diff one method at a time. But it may require some work, perhaps even more than a manual review of the whole set of changes.

Either way, I'd suggest you port the existing code to CindyJS with as few modifications of your own as possible. Sure, if there are whole classes you know you'll drop, dropping them now probably saves a lot of work. So getting rid of those vector and point classes was right, imo. But for the rest, I'd create classes and methods the canonical way at first, and review and test that. If everything works as it should, *then* we can think (or measure) whether changing from constructor invocations to object literals can speed things up, and maybe let you implement that in a separate set of changes. The same goes for any other improvements you have in mind: write them down, and apply them later, to help review.

If there are any specific portions of the porting efforts you'd like me to have a closer look at, let me know. Otherwise I'll have a look at the whole code when I have more time, in which case you shouldn't wait on me.

nasay:
Hello, Martin, 

>and a sufficient number of tests say it is

I actually did not find any tests for convex hull. Can you please show me, where they are, if they exist? 

> Reviewing the intermediate state of things by looking at the diffs is a bit hard. I'm just wondering whether I can artificially generate some intermediate step, which at some points already follows the new naming and indentation, but doesn't contain any real code changes yet.

I can restore Java-files and temporally rename my ```convexhull3d``` to ```convexhull3d_1``` or something. And if I comment <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/QuickHull3D.java#L1290-L1296">this lines</a> it will generate just initial simplex. So I probably can make an example with both implementations and it will not be very hard.

 > at least provide a diff one method at a time

Will it help and will it be enough if I split my commit to a several for example: one for faces, one for half-edges and three for quick hull (min-max calculation, creation of the initial simplex and getting a list of faces and vertices)?

> I'd suggest you port the existing code to CindyJS with as few modifications of your own as possible.

Okay, that makes sense. But can I change names a little bit: ```he0``` to ```halfEdge0```? I will revert more serious name changes and modifications (well, not revert, I can save them to another branch).

gagern:
> I actually did not find any tests for convex hull. Can you please show me, where they are, if they exist?

We have [one disabled test case](https://github.com/CindyJS/CindyJS/blob/0e1053487e67e1d8918e3162a643efdb71c76779/ref/Vectors_and_Matrices.md#computing-a-convex-hull-in-3d-convexhull3dlist-of-vectors). #303 is looking for ways to actually run those. But either way these are not enough, so it would be good to create suitable test cases, and run them through Cinderella and CindyJS.

> So I probably can make an example with both implementations and it will not be very hard.

That sounds like a step in the right direction, with regards to test coverage. At the moment it's probably easiest to build an example HTML file, starting from [`template.html`](https://github.com/CindyJS/CindyJS/blob/master/examples/template.html). There you can write the output to console, or check it in CindyScript code and use e.g. the label or position of a point to indicate whether the check succeeded.

> I can restore Java-files and temporally rename my `convexhull3d` to `convexhull3d_1` or something.

I'd do three commits: one which changes the file names from `*.java` to `*.js` (and drops those you want to drop), one which changes the syntax from Java classes to JavaScript prototypes, and one which adjusts the indentation. Then I'd compare what I get there against your code. Only if the resulting diff is easier to read, then I'd consider whether we want these auto-generated commits as part of the history (via a `git rebase`). But perhaps a diff between branches is already enough. Either way, no need for action on your part just now.

> Will it help and will it be enough if I split my commit to a several

In terms of review, there is not much difference between different commits and a single commit which changes different files. So I see no need to split your commits along file boundaries or the likes. Different commits are more important the closer the affected regions of code are. So if you do some porting and then some renaming or optimization or whatever on that same piece of code, then having those two changesets in different commits helps a lot.

> But can I change names a little bit: `he0` to `halfEdge0`?

Just keep in mind that I tend to review one method at a time. So changes to variables which are local to a given method should be no problem unless the method is *very* long. Changes to private members of a class are more problematic, and changes to the public (or package-private) interface of a class should be kept to a minimum. I guess I can keep track of five such global interface renames in my head, but if there are more, then I'd welcome having them in a separate commit.

gagern:
I created a [`quickhullJS` branch](https://github.com/gagern/CindyJS/commits/quickhullJS) and diffed that against your code using `git diff gagern/quickhullJS..nasay/convex-hull src/js/quick-hull-3d/`. (Notice the two dots instead of three, for a real difference instead of just added commits. GitHub doesn't seem to be able to do this using its web interface.) The [result](https://gist.github.com/gagern/b0d8ac715280ba94e0d9) is slightly easier to read than what we had before, but not by much. Often the first line of a method still doesn't match in the files, due to an empty body, or due to an added underscore, or due to a removed parameter. So I'll probably have to go through all of this manually, without technical help from diff.

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,270 @@
...
+
+    do {
+        this.centroid = List.add(this.centroid, halfEdge.head().point);
+        halfEdge = halfEdge.next;
+    } while (halfEdge !== this.halfEdge0);
+
+    this.centroid = List.scaldiv(CSNumber.real(this.numberOfVertices), this.centroid);
+};
+
+Face.prototype.computeNormal = function(minArea) {
```
nasay: There were two version of ```computeNormal```: ```computeNormal(normal)``` and ```computeNormal(normal, minArea)```. Now it is ```computeNormal(minArea)``` (I hope to split it later).
And since instead of Vector I am using List, which operations actually returns a new object, I save normal in this normal. It seems to be correct since in java files to compute normal ```this.normal``` was used: https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L62. I hope that it is an acceptable change of the API.

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,270 @@
...
+        this.normal = List.add(this.normal, List.cross(d1p, d2p));
+
+        halfEdge2 = halfEdge2.next;
+        this.numberOfVertices++;
+    }
+
+    this.area = List.abs(this.normal);
+    this.normal = List.scaldiv(this.area, this.normal);
+
+    if (minArea === void 0) {
```
nasay: Here ```computeNormal(normal, minArea)``` starts.

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,268 @@
...
+        oppositeFace = adjacentEdge.oppositeFace();
+
+    discardedFaces[numberOfDiscarded++] = oppositeFace;
+
+};
+
+Face.prototype.getSquaredArea = function(halfEdge0, halfEdge1) {};
+Face.prototype.triangulate = function(newFaces, minArea) {};
+
+Face.prototype._computeNormalAndCentroid = function(minArea) {
```
nasay: There also were two versions: with and without ```minArea```.

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,268 @@
...
+    var numberOfVertices, halfEdge;
+
+    this.computeNormal(minArea);
+    this.computeCentroid();
+    this.planeOffset = List.scalproduct(this.normal, this.centroid);
```
nasay: This 3 lines are common: https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L154-L158

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,268 @@
...
+
+    if (minArea !== void 0) {
```
nasay: ```computeNormalAndCentroid()```: <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L136">java version</a>

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,268 @@
...
+            halfEdge = halfEdge.next;
+        } while (halfEdge !== this.halfEdge0);
+
+        if (numberOfVertices !== this.numberOfVertices) {
+            throw new Error('Face ' + this.getVertexString() + ' should be ' + this.numberOfVertices);
+        }
+    }
+};
+
+Face.prototype._connectHalfEdges = function(previousEdge, halfEdge) {
```
nasay: This function is not in use right now. But <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L330">java version</a>

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,271 @@
...
+            halfEdge = halfEdge.next;
+        } while (halfEdge !== this.halfEdge0);
+
+        if (numberOfVertices !== this.numberOfVertices) {
+            throw new Error('Face ' + this.getVertexString() + ' should be ' + this.numberOfVertices);
+        }
+    }
+};
+
+Face.prototype._connectHalfEdges = function(hedgePrev, hedge) {
```
nasay: I do not use this function right now. 
<a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L330">java version</a>

src/js/quick-hull-3d/face.js
```diff
@@ -0,0 +1,269 @@
...
+    halfEdge2.next = halfEdge0;
+
+    face.halfEdge0 = halfEdge0;
+
+    face._computeNormalAndCentroid(minArea);
+
+    return face;
+};
+
+Face.create = function(vertices, indices) {
```
nasay: <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/Face.java#L195">java version</a>

src/js/quick-hull-3d/quick-hull-3d.js
```diff
@@ -0,0 +1,447 @@
...
+QuickHull3D.prototype.setExplicitDistanceTolerance = function(tolerance) {};
+QuickHull3D.prototype.getExplicitDistanceTolerance = function() {};
+QuickHull3D.prototype.addPointToFace = function(vertex, face) {};
+QuickHull3D.prototype.removePointFromFace = function(vertex, face) {};
+QuickHull3D.prototype.removeAllPointFromFace = function(face) {};
+
+/**
+ * Constructs the convex hull of a set of points
+ */
+QuickHull3D.prototype.build = function(points, numberOfPoints) {
```
nasay: 4 versions of ```build```: for point and for coordinates, with and without number of points: https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/QuickHull3D.java#L451-L534

src/js/quick-hull-3d/quick-hull-3d.js
```diff
@@ -0,0 +1,447 @@
...
+        this.pointBuffer = points;
+    }
+};
+
+/**
+ * Finds extreme vertices along each cardinal axis and
+ * sets them to this.maxVertices and this.minVertices
+ * @todo split to several functions
+ */
+QuickHull3D.prototype._computeMaxAndMin = function() {
```
nasay: <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/QuickHull3D.java#L608">java version</a>

src/js/quick-hull-3d/quick-hull-3d.js
```diff
@@ -0,0 +1,447 @@
...
+    var max = List.turnIntoCSList(this.pointBuffer[0].point.value.slice()),
+        min = List.turnIntoCSList(this.pointBuffer[0].point.value.slice()),
```
nasay: ```max``` and ```min``` are to be completely different objects.

src/js/quick-hull-3d/quick-hull-3d.js
```diff
@@ -0,0 +1,447 @@
...
+    } else {
+        this.tolerance = this.explicitTolerance;
+    }
+};
+
+/**
+ * Creates the initial simplex from which the hull will be built.
+ * @todo split to several functions
+ */
+QuickHull3D.prototype._createInitialSimplex = function() {
```
nasay: <a href="https://github.com/CindyJS/CindyJS/blob/master/src/java/cindyjs/quickhull3d/QuickHull3D.java#L665">java version</a>

examples/test-convex-hull.html
```diff
@@ -0,0 +1,195 @@
+<!DOCTYPE html>
```
nasay: I added this example to compare two implementations and reverted all my changes of the example 50. 
Mostly it is copy/paste from the example 50. The difference is that ```obj1``` has 5 points and there is ```println``` of faces and points.

nasay:
new one https://github.com/CindyJS/CindyJS/pull/360

PR #305, opened 2016-03-22
Allow setting the homog field of horizontal and vertical lines

### LINKED ISSUE #95: Programatically setting positions for lines

gagern:
This should finally fix #95.

The homog setters for horizontal and vertical lines just set one coordinate to zero. So they intersect the coordinate axes at the same position as the line described by the input coordinates. The provided example demonstrates this.

### DIFF
--- a/examples/139_LineHomogSetter.html
+++ b/examples/139_LineHomogSetter.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


kranich:
@gagern Thanks, looks good. I especially appreciate handling of `Through` in src/js/libcs/Accessors.js being cleaned up.

PR #309, opened 2016-04-02
export macro for ellipse by 2 foci and 1 perimeter point 

strobelm:
added a missing macro and removed some default debugging output

### DIFF
--- a/examples/142_foci_macro.html
+++ b/examples/142_foci_macro.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


kranich:
I do not like the ambiguousness of `ConicFoci`. One could also expect a collection of the two foci as the result of that operation. Should we rather change Cinderella export?

@gagern: What do you think?

gagern:
After some past discussion we had decided that in case of a naming conflict, the more precise name should take preference. Applying this policy to `ConicFoci` vs. `ConicBy2Foci1P` I think we should stay with the latter, i.e. change Cinderella to emit that. Will write a pull request for it.

PR #310, opened 2016-04-04
Add macro ConicFoci as alias for ConicBy2Foci1P

kranich:
Cinderella exports `ConicFoci` as the name for a conic defined by two foci and one point, whereas CindyJS uses `ConicBy2Foci1P`.

This pull request adds a macro `ConicFoci` as an alias for `ConicBy2Foci1P`.

I am slightly unhappy with the ambiguousness of `ConicFoci`. So it might be better to adapt the export to CindyJS in Cinderella. What do you think?


kranich:
Sorry, missed the duplicate #309.

PR #312, opened 2016-04-04
Source maps

gagern:
Two improvements aimed at making source maps more useable. The first fixes an incorrect path for closure-compiled plugins, introduced by 6c689f3706ad671879b9cb19f4d463a5dac38fde for #191. The other includes the content of generated sources in the sourcemap itself, since those are obviously not available from GitHub.


kranich:
@gagern Thanks.

PR #314, opened 2016-04-06
Reduce number of assertion frameworks

gagern:
Instead of using assert, should, expect and chai, we now only use the different aspects of chai.  This will reduce the memory and download time required for our dev dependencies, and should also make our test cases slightly more consistent with one another.

The CSNumber test was written using expect but essentially was using assertions, so I switched it to assertion style for brevity and clarity.


strobelm:
The code looks fine, nevertheless travis failed. Something seems to be wrong with the node executable. I'm not sure if this is related to this pull request at all.

kranich:
Travis complains that `.mailmap text` is missing in `.gitattributes`.

kranich:
@strobelm, @gagern Will this be merged soon or shall I reopen #313, restore gagern/CindyJS:people, push 86bb8ea there and merge?

gagern:
I took @strobelm's comment as a positive review, so I merged this now.

PR #317, opened 2016-04-11
Allow grid, snap, axes settings per port, draw axes

### LINKED ISSUE #315: Drawing coordinate axis
In Cinderella, it is possible to show coordinate axes. This should also be possible in CindyJS.

A property that indicates whether axes are to be shown should also be exported from Cinderella.

gagern:
This fixes #315 since it adds support for an `axes` property of the port elements. However, it's not perfect since it does alpha blending, which results in darker colors in areas covered by more than one line, like the origin or at the tips of the arrow heads. Cinderella explicitely mixes foreground and background color and then performs an opaque drawing with the result. Should we do the same? What about cases where we don't even know the background color? Fall back to what the code does now? Or render grid and axes to some other canvas and then draw that canvas with some global alpha? Any other suggestions?


kranich:
@gagern Do you want to merge this now, or do you want to work on what we discussed in today's meeting before that?

gagern:
I'd say merge now, since imo it's better than what we used to have, at least for the case of content exported from Cinderella. And not *that* much worse for other content. I'll still work on this, and just [posted to Stack Overflow](http://stackoverflow.com/q/36578651/1468366) looking for a more elegant solution, but I have doubts whether one exists. So I guess I might end up with the two-canvas approach. If you have concerns about how this affects existing content, feel free to delay till I've got something better.

kranich:
@gagern Thanks for your work. Please feel free to reopen #315 as a reminder for pending improvements, if you want.

PR #325, opened 2016-04-13
faster center of conic

strobelm:
This pull request removes an unnecessary matrix-vector multiplication in the center of conic routine.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -618,8 +618,14 @@ geoOps.PointOnSegment.stateSize = 2;
 
 
 geoOps._helper.CenterOfConic = function(c) {
-    // The center is the pole of the line at infinity.
-    return General.mult(List.adjoint3(c), List.linfty);
+    // The center is the pole of the dual conic of the line at infinity
+    var adj = List.adjoint3(c);
+    // return General.mult(adj, List.linfty);
+    // do not use matrix vector multiplication, we know the result
+    return {
+        'ctype': 'list',
+        'value': [adj.value[2].value[0], adj.value[2].value[1], adj.value[2].value[2]]
+    };
 };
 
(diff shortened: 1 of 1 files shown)


kranich:
@strobelm I find the following a bit clearer:
```JavaScript
var adj = List.adjoint3(c);
// return General.mult(adj, List.linfty);
// do not use matrix vector multiplication, we know the result
return {
     'ctype': 'list',
     'value': [adj.value[2].value[0], adj.value[2].value[1], adj.value[2].value[2]]
};
```

Would you mind to change this?

strobelm:
Not at all, it' done. By the way this is about 3 times faster than the
previous version.

Am 13.04.2016 um 17:07 schrieb Stefan Kranich:
>
> @strobelm <https://github.com/strobelm> I find the following a bit
> clearer:
>
> var adj = List.adjoint3(c);
> // return General.mult(adj, List.linfty);
> // do not use matrix vector multiplication, we know the result
> return {
>      'ctype': 'list',
>      'value': [adj.value[2].value[0], adj.value[2].value[1], adj.value[2].value[2]]
> };
>
> Would you mind to change this?
>
> —
> <https://github.com/CindyJS/CindyJS/pull/325#issuecomment-209497910>
>

-- 
Michael Strobel, M.Sc.

kranich:
@strobelm Thanks.

gagern:
While you are at it, note that the third column of the adjoint matrix the same as the cross product of the first two columns of the original matrix. Or in more geometric words, the pole of the line at infinity is the intersection of the polar of two distinct points at infinity, like (1,0,0) and (0,1,0). This should shorten the operation by yet another few unneccessary computations.

But in the long run I'd like to have some machinery which automatically removes all those unneccessary computations so that you can write readable high-level descriptions of operations without incurring unneccessary costs. Still way to go in that direction, though.

PR #326, opened 2016-04-13
Render grid and axes in a single stroke operation each

gagern:
The individual subpaths of a single stroke operation won't alpha-compose
onto one another.  So by including all grid lines in one stroke operation,
and by including both axes and their arrowheads in a second stroke
operation, we can avoid increased opacity at the overlaps.
Thanks to http://stackoverflow.com/a/36582168/1468366 for the hint.

There is still increased opacity at the points where the axes intersect the
perpendicular grid lines, but those might even be seen as a feature,
somewhat like very slight tick marks.  If this should be undersirable,
http://stackoverflow.com/a/36586765/1468366 might offer an alternative.

This switches the grid drawing from CindyScript to JavaScript, since it
currently is not possible to construct complex paths in CindyScript and have
them stroked in a single operation.  The machinery for compilation of
CindyScript to JavaScript stays in place, though, ready to be re-used.

The ends of the axes were shifted in such a way that the tip of the
arrowhead outline points at the border of the canvas, as opposed to the
vertex at the center of the arrowhead line being stroked.  Both this and the
combination of subpaths into a single stroke operation are things worth
considering for the arrow stroking used by scripts and geometric elements.


kranich:
>There is still increased opacity at the points where the axes intersect the
perpendicular grid lines, but those might even be seen as a feature,
somewhat like very slight tick marks. If this should be undersirable,
http://stackoverflow.com/a/36586765/1468366 might offer an alternative.

@gagern Alternatively, we might also use a single path for grid and axes, i.e. `beginPath` if `grid` or `axes` and then individual checks for `grid` and `axes`, at the cost of another `if`.

kranich:
@gagern Sorry, forget it, this does not work due to different stroke styles.

kranich:
@gagern I have just noticed that the size of the arrows has increased quite a bit. Maybe you want to fix this?

I should have noticed before merging, mea culpa, I am sorry. I think I should not merge anything else before Tuesday, when my head will hopefully stop spinning.

Issue #88, opened 2015-11-20
List.triples duplicated


elkins0:
This is just so that I don't forget to report things I find. This is large enough that you might care. You need to let me know to down what level of detail you want me to report. I'll add a comment pertaining to examples of some other things I've noticed.

List.triples appears twice in https://github.com/CindyJS/CindyJS/tree/master/src/js/libcs/List.js:
```javascript
List.triples = function(a) {
    var erg = [];
    for (var i = 0; i < a.value.length - 2; i++) {
        for (var j = i + 1; j < a.value.length - 1; j++) {
            for (var k = j + 1; k < a.value.length; k++) {
                erg.push({
                    'ctype': 'list',
                    'value': [a.value[i], a.value[j], a.value[k]]
                });
            }
        }
    }
    return {
        'ctype': 'list',
        'value': erg
    };
};

List.triples = function(a) {
    var erg = [];
    for (var i = 0; i < a.value.length - 2; i++) {
        for (var j = i + 1; j < a.value.length - 1; j++) {
            for (var k = j + 1; k < a.value.length; k++) {
                erg.push({
                    'ctype': 'list',
                    'value': [a.value[i], a.value[j], a.value[k]]
                });
            }
        }
    }
    return {
        'ctype': 'list',
        'value': erg
    };
};
```

elkins0:
Here are examples of some other things I've noticed.

For instance, in https://github.com/CindyJS/CindyJS/tree/master/src/js/libcs/CSNumber.js we have:

```javascript
CSNumber.zero = CSNumber.real(0);
```

and yet there are 23 occurances of `CSNumber.real(0)` in the code.

In https://github.com/CindyJS/CindyJS/tree/master/src/js/libgeo/GeoOps.js,

```javascript
geoOps.Through.getParamForInput = function(el, pos, type) {
...
        l = List.turnIntoCSList([CSNumber.zero, CSNumber.zero, CSNumber.zero]);
```

using `l = List.realVector([0, 0, 0])` matches the more common way of doing this.

And here ...

```javascript
geoOps._helper.initializePoint = function(el) {
    var sx = 0;
    var sy = 0;
    var sz = 0;
    if (el.pos) {
        if (el.pos.length === 2) {
            sx = el.pos[0];
            sy = el.pos[1];
            sz = 1;
        }
        if (el.pos.length === 3) {
            sx = el.pos[0];
            sy = el.pos[1];
            sz = el.pos[2];
        }
    }
[...]

PR #327, opened 2016-04-13
Createtool

gagern:
I've been sitting on this for far too long now, rebasing it every now and then but never finding the time to get it polished to my satisfaction. I think it might be more fruitful for someone to have a second look, for us to merge this with all the rough edges, and then to address issues as they arise in practice.

Remaining issues that I'm aware of (and that have kept me from requesting this to be merged before):

* icons should be SVG not PNG
* position of the newly added element needs to be updated if some temporary point gets merged with an existing point upon mouse up
* removal of elements doesn't handle depending elements
* removal of elements doesn't deallocate state
* need to use prover to avoid creating duplicate elements

Let me know whether you consider any of these issues a show stopper, to be addressed before merging. Pull requests into my branch highly welcome.


examples/createtool/createtool.html
```diff
@@ -0,0 +1,36 @@
...
+        </style>
+        <script type="text/javascript" src="../../build/js/Cindy.js"></script>
+        <script type="text/javascript">
+            var cindy = createCindy({ 
+        	scripts: "cs*", 
+        	defaultAppearance: {}, 
+        	geometry: [], 
+        	ports: [ 
+        		{ id: "CSCanvas", width: 680, height: 336, transform: [ { visibleRect: [ -9.04, 9.32, 18.16, -4.12 ] } ], background: "rgb(168,176,192)" } ] });
+            </script>
```
elkins0: Erroneous `</script>` on line 25

examples/createtool/createtool.html
```diff
@@ -0,0 +1,36 @@
...
+        <style>
+            body {
+                margin: 0;
+                padding: 0;
+            }
+
+            button {
+                background-color: #FFFFFF;
+                border: 1px solid #CCCCCC;
+            }
```
elkins0: More interesting buttons are possible with this instead of the button style used here :smile:
```css
            #toolbar button {
                padding: 0px 0px 0px 0px;
                line-height: 0px;
                border:0px none red;
                padding: 1px;
                margin: 0px;
            }
            #toolbar img {
                background-image: url(../../build/js/images/Empty.png);
                padding: 2px;
            }
```
empty

src/js/libcs/Operators.js
```diff
@@ -4374,3 +4374,62 @@ evaluator.setsimulationquality$1 = function(args, modifs) {
...
+
+    var button = document.createElement("button");
+    button.innerHTML = "<img src='" + createCindy.getBaseDir() + "images/" + name + ".png'>";
+    button.onclick = function() {
+        if (typeof activeButton !== "undefined") {
+            activeButton.style.border = "";
+        }
+
+        activeButton = this;
+        activeButton.style.border = "1px solid black";
```
elkins0: For a depressed button look like in Cinderella change the `activeButton.style.border` settings to be:
```javascript
4426  -            activeButton.style.border = "";
4426  +            activeButton.children[0].style.backgroundImage = "";
...
4430  -        activeButton.style.border = "1px solid black";
4430  +        activeButton.children[0].style.backgroundImage = "url(" + createCindy.getBaseDir() + "images/Frame.png)";
```
Here is "images/Frame.png": frame

examples/createtool/createtool.html
```diff
@@ -0,0 +1,36 @@
...
+        </style>
+        <script type="text/javascript" src="../../build/js/Cindy.js"></script>
+        <script type="text/javascript">
+            var cindy = createCindy({ 
+        	scripts: "cs*", 
+        	defaultAppearance: {}, 
+        	geometry: [], 
+        	ports: [ 
+        		{ id: "CSCanvas", width: 680, height: 336, transform: [ { visibleRect: [ -9.04, 9.32, 18.16, -4.12 ] } ], background: "rgb(168,176,192)" } ] });
```
elkins0: A sign of Cinderella's ever shrinking "Euclidean View" -- opening another "Euclidean View" and closing the first before exporting gives the consistent port size shown here...
```javascript
width: 680, height: 336, transform: [ { visibleRect: [ -9.04, 9.32, 18.16, -4.12 ] } ]
                     ^^                                    ^     ^      ^     ^^
width: 680, height: 350, transform: [ { visibleRect: [ -9.06, 9.34, 18.14, -4.66 ] } ]
```

elkins0:
Highlighting of selected element(s) to give the user feedback is generally sorely lacking in CindyJS.
Also useful would be using the letters A-G,K-Z for points before resorting to P#. And please consider displaying the line names instead of just the point names...

kortenkamp:
I don't think there are any showstoppers right now, au contraire, it would be great to have this in the widget service to allow for better bug reporting by users from the MC Squared project.

strobelm:
I'm also in favor of merging this as soon as possible. Although we
shouldn't brag that much about it due to its proof of concept character.

Am 17.04.2016 um 22:49 schrieb kortenkamp:
>
> I don't think there are any showstoppers right now, au contraire, it
> would be great to have this in the widget service to allow for better
> bug reporting by users from the MC Squared project.
>
> —
> You are receiving this because you were assigned.
> <https://github.com/CindyJS/CindyJS/pull/327#issuecomment-211111856>
>

-- 
Michael Strobel, M.Sc.

examples/createtool/createtool.html
gagern: Thanks for the hint. But I'd like to see SVG buttons for our tools soon, in which case we won't be needing PNG backgrounds any longer.

examples/createtool/createtool.html
gagern: If I understand you correctly, you're currently saying that Cinderella is failing to recreate its own port sizes. Do you know of a Cinderella issue filed for this? I see no urgent need to fix this here, since we don't really care about which portion of the plane we see; it's all empty anyway.

elkins0: @gagern wrote:
>If I understand you correctly, you're currently saying that Cinderella is failing to recreate its own port sizes. Do you know of a Cinderella issue filed for this? I see no urgent need to fix this here, since we don't really care about which portion of the plane we see; it's all empty anyway.

This was reported 6 months ago as [#88 Save/Close/Load cycle shrinks every window by 14 pixels in height](https://gitlab.cinderella.de:8082/cinderella/driftin/issues/88)

examples/createtool/createtool.html
elkins0: Actually I think it is possible to wrap image files like PNG within an SVG file. So that could be done in the interim. However it is probably not important if user interface entirely contained within this example and none of it ends up being entangled in the `CindyJS` library. To that end, the variable `activeButton;` and the functions `evaluator.createtool$3` and `addTool` belong within the UI on the outside of the `CindyJS` library. The `setActiveTool` is the function that needs to be exported for use by that UI.

elkins0:
@gagern: Here's a gist link to [Line.svg](https://gist.github.com/elkins0/a9958059bd7394f76d065b956c93ca9b) that I created to match Line.png. Sadly github won't let me paste it here (the gist file displays there as an image--but with the wrong aspect ratio, 1:1 instead 4:3 to match 32x24), so here's a 400% enlargement as a PNG:
line
As you can see it is a mighty fine replica even if I say so myself. It is handmade using *inkscape* with the original Line.png shown here:
line
I have rationalized the coordinates with help from *Mathematica*. Hope it meets with your acceptance!
I will finish up the rest of the ones in this pull over the weekend ...

gagern:
@elkins0 Thank you for the enthusiasm, but we are not aiming to exactly replicate the existing icons. In particular, we are looking for some way to more clearly distinguish between input and output of a given primitive. For example the buttons for pole and polar can be hard to distinguish, and we would want something clearer there. We are discussing alternatives in the dev team. Personally I'd also aim for compact had-written SVG, since inkscape can add quite a lot of bloat, which may add up as the number of icons increases.

PR #328, opened 2016-04-13
Reduce arrow size

gagern:
I misread the arrow-drawing code in Render2D, leading to twice the intended arrow size. Found by @kranich in https://github.com/CindyJS/CindyJS/pull/326#issuecomment-209522413.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -295,14 +295,14 @@ function updateCindy() {
         csctx.beginPath();
         csctx.moveTo(0, -m.ty);
         csctx.lineTo(csw - 6, -m.ty);
-        csctx.moveTo(csw - 23, -10 - m.ty);
+        csctx.moveTo(csw - 13, -5 - m.ty);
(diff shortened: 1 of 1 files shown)


kranich:
@gagern Still looks a bit different than before. Arrows were a bit wider before. But I prefer the new look and don't think that the arrows must look exactly the same as other arrows. If you agree, we can merge this.

gagern:
@kranich What do you mean by “before”? Cinderella had different slopes for the arrowheads, but I prefer to keep arrowhead shapes consistent between axes and other arrows.

Or are you comparing with the state of affairs as merged in #317? In that case, the apparent difference is probably because the tip of the arrow head used to be cut off, but now is fully visible, making the arrow look longer:

Comparison of arrowheads
The same at twice the size

kranich:
@gagern I was comparing arrow shapes before and after #326 was merged. So this is ready to be merged now. Thank you for your improvements.

PR #330, opened 2016-04-17
better midpoint handling

strobelm:
Hi,

the midpoint algorithm does not handle the case where both points are at l_infty and returns the zero vector (or perhaps even worse). 

We can specify a correct midpoint analogously to the RP^2 case in RP^1, multiplying with the y-coordinate instead of the z-coordinate. 

On the go i implemented some tests.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -127,9 +127,21 @@ geoOps.Meet.visiblecheck = function(el) {
 };
 
 geoOps._helper.midpoint = function(a, b) {
+    var aZ = a.value[2];
+    var bZ = b.value[2];
+
+    // case both points on l_infty
+    if (CSNumber._helper.isAlmostZero(aZ) && CSNumber._helper.isAlmostZero(bZ)) {
+        var aY = a.value[1];
+        var bY = b.value[1];
+        return List.normalizeMax(List.add(
+            List.scalmult(bY, a),
+            List.scalmult(aY, b)));
+    }
+    // finite points + 1 finite and 1 infinite 
     return List.normalizeMax(List.add(
-        List.scalmult(b.value[2], a),
-        List.scalmult(a.value[2], b)));
+        List.scalmult(bZ, a),
+        List.scalmult(aZ, b)));
 };
 
 geoOps.Mid = {};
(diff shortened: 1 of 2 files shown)


kortenkamp:
I am a bit confused, as I don't see an obvious definition of a midpoint of two points at infinity. Maybe you can explain that to me tomorrow.

elkins0:
For this to make sense it should not matter if you chose to scale by x or y values if the z values are zero. Yet it does:
```javascript
[by*ax+by*ay, ay*bx+ay*by, 0] !== [bx*ax+bx*ay, ax*bx+ax*by, 0]
```
My suggestion would be to use List.abs like so:
```javascript
geoOps._helper.midpoint = function(a, b) {
    var aZ = a.value[2];
    var bZ = b.value[2];
    // case both points on l_infty
    if (CSNumber._helper.isAlmostZero(aZ) && CSNumber._helper.isAlmostZero(bZ)) {
        aZ = List.abs(a);
        bZ = List.abs(b);
    }
    // else finite points or 1 finite and 1 infinite
    return List.normalizeMax(List.add(
        List.scalmult(bZ, a),
        List.scalmult(aZ, b)));
};
```
To me this looks like an angle bisector of lines through the origin with a line of `[dy,-dx,0]` being the dual of a point at infinity `[dx,dy,0]` with the angle bisector line then being the dual of the midpoint.

strobelm:
Thanks @elkins0 for the input! @montaga and i discussed the topic recently in the office and came up with the same solution. Nevertheless more thoughts should be spend on that since @kortenkamp and @richter-gebert stated quite clearly that in their opinion a mid point on l_inf does not make sense at all.

gagern:
There is one crucial distinction: there are *two* angle bisectors at the origin, but there is only one Euclidean midpoint. The choice which one to take is hard to justify. If we were doing generic Cayley-Klein geometry, where there are two midpoints in general, then it *might* make sense to consider midpoits of points at infinity at first glance.

But even then it does not. Consider the following scenario: *A* and *B* are finite points, and *M* is their midpoint. You take *A* and drag it to infinity. The distance *BM* is half the distance *AM*, but in the limit, *A* and *M* have to coincide. We should have this case covered by the current implementation. Next you drag *B* around as well. As long as it remains finite, the midpoint *M* has to remain identical to *A*. You can't explain *M* being anywhere else via a limit process. Yet you suggest that if *B* were at infinity, *M* should suddenly become something else. For that it would have to jump, which is bad™. Instead, *M* would have to be a multiple of *A* and a multiple of *B* at the same time, which is only possible if *M* is the zero vector. Hence getting zero is *exactly* the right result here.

PR #333, opened 2016-05-02
avoided .new functions for FloatArray, UInt16Array etc

montaga:
since they threw errors when using Juergens Safari
+beautify


gagern:
For clarification: this is about the `TypedArray.map` methods, not some `.new` method, right?

[The MDC page for `TypedArray.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/map) indeed indicates rather poor browser compatibility. Which is only half the story. The other half is the fact that the closure compiler doesn't compile this specific ES6 feature to some ES5 analogon since there are no typed arrays in ES5 at all. The [compatibility table row](http://kangax.github.io/compat-table/es6/#test-typed_arrays_%TypedArray%.prototype.map) for this feature may be the better resource here, since it indicates support by many browsers but not by Safari 9. Worth remembering, hadn't known of this myself…

PR #334, opened 2016-05-02
Support drag & drop

gagern:
This is a rebased version of my bastelbogen branch, which demonstrates drag and drop of pictures from iOrnament in order to create a net of some platonic solid using the symmetry group in question.

* `ctype: "image"` is a new value type, which allows access to images without using their name.
* `ondrop` is a new event handler (like `init` or `draw`) which will get evaluated after things got dropped onto the widget.
* `allimages()` is a new CindyScript function to list the names of all named images.
* `dropped()` is a new CindyScript function that returns the list of dropped files, as quadruples `[value, type, mimeType, name]` where `type` currently is one of `"string"` (for MIME type `text/*`), `"image"` (for MIME type `image/*`)  or `"undefined"` (for other files, which have a `value` of `___`).


strobelm:
Great! Thanks @gagern and merged.

PR #335, opened 2016-05-09
New representation of images

gagern:
I recently started working on video support. Videos are mostly like images, since they can be passed as an argument to a `CanvasRenderingContext2D.drawImage` call or a `WebGLRenderingContext.texImage2D` call. But some other things are different. In particular, `width` and `height` represent the attributes of these names, and will therefore be missing if no such attributes were given for the DOM element. The native size of the image is `videoWidth` resp. `videoHeight`.

This change here abstracts away that difference, by introducing an intermediate layer. Now an image object has `ctype:"image"` and `value:{img,width,height,ready,whenReady,…}`. The `value.img` element is the actual HTMLImageElement or HTMLCanvasElement or HTMLVideoElement in question. The width and height are computed in a suitable fashion. An image is considered ready when it's fully loaded, but a video is considered ready as soon as we have sufficient metadata to know its size. We might want to wait for some other state here, I'm not sure. Getting the `ready` flag out of the image element avoids conflict with future or vendor-specific extensions. The `whenReady` function allows registering a callback to be invoked once the image is ready, abtracting away the different underlying events.

The instance-global `images` dictionary now holds complete image objects, i.e. including the outer layer with its `ctype` and `value` properties. This should make some things more consistent.

### DIFF
--- a/examples/Bastelbogen.html
+++ b/examples/Bastelbogen.html
@@ -172,18 +172,14 @@ function utf8Decode(bytes) {
 createCindy.registerPlugin(1, "imgmeta", function(api) {
     api.defineFunction("xmpdescription", 1, function(args, modifs) {
         var img = api.evaluate(args[0]);
-        if (img.ctype === "image") { // dropped image file
-            img = img.value;
-        } else if (img.ctype === "string") {
-            img = api.getImage(img.value);
-            if (!img || !img.src) {
-                console.warn(img.value + " does not name an image");
-                return api.nada;
-            }
-        } else {
+        console.log(img);
+        img = api.getImage(img);
+        console.log(img);
+        if (!img) {
             console.warn("Argument does not name an image");
             return api.nada;
         }
+        img = img.img;
         var data = img.src.replace(/^data:image\/png;base64,/, "");
         if (data === img.src) {
             console.warn("Only data:image/png;base64,… supported");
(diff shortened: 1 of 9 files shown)


examples/Bastelbogen.html
```diff
@@ -172,18 +172,14 @@
...
-        if (img.ctype === "image") { // dropped image file
-            img = img.value;
-        } else if (img.ctype === "string") {
-            img = api.getImage(img.value);
-            if (!img || !img.src) {
-                console.warn(img.value + " does not name an image");
-                return api.nada;
-            }
-        } else {
+        console.log(img);
```
strobelm: Some debug code still active here

plugins/cindygl/src/js/CanvasWrapper.js
```diff
@@ -4,22 +4,21 @@
 function addCanvasWrapperIfRequired(name, api) {
   if (!canvaswrappers.hasOwnProperty(name)) {
     let img = api.getImage(name, true); //this might be a canvas as well
```
strobelm: 'let' should be 'var' to keep ecmascript 5 compatibility.

strobelm:
The code looks fine to me (only minor remarks). 
Some examples seem to be broken (also in the master branch):
cindygl/22_explorer.html
cindygl/17_images_blur.html
@montaga can you comment on them?

plugins/cindygl/src/js/CanvasWrapper.js
gagern: CindyGL, like Cindy3D, will be cross-compiled by Closure from ES6 to ES5. It has ES6 syntax all over the place.

PR #336, opened 2016-05-09
Webcam support

gagern:
This adds elementary webcam support to CindyJS, in the form of a `cameravideo()` function. It builds on the new internal image representation suggested in #335. So it may make sense to review that first, unless you want to review both at once here. You can also look at [the diff relative to that](https://github.com/gagern/CindyJS/compare/images...webcam), if you want to discuss this feature here before #335 has been merged.

Note that CindyGL support doesn't work with this new feature yet. That's a separate endeavor.

### DIFF
--- a/examples/Bastelbogen.html
+++ b/examples/Bastelbogen.html
@@ -172,18 +172,14 @@ function utf8Decode(bytes) {
 createCindy.registerPlugin(1, "imgmeta", function(api) {
     api.defineFunction("xmpdescription", 1, function(args, modifs) {
         var img = api.evaluate(args[0]);
-        if (img.ctype === "image") { // dropped image file
-            img = img.value;
-        } else if (img.ctype === "string") {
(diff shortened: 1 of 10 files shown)


strobelm:
The code looks good and works on firefox, but i couldn't get it running with chromium or chrome. Can someone very that?

gagern:
@strobelm: I've tested on chrome. Apparently it doesn't allow local files access to the webcam. So I used [the `st` module](https://www.npmjs.com/package/st): `npm install st` followed by `node_modules/.bin/st -nc` then visited http://127.0.0.1:1337/examples/webcam1.html. Perhaps we should declare `st` as a dev dependency of CindyJS, and suggest its use somewhere in the documentation?

montaga:
Alternatively, you can also run chrome with the --allow-file-access-from-files flag.

strobelm:
I added a note in the example how work around the problem.

PR #337, opened 2016-05-10
CindyGL: New representation of images

montaga:
This pull request includes #335 and #336.

The new image-representation was tranfered to CindyGL.
Now it is possible to exchange image objects, accessing the webcam and using drag&drop in CindyGL. Examples were added.


gagern:
Thanks for bringing all these things together!

This does not include all of #336, but the core of it. Nevertheless, it will conflict (due to semi-manual beautification in 4664483a2fa99e0260baeef1eb97c9cf3bf71abf vs. automatic beautification in e5d4b2f349585e33b39df1b9cddfb498ebcae88a), so you may want to merge my `newinternalimagestructure` branch where I resolved the conflict.

I don't feel that it would be good for me to review my own contributions to this. So I'll concentrate on [the diff relative to \#336](https://github.com/gagern/CindyJS/compare/4664483a2fa99e0260baeef1eb97c9cf3bf71abf...57d8bb9d4fce951fc1e4a445b8dd54564c920f15), and wait for that one to get reviewed by @strobelm or someone else. Although those changes are almost completely restricted to the CindyGL directory tree, except for changes to `imageready` and an error message in `if`, both of which I introduced myself in f77748d99f778d2e6c75c6f50dcb14585ecd0d37. So I'd be OK with having these changes merged once #336 is in, but I'll try to give it some review nonetheless.

I just gave `examples/cindygl/25_dragdrop.html` a try, and observed the following: Dropping an image into the area looked correct at first, but when I started dragging the points, then suddenly the upper portion of the canvas would be covered by the undimmed and apparently unscaled image, while the lower portion updated as intended. Any clue why this is happening?

montaga:
Thank you for giving examples/cindygl/25_dragdrop.html a try. There was a bug with the computation of the texturecoordinates. It should be fixed now.

PR #339, opened 2016-05-10
Avoid Number.isFinite in favor of unqualified isFinite function

### LINKED ISSUE #338: Number.isFinite not supported old versions of Safari
When trying CindyJS I noticed that it does not work anymore on the old M10-Macbook. The reason for that is that Number.isFinite is used, which is not supported by Safari (<9).
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite it is not supported by IE at all.
Instead we could use isFinite.

gagern:
The former is an ES6 addition, so we should stick to the latter for ES5 (plus typed arrays) compatibility.

This fixes #338. @montaga, can you please test using that same equipment?

### DIFF
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -71,6 +71,10 @@ var haveCanvas = function(canvas) {
     return canvas;
 };
 
+var isFiniteNumber = Number.isFinite || function(x) {
+    return (typeof x === 'number') && isFinite(x);
+};
+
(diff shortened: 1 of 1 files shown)


montaga:
Now CindyJS works on "the old M10-Macbook" again :+1:

PR #343, opened 2016-05-17
Segment clipping

### LINKED ISSUE #332: Do not draw segments beyond boundary of canvas
Segments should be clipped to canvas in `Render2D.drawsegcore`. This is a follow up of #292.

Does anyone have an idea how to achieve this without too many case distinctions?

kranich:
This is based on #292 and should fix #332.

I am not too happy with the performance of examples/18_CompleteGraph.html after these changes. Suggestions of improvements are generally very welcome.

### DIFF
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -523,9 +523,9 @@ Render2D.drawline = function(homog) {
     // do not draw beyond canvas boundary
     var m = csport.drawingstate.matrix;
     var b1 = [m.a, m.b, m.tx]; // left boundary
-    var b2 = [m.a, m.b, m.tx - csctx.canvas.width]; // right boundary
+    var b2 = [m.a, m.b, m.tx - csw]; // right boundary
     var b3 = [m.c, m.d, m.ty]; // upper boundary
-    var b4 = [m.c, m.d, csctx.canvas.height - m.ty]; // lower boundary
+    var b4 = [m.c, m.d, csh - m.ty]; // lower boundary
     if (l[0] === 0) { // horizontal line
         Render2D.drawsegcore(crossxy(l, b1), crossxy(l, b2));
         return;
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -569,13 +569,7 @@ Render2D.clipLineCore = function(a, b, c) {
 
 Render2D.clipLine = function(homog) {
     // transformation to canvas coordinates
-    var m = csport.drawingstate.matrix;
-    var mInverse = List.inverse(List.realMatrix([
-        [m.a, -m.b, m.tx],
-        [m.c, -m.d, -m.ty],
-        [0, 0, 1]
-    ]));
-    var n = List.normalizeMax(List.productMV(List.transpose(mInverse), homog));
+    var n = List.normalizeMax(List.productMV(List.transpose(csport.toMat()), homog));
     var a = n.value[0].value.real;
     var b = n.value[1].value.real;
     var c = n.value[2].value.real;
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -547,16 +547,18 @@ Render2D.drawline = function(homog) {
         y: (-c - a * xMin) / b
     });
 
-    var toUserSpace = function (p) {
-        var q = List.productMV(mInverse, List.realVector([p.x, p.y, 1]));
-        return {
-            x: q.value[0].value.real / q.value[2].value.real,
-            y: q.value[1].value.real / q.value[2].value.real
-        };
-    };
+    if (erg.length === 2 && Render2D.lsize >= 0.01) {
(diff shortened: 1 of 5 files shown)


src/js/libcs/Render2D.js
```diff
@@ -499,60 +535,68 @@ Render2D.drawpoint = function(pt) {
...
-    };
+Render2D.clipLine = function(homog) {
+    // transformation to canvas coordinates
+    var m = csport.drawingstate.matrix;
+    var mInverse = List.inverse(List.realMatrix([
+        [m.a, -m.b, m.tx],
+        [m.c, -m.d, -m.ty],
+        [0, 0, 1]
+    ]));
+    var n = List.normalizeMax(List.productMV(List.transpose(mInverse), homog));
```
gagern: Compare this to [my `csport.toMat`](https://github.com/gagern/CindyJS/blob/7198d9d0f76080528c91006a9d05022badf6408a/src/js/libgeo/GeoState.js#L87-L95) from #280. Perhaps you want to include that here? If so, please make sure to use the same source code location, so that git will recognize the two additions as equivalent and not require manual conflict resolution.

src/js/libcs/Render2D.js
```diff
@@ -499,60 +535,68 @@ Render2D.drawpoint = function(pt) {
...
-        l[2] * b2[0] - l[0] * b2[2],
-        l[0] * b2[1] - l[1] * b2[0]
-    ];
-
-    var pt1 = {
-        x: erg1[0] / erg1[2],
-        y: erg1[1] / erg1[2]
+Render2D.clipLineCore = function(a, b, c) {
+    // clip to canvas boundary (up to line size)
+    var margin = Math.sqrt(0.5) * Render2D.lsize;
```
gagern: I'd like to see `Math.sqrt(0.5)` as a constant somewhere so we don't have to re-compute it every time we run this code.

src/js/libcs/Render2D.js
```diff
@@ -373,6 +373,32 @@ Render2D.arrowShapes = {
     },
 };
 
+Render2D.onSegment = function(startpos, endpos, pt) {
+    var det = List.det(List.realMatrix([
+        [startpos.x, startpos.y, 1],
+        [endpos.x, endpos.y, 1],
+        [pt.x, pt.y, 1]
+    ]));
+    var onLine = CSNumber._helper.isAlmostEqual(det, CSNumber.zero, CSNumber.epsbig);
```
gagern: Do we need this check? The way I see it, all use cases of this function will ensure collinearity of points, so the relative positions are the only important thing.

src/js/libcs/Render2D.js
```diff
@@ -373,6 +373,32 @@ Render2D.arrowShapes = {
...
+Render2D.clipSegment = function(pt1, pt2) {
+    var clipPoints = Render2D.clipLineCore(pt1.y - pt2.y, pt2.x - pt1.x, pt1.x * pt2.y - pt2.x * pt1.y);
+    if (clipPoints.length !== 2) return [];
+    var res = [];
+    if (Render2D.onSegment(clipPoints[0], clipPoints[1], pt1)) res.push(pt1);
+    if (Render2D.onSegment(clipPoints[0], clipPoints[1], pt2)) res.push(pt2);
+    if (Render2D.onSegment(pt1, pt2, clipPoints[0])) res.push(clipPoints[0]);
+    if (Render2D.onSegment(pt1, pt2, clipPoints[1])) res.push(clipPoints[1]);
+    return res;
+};
```
gagern: I'm slightly worried about situations where endpoints and clippoints are *really* close, perhaps even equal. Then you might get more than 2 elements in that array. Can we find a method to avoid that? I guess I'll experiment a bit.

gagern:
On the whole I like the shape this is taking. I think it's reasonably easy to understand now, and should be fairly robust as well.

Some of my comments may help improve performance. And in kranich#2 I've suggested some code to hopefully further improve performance. So I'll wait for the comments to be addressed / answered and my code to be reviewed, then I think we should be ready to go.

kranich:
@gagern I have incorporated your changes and suggestions. Please review.

src/js/libcs/Render2D.js
```diff
@@ -569,13 +569,7 @@ Render2D.clipLineCore = function(a, b, c) {
...
 Render2D.clipLine = function(homog) {
     // transformation to canvas coordinates
-    var m = csport.drawingstate.matrix;
-    var mInverse = List.inverse(List.realMatrix([
-        [m.a, -m.b, m.tx],
-        [m.c, -m.d, -m.ty],
-        [0, 0, 1]
-    ]));
-    var n = List.normalizeMax(List.productMV(List.transpose(mInverse), homog));
+    var n = List.normalizeMax(List.productMV(List.transpose(csport.toMat()), homog));
```
gagern: Minor simplification: one could use `productVM` in order to avoid the `transpose` step.

PR #344, opened 2016-05-19
Minor simplification

kranich:
Follow up of #343.

> Minor simplification: one could use productVM in order to avoid the transpose step.

@gagern Thanks for the suggestion.


kranich:
@gagern Again, this makes the complete graph example noticeably faster!

PR #345, opened 2016-05-19
Consistent text sizes using defaultAppearance

gagern:
This is yet another attempt to address #49, taking the direction discussed by the dev team some time ago:

* The CindyJS default remains at 20
* Cinderella exports `defaultAppearance: { textsize: 12, … }` (already does so for some time now)
* This defaultAppearance affects both `drawtext` and geometric elements, so that the default exported by Cinderella is enough to provide full Cinderella compatibility

This is a breaking change, for two reasons:

1. The default size of point labels is now 20, as opposed to the hard-wired default of 12 we had before.
2. The setting exported from Cinderella is now honored for the first time, causing widgets created using Cinderella but edited subsequently to look different than before.

I had a look at all the examples which either use `drawtext` or labels. As far as I can tell, many are affected by the increased point labels, but none look considerably worse due to this change. So I modified none of the examples contained in the repository.

### DIFF
--- a/examples/textsize.html
+++ b/examples/textsize.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Text sizes</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csdraw" type="text/x-cindyscript">
+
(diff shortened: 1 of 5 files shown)


strobelm:
Also fine, thanks @gagern

Issue #322, opened 2016-04-12
Rename createCindy to CindyJS


gagern:
It is customary that a package exports a single symbol, which is named as the package itself and which serves as the namespace containing all their public API of the module. In CindyJS the names are different: the package called `CindyJS` exports a symbol `createCindy`. It would be good to rename `createCindy` to `CindyJS`, while maintaining the former for a while for the sake of backwards compatibility. Probably with a deprecation notice to the web developer console.

strobelm:
Resolved by #348.

PR #348, opened 2016-05-20
Rename createCindy to CindyJS

gagern:
This implements #322. It's mostly a bulk rename, but I checked the diff of that rename manually, to make sure it doesn't do anything unintended. Nevertheless, it doesn't hurt if a second person has a look at this, too.

This is not an incompatible change, since the `createCindy` name is preserved for the time being. It's deprecated, though, and should be removed once we can be reasonably sure that all users of that old name have made the transition. Of course we'll need to adapt Cinderella first, to make sure that we won't be creating new uses of this deprecated name.

Some outstanding pull requests and other work in progress may be using the old name. Hopefully the forbidden pattern I introduced into the build system will spot such uses and pervent us from merging the branches in question until they have been adjusted.

The file name `ref/createCindy.md` was preserved for now. In the long run, we might want to change that to something better, but at the moment I don't have a good idea for this yet. The current name is as good as all the other bad ideas I had, so I'll keep it for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ and are accompanied by a suitable test case or demonstrating example
 
 ## Documentation
 
-[The `CindyJS` documentation]
+[The CindyJS API documentation]
 (https://github.com/CindyJS/CindyJS/blob/master/ref/createCindy.md)
 describes how to create a widget on an HTML page using this framework.
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ and are accompanied by a suitable test case or demonstrating example
 
 ## Documentation
 
-[The `createCindy` documentation]
+[The `CindyJS` documentation]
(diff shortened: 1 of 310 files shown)


gagern:
Rebased and adjusted to take recent merges (particularly those involving CindyGL) into account.

strobelm:
@gagern the PR got some merge conflicts due to merges in the meantime (sorry for the delay in reviewing). Could you just rerun the renaming process? That should be faster than resolving the conflicts by hand.

gagern:
@strobelm: I no longer had the bulk rename commands in my bash history, so I ended up manually reviewing the commit conflicts and relying on the forbidden patterns to spot any other use cases of the old name. Hope I caught them all.

strobelm:
I clicked through the examples by hand, all seem to run fine. Merged.

PR #349, opened 2016-05-23
Use latest closure compiler (v20160619)

kranich:
Among the benefits are ES6 library polyfills that improve transpilation of ES6 to ES5.

See https://github.com/google/closure-compiler/wiki/Releases#may-17-2016-v20160517.


gagern:
I won't have time to try examples with this before wednesday. If there is any urgent need for this, I'd be willing to merge as is, assuming you did some testing. Otherwise just wait.

kranich:
Meanwhile, Closure Compiler v20160619 has been released.

kranich:
I have tested some randomly examples of core CindyJS, Cindy3D and CindyGL.

PR #350, opened 2016-05-23
comparison of geometric objects

strobelm:
Currently we can't compare geometric objects in eval_helper.equals which leads to bugs in some use cases like list operations: 
```
allelements()--[H,K,L,p];
```
This pull request adds comparison by type and name which should be sufficient (or am i missing something?).

### DIFF
--- a/examples/143_assoc_mult_compare_geoops.html
+++ b/examples/143_assoc_mult_compare_geoops.html
@@ -1,0 +1,95 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>AssocMult.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csdraw" type="text/x-cindyscript">
+els=allelements()--[H,K,L,p];
+println([H,K,l,p]--[K,p]);
+
+println(sort([p,K,L,H]));
+
+n=length(els);
+i=round((n-4)*|H,L|/|H,K|)+5;
+repeat(n,j,
+  (els_j).visible=(j<i)
+);
+
+;
+</script>
+
+    <script type="text/javascript">
+createCindy({
+  scripts: "cs*",
+  defaultAppearance: {
+    dimDependent: 0.7,
+    fontFamily: "sans-serif",
+    lineSize: 1,
+    pointSize: 5.0,
+    textsize: 12.0
+  },
+  angleUnit: "°",
+  geometry: [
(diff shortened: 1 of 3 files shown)


gagern:
Geometric elements have by-reference semantics. They are the same value iff they are the same object, i.e. `v0.value === v1.value`.

strobelm:
One of the few occasions where object comparison really works, thanks @gagern! Update the branch.

gagern:
Now you're just comparing the type!

By the way, what's the state of affairs with the sorting operations? Do we sort elements in gslp order?

strobelm:
Oops, actually it wasn't that late. Fixed now.
I added lexicographical sorting for gslp elements (which is what Cinderella does).

gagern:
I like the core code changes. I'm slightly unhappy with the path of the example. Any reason it's in that `bugs` subdirectory? Do you want to move it out of that? Or was the file included by accident, since at some point you had `bugs` suggested for inclusion in `.gitignore`? Is there still a bug here somewhere to motivate that name?

strobelm:
Well it started as example of a bug (points disappeared due to the lack
of "--" for gslp elements). Since it shows how to reveal objects one
after another and the -- operation it's probably better off in examples.
I've moved it.

Am 24.05.2016 um 15:05 schrieb Martin von Gagern:
> I like the core code changes. I'm slightly unhappy with the path of the
> example. Any reason it's in that |bugs| subdirectory? Do you want to
> move it out of that? Or was the file included by accident, since at some
> point you had |bugs| suggested for inclusion in |.gitignore|? Is there
> still a bug here somewhere to motivate that name?
> 
> —
> You are receiving this because you authored the thread.
> <https://github.com/CindyJS/CindyJS/pull/350#issuecomment-221262130>
> 

-- 
Michael Strobel, M.Sc.

gagern:
If you move an example to a different directory nesting depth, you have to adust the relative path to `Cindy.js`.

strobelm:
quite a lot of silly mistakes for one simple pr, thanks @gagern again.

Am 24.05.2016 um 15:24 schrieb Martin von Gagern:
> If you move an example to a different directory nesting depth, you have
> to adust the relative path to |Cindy.js|.
> 
> —
> You are receiving this because you authored the thread.
> <https://github.com/CindyJS/CindyJS/pull/350#issuecomment-221266994>
> 

-- 
Michael Strobel, M.Sc.

gagern:
Thanks for identifying and resolving the core issue here, @strobelm!

kortenkamp:
I am not sure that Cinderella is sorting elements in the GSLP. Actually I am sure it does not.

strobelm:
I think there is small misunderstanding, i'm talking about sorting lists
of geometric objects (which is now implemented) and not the gslp itself.

Am 24.05.2016 um 15:57 schrieb kortenkamp:
> I am not sure that Cinderella is sorting elements in the GSLP. Actually
> I am sure it does not.
> 
> —
> <https://github.com/CindyJS/CindyJS/pull/350#issuecomment-221278444>
> 

-- 
Michael Strobel, M.Sc.

gagern:
And I was originally thinking about sorting lists of geometric elements by taking the same order the have in the gslp. That would be the way I'd have implemented this if working from scratch. But Michael is right: Cinderella uses a comparison of IDs, so CindyJS should do the same for the sake of compatibility.

Issue #321, opened 2016-04-12
Lexical scoping


gagern:
Cinderella is currently making progress towards lexical scoping, and CindyJS should follow suit before the 1.0.0 release.

* [ ] lexical scoping for function arguments
* [ ] lexical scoping for constructs with a (user-defined or implicit) running variable, e.g. `repeat`
* [ ] introduction of `block(‹var1›, ‹var2›, …, ‹varN›, ‹code›)` for lexically scoped local variables
* [ ] closures

PR #352, opened 2016-05-24
CindyGL: reload images to GPU that were modified by canvas(...) or clearimage()

montaga:
This commit adds a flag `copiedToCindyGL` to each image/canvas which indicates weather the current corresponding image/canvas was copied to a WebGL-Texture. It is set to false once the image/canvas has been changed. It is set to true once it was copied to the WebGL-Texture.

An example of usage can be seen in `26_readcanvasimage.html`.

Please review this commit since it causes modifications in `src/js/libcs/OpDrawing.js`.
I am not sure whether it is a good style to communicate to a plugin via a flag.

Alternatively, - in a similar manner as the `whenReady`-function - we could add a `whenModified`-function to the imageobeject returned by the `getImage`-API-function that adds functions that are supposed to be called once they are modified.
But this would require some more elaborate internal refactoring: Instead of accessing plain images from the `image`-Array or accessing canvas elements by `document.getElementById`(see implementation of `evaluator.canvas` in  `src/js/libcs/OpDrawing.js`), they must be embedded in the new image data structure in order to communicate to the list of functions that are supposed to be called.


gagern:
I'm slightly worried by the fact that this is so plugin-specific. I'd prefer if core code wouldn't contain anything specific to a single plugin. For this reason, the `whenModified` code sounds more appropriate. Another alternative would be some `generation` field, which is simply some integer that gets incremented each time the image is modified. That way, you could just store what's the last generation of the image imported into the texture, and re-load the texture if that generation is different.

No matter which of these two alternatives we decide on, we might want to implement the same for (user-defined) functions. So that you'll notice when to re-compile code because one of the functions you were using has been changed its implementation.

gagern:
From a stylistic viewpoint, I'd like to understand why this change explicitly *removes* trailing newlines to files. It is my understanding that it is common convention to end source files with a newline. One benefit is that one can concatenate files without concatenating lines. The JS mode of my editor will append a trailing newline if I edit some file. So if I don't want to commit this (since it's not part of the core of some change), I have to manually make sure to remove it, or at least not stage it for git. Is there a pressing reason why you want no trailing newlines in your files, @montaga?

montaga:
@gagern Thank you for your review!

I am going to add some `generation` field to the canvas-object, which might be also used by other plugins.

The idea with a `generation` field for user-defined functions sounds like a smart idea as well in order to avoid calling the `forcerecompile` command. The only problem that I see right now is that the user might define functions within the `csdraw` environment without changing function bodies. Semantically they at the wrong place, but it would be a showstopper if this would cause performance bottle necks, because they would cause a slow recompilation for every rendered frame. (And if I take a look at our examples, it seems quite intuitive have the user defined functions there). We would need a fast way to check whether functions bodies are changed.

Regarding the removed trailing newlines: It was no conscious decision; Actually I did not care at all. Probably it was caused by my editor or by `js-beautify`.

gagern:
I've been thinking about functions getting defined in the draw script as well. I guess it would make sense to extend the operator calling convention so that in addition to the list of arguments and the dictionary of modifiers, the actual node from the abstract syntax tree gets passed as a third argument. Then we could keep track of that syntax node in the data structure of the function, and increment the generation only if the function gets defined by a different instance of the definition operation. (Once we get closures in the course of #321, we should take the surrounding closure into account as well: the same definition executed inside a different closre would lead to an incremented generation number, too.) If you want me to, I can address these things, otherwise I'll be happy to review any changes in this direction.

The newlines were definitely not stripped by `make beautify`, since [we have a rule](https://github.com/CindyJS/CindyJS/blob/c82d8bc18a2b67c6b10420559281730d21af1ff1/Administration/beautify.conf#L4) to ensure files end with newlines. But that command is only applied to the core code, not to the plugins, so CindyGL isn't affected. If you apply `js-beautify` yourself, with no configuration file, then this might indeed be the cause of the problem, since by default it apparently does strip newlines 😦. If you beautify the cindygl plugin on a regular basis, perhaps we should include that beautification as part of [the *beautify* build step](https://github.com/CindyJS/CindyJS/blob/c82d8bc18a2b67c6b10420559281730d21af1ff1/make/build.js#L109-L121)?

I'm not sure whether it would make sense to beautify all plugins. On the one hand, doing so and using the same set of settings throughout would make the whole codebase more consistent. On the other hand, I'd like to think that I'm using a fairly consistent coding style within Cindy3D, and sometimes I'm deliberately deviating from what js-beautify would do, e.g. when initializing several coordinate variables for a single vector in a single line, like `let x=…, y=…, z=…` which I consider a single operation.

montaga:
As I am not aware yet of the core code that handles the user defined functions, I would be glad, if you could add the `generation` field to them. After that I am going to modify the CindyGL plugin such that it will take the changes of the generation into account.

The last commit of this pull request adds a `generation` field to the image object, which is a counter that increased whenever the encoded image is changed. It is used to trigger CindyGL.

With #355 the beautification of CindyGL becomes  part of the beautify build step.

PR #354, opened 2016-05-26
AST performance

gagern:
There is a noticable drop in rendering performance of some examples in some environments. Most notably an example of the @richter-gebert and I have bisected this to [somewhere in the new parser implementation](https://github.com/CindyJS/CindyJS/compare/407d506cb0be5cccf65de17bf7c0483208653a29...8842ce842f6976b28b6b7f180d3757bffc894a4d). Which is really strange, since the change of parser should not affect runtime behavior.

But apparently the internal optimization has problems with the way the objects are generated by the new parser. So this pull request changes the representation to re-create each object in the postprocessing step using an object literal. (The handling of hidden classes *might* be the key here, assuming that's what Nitro is using.)

It also simplifies the `evaluate` function, since the additional `evaluate` call for `variable` type objects (needed to properly handle `name := expression` which requires re-evaluation when used) used to contribute part of the performance penalty.


strobelm:
Very very interesting, the performance increased notably in some examples (like 12_L-System.html)! Thanks @gagern for the super quick solution, i think @richter-gebert will leap for joy ;-).

PR #355, opened 2016-05-27
Now `make beautify` will also beautify cindyGL code

montaga:
The only semantic change is within `make/build.js`. Not any file of `plugins/cindygl/src/js/*` was changed semantically.

In order to avoid merge conflicts, all other pull requests regarding CindyGL should be merged first.

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -333,6 +333,14 @@ module.exports = function build(settings, task) {
         "ShaderProgram"
     ];
 
+    var cgl_mods_srcs = cgl_mods.map(function(name) {
+        return "plugins/cindygl/src/js/" + name + ".js";
+    });
+
+    var cgl_mods_from_c3d_srcs = cgl_mods_from_c3d.map(function(name) {
+        return "plugins/cindy3d/src/js/" + name + ".js";
+    });
+
     task("cglres", [], function() {
         cgl_str_res.forEach(this.input, this);
         this.node(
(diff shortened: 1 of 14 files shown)


gagern:
> In order to avoid merge conflicts, all other pull requests regarding CindyGL should be merged first.

Or users should merge resp. rebase with [the `-X ignore-all-space` setting](https://git-scm.com/docs/git-merge#_merge_strategies). Followed by a re-run of the beautification. My experience with this approach is qute good.

The failing Travis build is a problem, though.

gagern:
I guess the problem with `beautified` is due to the order of variable assignments versus task definitions. I'm windering whether it would be possible to evaluate the task definitions a bit later, so that all the variables are in place by then, no matter the order.

make/build.js
```diff
@@ -391,6 +377,21 @@ module.exports = function build(settings, task) {
     });
     
     //////////////////////////////////////////////////////////////////////
+    // Run js-beautify for consistent coding style
+    //////////////////////////////////////////////////////////////////////
+
+    var beautify_args = [
+        "--replace",
+        "--config", "Administration/beautify.conf",
+        (src.ours).concat(cgl_mods.map(function(name) { return 'plugins/cindygl/src/js/' + name + '.js';})).filter(function(name) { return !/^build\//.test(name); }),
```
gagern: I find this line very long and hard to read. Can you distribute the information accross several lines? Those module names occur in the CindyGL build tasks, too, so perhaps having them in some resonable location makes sense anyway.

montaga:
Thank you for fixing the issue in #356.
I took your comment regarding this long and unreadable line into account.

PR #360, opened 2016-05-31
Convex hull

nasay:
This is a PR with an implementation of Convex hull on JS.
I added three examples: 
- two where you can compare JS and JAVA version (`example/test-convex-hull.html` for 12 points and `example/test-convex-hull-3.html` for more than 5000 points);
- `example/test-convex-hull-2.html` where you can see that it still moves smoothly.

There is unit-test file `tests/quick_hull_tests.js` with tests for vector operations and some checks for convex hull.
I did change API of Face and HalfEdge: I removed most of getters and setters and left only those that has some logic inside.

I also have <a href="https://github.com/nasay/CindyJS/tree/HEOS-1">branch</a> with some optimisations, where building the example with 5000 points requires 60-90 milliseconds.


gagern:
Haven't looked at the code yet. But I'd say we want a `git rebase --autosquash -i` to squash the fixup messages, and of course we want Travis happy, which means updating the style of the example, e.g. based on [the template](https://github.com/CindyJS/CindyJS/blob/master/examples/template.html).

PR #362, opened 2016-06-02
Added workaround for iPhone/iPad.

montaga:
This is a more a fix than a proper solution. In order to deal properly with this situation, one has to build a little toy example for writing to float textures and finding out whether it fails.


gagern:
To document this for posterity: what exactly is the problem here? They claim to support the extension but compiling a shader for it fails nonetheless? With what error message?

montaga:
The problem is that writing to framebuffers with an higher precision is not possible yet the testing iPads. On the testing device the following error is thrown in `plugins/cindy3d/src/js/ShaderProgram.js` when one tries to link a shader having an  high precision framebuffer object attached:
`Error validating shader: Validation Failed: Current draw framebuffer is invalid.`

PR #363, opened 2016-06-07
CindyGL: Added modifiers for imagergb

montaga:
Now there are three new modifiers for `imagergb[a]`:

-`interpolate`: An Boolean that indicates whether a read access should be interpolated by `GL_LINEAR`. If it is set to false, `GL_NEAREST` is used. Per default it is set to `true`.

-`mipmap`: A Boolean to enable `GL_LINEAR_MIPMAP_LINEAR` insted of `GL_NEAREST_MIPMAP_NEAREST`. If set to true, also a ` gl.generateMipmap(gl.TEXTURE_2D);` is called after each texture update. As default, it is set to `false`.

-`repeat`: A Boolean that indicates weather the texture is supposed to be interpreted as periodic pattern. (p1 (o) group). If it is used in combination with mipmaping then a modified 4-tap trick from https://0fps.net/2013/07/09/texture-atlases-wrapping-and-mip-mapping/ will be used(In CindyGL we consider a 4 map atlas of the Torus). Usually it should be set for any periodic rendering even if the texture is accessed after applying a mod-operator, because then CindyGL will avoid seams. Default: `false`.

For technical reasons, all we CindyGL handles only modifiers that are uniform, i.e. independent of #.


gagern:
I like these! How about flipping the image while repeating, i.e. allowing the user to choose symmetry groups pm (two orientations) and pmm in addition to p1? I haven't looked at the internal code, but I would hope that such settings should map pretty directly to [`gl.texParameter`](https://www.khronos.org/opengles/sdk/docs/man/xhtml/glTexParameter.xml) calls with `gl.TEXTURE_WRAP_S` resp. `gl.TEXTURE_WRAP_T` set to `gl.MIRRORED_REPEAT`.

PR #366, opened 2016-06-07
Implement field accessors for the drawtrace property

gagern:
The getter was broken, since it did access the wrong property name, and since it failed to convert unset values to boolean false. The setter is completely new, and makes use of a refactored piece of tracing setup code. Disabling and re-enabling tracing will clean all existing traces. This is consistent with what Cinderella does.


strobelm:
Looks fine, thanks for fixing this.

Issue #14, opened 2015-10-08
Support Text in geometry


gagern:
We should support Text as a geometric element. In its easiest form, the renderer could translate this directly to a call to drawtext. For added compatibility, we should also support styling text as buttons, with all the machinery associated with this. And there is also `EditableText` to provide a text input field.

* [x] plain text printed
* [x] value substitution for measurements
* [x] docking to points
* [x] docking to window edges and corners
* [x] newlines in static text
* [x] button styling, both pressed and non-pressed
* [x] push button triggers script
* [x] toggle button has accessible state
* [x] editable text shows text field
* [x] text of editable text field accessible from CindyScript
* [x] editable text actually is editable
* [x] reasonable input focus handling

kortenkamp:
This feature is a showstopper for most of the interesting examples. It is urgently necessary to implement this, so I comment here to push it on top of the to-do list…

kortenkamp:
Just a meta-question – who is able to edit and assign issues?

kranich:
>Just a meta-question – who is able to edit and assign issues?

People with Write, Admin, or Owner permissions, see

https://help.github.com/articles/repository-permission-levels-for-an-organization/

kortenkamp:
ok… thanks to Martin I am now able to assign people to this task - thank you! 

@gagern - if you don't like this assignment, please find somebody else… I  really think this is urgent.

gagern:
@kortenkamp Since I already did work on the first few points here for #225, it's only natural that I'm the one to continue that work.

gagern:
Finally done.

PR #368, opened 2016-06-09
Dynamic resize

gagern:
This is one step along the way towards buttons (#14) since it introduces a `<div>` around our `<canvas>` which can then be used to put in HTML elements. This requires some work to get the size of the `<div>` and the `<canvas>` synchronized properly. That felt like a good opportunity to make that synchronization work not only at initialization time, but also dynamically when the size gets updated.

This is a breaking change, since it is quite possible to specify CSS style sheets where the new DOM tree will render differently than the old one. Some care should be taken when updating to a version including this change, but sane content should be working all right or, in the worst case, quite easy to fix.

### DIFF
--- a/examples/00_Playground.html
+++ b/examples/00_Playground.html
@@ -20,7 +20,7 @@
 
 
 
-        <canvas  id="CSCanvas" width=500 height=500  style="border:2px solid #000000"></canvas>
+        <div  id="CSCanvas" style="width:500px; height:500px; border:2px solid #000000"></div>
 
         <script type="text/javascript">
 
--- a/examples/01_sunflower.html
+++ b/examples/01_sunflower.html
@@ -25,7 +25,7 @@
 
 
 
-        <canvas  id="CSCanvas" width=500 height=500  style="border:2px solid #000000"></canvas>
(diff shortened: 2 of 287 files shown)


montaga:
Maybe we should forbid the pattern
`<canvas id="CSCanvas"`
in favor of the preferred `div`-tag?

The FakeCanvas in '/tests/GeoFuncs_tests.js' needs an attribute `setTransform`.

montaga:
As far as I understood the changes, they are fine.

But before merging, we should make `node ref/js/runtests.js` work again. In order to fix the testing environment I know to little about how it is built up; so it would mean a lot of afford for me.
@gagern Maybe you could make these changes.

gagern:
@montaga thanks for the review. I've addressed the test suite, adjusted the examples and added the forbidden pattern as you suggested. Waiting for OK from Travis, but local `make alltests` succeeded.

gagern:
Good thing I waited for Travis, since our master had a few examples missing in my local branch. Rebased and re-converted.

gagern:
@montaga Now that Travis is happy, you are happy with the core code change and I addressed the suggestions, I took the liberty of merging this myself. Thanks for looking into this.

PR #369, opened 2016-06-13
Buttons

gagern:
This implements buttons, a much-needed feature and part of #14. The buttons have type `PushButton` or `Switch` in accordance with Cinderella merge request !110. Currently buttons don't cooperate with the KaTeX plugin yet. I'd include that in a subsequent pull request.

The approach is somewhat hackish, since it re-uses part of the `drawtext` pipeline but takes a different branch at the end of that code path. Feel free to suggest other alternatives which resolve this in a cleaner fashion without too much code duplication.

The style is handled by a new stylesheet, `CindyJS.css` built from [`CindyJS.scss`](https://github.com/gagern/CindyJS/blob/buttons/src/scss/CindyJS.scss) using [SASS](http://sass-lang.com/). It provides settings which I consider crucial for the operation of CindyJS widgets. All classes are prefixed by `CindyJS-`, and all effects are restricted to CindyJS widget containers. So it should be safe to include this CSS into all documents alongside `Cindy.js`, and we should advocate doing so at least if features that need it might be used (like buttons). This differs from the existing [`cindy.css`](https://github.com/gagern/CindyJS/blob/buttons/css/cindy.css) which does pretty non-standard settings like forcing the position of all `canvas` elements to be absolute. That file is used in some examples, but I'd vote to get rid of that file an all its uses in order to avoid confusion, and I'd definitely not request third party sites to include that file. If you prefer a different default style for these buttons, feel free to suggest changes to `CindyJS.scss`.

I'm making use of the newly found fact (or perhaps newly added feature?) that I can assign an issue to multiple people, and hope that one of the assignees finds time to review this.


kranich:
> I'm making use of the newly found fact (or perhaps newly added feature?) that I can assign an issue to multiple people, and hope that one of the assignees finds time to review this.

@gagern Newly added feature, see https://github.com/blog/2178-multiple-assignees-on-issues-and-pull-requests.

We have a build failure (missing beautification).

gagern:
Once more Travis saved the day. This time because the fact that it complained despite my local successful tests made me notice that I had forgotten to push my latest changes.

kranich:
@gagern Everything looks OK.

I found that Firefox's inspector indicates that whenever I move point A or click one of the buttons in examples/Buttons1.html, the class and font of the surrounding div of every button are updated. I do not understand this based on what I have seen in the code. The only line where the class is set to 'CindyJS-button' seems to be [this](https://github.com/CindyJS/CindyJS/pull/369/files#diff-29234c1c634943e10e04a1293e7363e7R2622). Should we care about this? If so, do you understand this behaviour?

kranich:
By the way, I have found out that one can easily check out pull requests as follows:

```
git fetch CindyJS refs/pull/369/head:buttons
git checkout buttons
```

where `CindyJS` is the remote pointing to our GitHub repository. Maybe you also find this useful.

gagern:
@kranich I tend to write `git fetch CindyJS refs/pull/369/merge && git checkout FETCH_HEAD` when reviewing, since in that case I want the `merge`, not the `head`, in order to test integration with the most recent changes to `master`. I found this looking at the Travis clone command sequence, and think it's a *very* useful and rather poorly documented feature. Hard to discover if you're not explicitely looking for it.

gagern:
The font update makes a lot of sense, due to [`html.style.font = csctx.font`](https://github.com/gagern/CindyJS/blob/45be47ac1e293838334915ffe0c46ad26ea82cf5/src/js/libcs/OpDrawing.js#L916) being executed on every rendering of the scene. Perhaps we should cache there, to only set that in the rare case of an actual font change. I can see about implementing that, and for the position as well. Perhaps that takes care of the class name for some magical reason, too. If not, I have no clue where this comes from, so I wouldn't care too much.

kranich:
@gagern jshint complains that `libgeo/GeoRender.js: line 210, col 16, 'outerWidth' is not defined`.

gagern:
I'm having some pretty serious problems here with the loading of resources. Particularly on Firefox. The core of the problem is that I refer to `CindyJS.css` as an external resource, i.e. using a `<link>` tag. That means that the CSS isn't ready yet when the `DOMContentLoaded` event fires. Since gagern@1f85cb4b485dc537a3fea4e22ada0e577a215542 an `updateCindy` call after the CSS has been applied will rectify the situation. But so far I haven't managed to reliably trigger that call. I've tried adding a `load` event handler to the `document`, or adding one to the `<link>` element. I even tried omitting the `<link>` from the HTML file, and adding it programmatically. The event gets triggered, but apparently the fact that the CSS has been loaded doesn't mean that it has been applied to the document. Presumably the [reflow queue](http://www.phpied.com/rendering-repaint-reflowrelayout-restyle/) has to catch up with those changes first. And I don't know how to capture when it does. Any suggestions?

kranich:
@gagern I see the problem. Have tried to investigate solutions for 2 hours now, but no success.

gagern:
I tried to reproduce this using [these file](https://gist.github.com/gagern/48ae8842f04efa73acb79da8324f755b), but I can't observe the problem there.

I've also started working with a negative margin to counter the added padding and boundary, but I still get incorrect results. Will investigate further.

gagern:
OK, I now have [a reproducing example](https://gist.github.com/gagern/48ae8842f04efa73acb79da8324f755b) and [posted this Stack Overflow question](http://stackoverflow.com/q/37813186/1468366) asking for help with this issue. I wouldn't rule out a browser bug as a possible cause for the observed behavior.

gagern:
This looks better now, at least on my Firefox. But it might impact portability with some older browsers. What versions to we intend to support, or Internet Explorer in particular? Can someone test this with the oldest IE we want to officially support? Do we want to add vendor-prefixed versions of [the `transform` property](http://caniuse.com/#feat=transforms2d)? I guess I should add another test case to test for centered and right-aligned buttons, since those are most likely to cause problems with this new implementation.

gagern:
I just found that as far as I can tell it's not possible to specify the alignment of a button (or other Text element) in Cinderella as anything other than left (or perhaps leading edge for rright-to-left writing systems?) So although I think that controlling text and button alignment would be a good thing t have, I don't think that having it is relevant for this issue here. Which means that we don't need support for the `transform` CSS property either, at least as far as content exported from Cinderella is concerned.

kortenkamp:
It is not necessary right now to support alignment, but I am happy to add alignment in Cinderella if it is available in CindyJS.

gagern:
OK, now we have a second test case which tests alignment, and we also have buttons using KaTeX-rendered labels, with a demo of that included in that second test case as well.

gagern:
I've just noticed that in the Buttons2 example, some parts of some buttons cannot be clicked because a button in a row below that one covers it with its invisible (and huge) upper portion which I use to get baseline alignment implemented. I need to find a fix for this.

gagern:
Ready for review - again.

src/scss/CindyJS.scss
```diff
@@ -0,0 +1,69 @@
...
+            @include naked;
+            display: inline;
+            position: static;
+            vertical-align: baseline;
+            visibility: hidden;
+            margin-right: -1px;
+        }
+
+        input[type=checkbox] {
+            position: absolute;
```
kranich: @gagern I would prefer `display: none` instead of this and the following line.

src/js/libcs/Render2D.js
```diff
@@ -225,6 +225,7 @@ Render2D.modifHandlers = {
     "align": function(v) {
         if (v.ctype === "string") {
             var s = v.value;
+            // TODO: Use values suitable for csctx.textAlign here
```
kranich: @gagern Maybe you want to open an issue in order to track this?

kranich:
@gagern In examples/Buttons1.html, I no longer see the green point on the right change its y-value when I click Button 4.

kranich:
At least in Firefox, the center alignment is off by 1 or 2 pixels to the right.

buttons2

kortenkamp:
@kranich it could still be that lines are off by one pixel - it's not clear to me how one can determine whether lines or buttons are correct.

kranich:
Please do not get me wrong, @gagern, I really appreciate your efforts. But all in all, this feels extremely hackish. I fear that it might easily break in the future or not be cross-browser compatible.

I think it might be cleaner to draw buttons directly onto the canvas instead of using HTML buttons that are causing so much pain here. What do you think?

src/scss/CindyJS.scss
gagern: I, too, but in my experience a checkbox with `display: none` cannot have the input focus, so it won't work with keyboard navigation using <kbd>⇥</kbd> and <kbd>space</kbd>.

gagern:
I found that the fact that the `<div>`s containing the buttons resp. labels are so much higher than they appear will also break interaction with the underlying canvas: in the Buttons1 example, one can't drag the point *A* into the area above a button.

Yes, this whole mess is incredibly hackish. For buttons it might be cleaner to do this in canvas. For text boxes, however, I fear that we'd cause serious grief for people using input methods. And doing buttons in a different way than text boxes feels weird, too. So I'd like some more discussion before changing directions here.

We *could* change the meaning of button positions, in such a way that buttons get positioned by center or corner or something like that, instead of baseline. That would simplify things a lot, with positioning by top left corner being the easiest. We could try to compensate for this when exporting. Adjusting for line height in the default font should be fairly easy. Adjusting for actual button dimensions would likely be hard in a headless setup, but it might be worth trying nonetheless.

gagern:
@kranich wrote:

> In examples/Buttons1.html, I no longer see the green point on the right change its y-value when I click Button 4.

Did you ever? That was meant as a negative example: In Cinderella that point never changes its position, even while the button is being pressed (i.e. the mouse button is held). CindyJS reproduces this, since for non-checkboxes the `pressed` field is always false.

kranich:
@gagern I thought I did but I cannot reproduce it now, so maybe I remember incorrectly.

src/js/libcs/Render2D.js
gagern: Opened #373 for this.

gagern:
I successfully tested this branch on IE 11 and Edge. I [had to add](http://stackoverflow.com/q/6771258/1468366)

```html
<meta http-equiv="X-UA-Compatible" content="IE=edge">
```

to the example pages to make them display at all in IE 11. Without this, the page looked blank, and the web developer tools showed a message:

    Die angefügte Seite hat den Dokumentmodus 7 als Ziel.
    The attached page targets document mode 7.

Document mode autodetection in IE appears to be [far more complicated](https://technet.microsoft.com/en-us/itpro/internet-explorer/ie11-deploy-guide/deprecated-document-modes) than I'd ever have imagined. Apparently part of the problem was that I tested the page as served from some intranet server. So we should be fine for widgets on public internet sites. I see no need to add that header to all our examples just to make them work in intranets or with strange configurations. But we should remember this, and perhaps document it in the README or some such.

gagern:
At the moment this solution appears to be working reasonably well on all browsers where I tried this. And it's less hackish than it used to be at some points. I'd personally vote for further pursuing this direction, as opposed to investigating a purely canvas-based approach now. That would make it easier to try this in a wide variety of scenarios.

One more thing to consider: we should move the `createtool`-related buttons to Cinderella-compatible positioning soon, too. And personally I think we should aim for consistent styling of buttons. Which would mean that any decision we make here will likely affect those buttons as well. We also decided to switch those buttons over to SVG, and as far as I recall, drawing SVG to canvas comes at a pretty huge hit in terms of quality.

kortenkamp:
When using SVG to draw buttons this will make it impossible to use them in Cinderella – and maybe a proper .png is better than vector graphics?

strobelm:
Although i think this won't be the last word on buttons, it's a good solution for now. Thanks @gagern for all the effort.

PR #370, opened 2016-06-14
Add position to dropped objects

strobelm:
This PR adds the drop position to drag and drop objects.

### DIFF
--- a/examples/141_fileDrop.html
+++ b/examples/141_fileDrop.html
@@ -20,7 +20,7 @@
   if (!isundefined(img), drawimage(A, B, img),
     if (!isundefined(txt), drawtext(A, txt, color->[0, 0.5, 0.75])));
   if (!isundefined(dropped), (
-    drawtext((0, 0), length(dropped) + " files dropped");
+    drawtext((0, 0), length(dropped) + " files dropped" + "at pos " +  dropped_1_5_1 + " " + dropped_1_5_2);
     forall (1..length(dropped), i, (
       drawtext((0, -i), dropped_i_2 + " named " + dropped_i_4);
     ));
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -196,34 +196,38 @@ function setuplisteners(canvas, data) {
         var files = dt.files;
         var dropped = Array(files.length);
         var countDown = files.length;
+        // drop position
+        var pos = List.realVector(csport.to(e.clientX, e.clientY));
+
         Array.prototype.forEach.call(files, function(file, i) {
             var reader = new FileReader();
             if ((/^text\//).test(file.type)) {
                 reader.onload = function() {
                     var value = General.string(reader.result);
-                    oneDone(i, value);
+                    oneDone(i, value, value.ctype, pos);
                 };
                 reader.readAsText(file);
             } else if ((/^image\//).test(file.type)) {
                 reader.onload = function() {
                     var img = new Image();
                     img.onload = function() {
-                        oneDone(i, loadImage(img));
+                        oneDone(i, loadImage(img), undefined, pos);
                     };
                     img.src = reader.result;
                 };
                 reader.readAsDataURL(file);
             } else {
-                oneDone(i, nada);
+                oneDone(i, nada, nada.ctype, pos);
             }
         });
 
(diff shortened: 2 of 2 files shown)


kranich:
@strobelm Looks useful. Maybe in the long run we want a more object-oriented interface for drop event data? `dropped_1_5_1` is not really transparent. Maybe it should become `(dropped_1).pos.x` or similar.

gagern:
The `clientX` and `clientY` coordinates are *not* local to the canvas. I modified your example to `draw(dropped_1_5)` and also included a `<h1>` headline above the widget. The resulting point is decidedly different from the actual drop location. You could probably use [this code](https://github.com/CindyJS/CindyJS/blob/e2158b8795f1929fd83f79588086c4edb7c05243/src/js/Events.js#L118-L121) as a template for how to convert HTML client coordinates to userspace coordinates. Or you could refactor this into a function of its own, making use of duck typing so that anything that contains a `clientX` and `clientY` property may server as the source location.

Since the position is independent of the individual dropped items, it feels somewhat weird to me to repeat that with every single dropped item. Mainly for reasons of model quality; this approach violates the [second normal form](https://en.wikipedia.org/wiki/Second_normal_form) of database normalization. I have to concede that this should have negligible performance impact here. A more tangible consequence of this approach is that you can't have a position without files. [MDN writes](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files):

> If the drag operation had no files, the list is empty.

For example, a user might be dragging some image or selected text or whatever from within the HTML page. We don't really support this kind of thing at the moment (since we don't call `cs_onDrop` if we don't call `oneDrop` at least once), but we might want to add that later on. So I'd feel more comfortable having a separate function `dropPosition` or similar to obtain the position of the most recent drop within the drop handler script.

gagern:
@kranich Named access to the different properties would be nice. But at the moment that would require introducing a new data type, or some similar hack. Once we have dictionary types (as planned), we can use them for this task here, without the need for custom code outside the drop event handler. So I'd rather wait till then. Writing `.x` to access the first element of a vector already works.

strobelm:
Thanks Martin, i'll address the problems asap (i.e. done by 2pm).

Am 14.06.2016 um 10:44 schrieb Martin von Gagern:
> The |clientX| and |clientY| coordinates are /not/ local to the canvas. I
> modified your example to |draw(dropped_1_5)| and also included a |<h1>|
> headline above the widget. The resulting point is decidedly different
> from the actual drop location. You could probably use this code
> <https://github.com/CindyJS/CindyJS/blob/e2158b8795f1929fd83f79588086c4edb7c05243/src/js/Events.js#L118-L121>
> as a template for how to convert HTML client coordinates to userspace
> coordinates. Or you could refactor this into a function of its own,
> making use of duck typing so that anything that contains a |clientX| and
> |clientY| property may server as the source location.
> 
> Since the position is independent of the individual dropped items, it
> feels somewhat weird to me to repeat that with every single dropped
> item. Mainly for reasons of model quality; this approach violates the
> second normal form <https://en.wikipedia.org/wiki/Second_normal_form> of
> database normalization. I have to concede that this should have
> negligible performance impact here. A more tangible consequence of this
> approach is that you can't have a position without files. MDN writes
> <https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/files>:
> 
>     If the drag operation had no files, the list is empty.
> 
> For example, a user might be dragging some image or selected text or
> whatever from within the HTML page. We don't really support this kind of
> thing at the moment (since we don't call |cs_onDrop| if we don't call
> |oneDrop| at least once), but we might want to add that later on. So I'd
> feel more comfortable having a separate function |dropPosition| or
> similar to obtain the position of the most recent drop within the drop
> handler script.
> 
> —
> <https://github.com/CindyJS/CindyJS/pull/370#issuecomment-225817244>, or
> mute the thread
> <https://github.com/notifications/unsubscribe/ANhz4gtQbrZdx3Nc7joey8eHdpyWQ1RNks5qLmnXgaJpZM4I1DO7>.
> 

-- 
Michael Strobel, M.Sc.

gagern:
For reference: follow-up is #371.

PR #371, opened 2016-06-14
reworked drop position handling

strobelm:
Here the hopefully less buggy version.


gagern:
Let's see the position in the example, i.e. draw a point there, to better judge correctness.

Personally I'd expand “droppedpos” to “the dropped position”, which doesn't make sense, since we dropped files *at* a position, we did not drop a position. Do others agree with my view? If so, please adjust that name, too.

strobelm:
I renamed the function to "droppoint()" but i'm open to suggestions. The drop point is now indicated with a green point on the canvas.

PR #378, opened 2016-06-15
Warn developers about node 6.2.1

gagern:
I've analyzed the [node 6.2.1 breakage](https://github.com/CindyJS/CindyJS/pull/374) some more, and decided that it's indeed a bug in their code, to be tracked at nodejs/node#7308. I think (and hope) that their next release will likely fix things. So I don't see much reason to add a README paragraph about problems with 6.2.1. But if developers should happen to use just that version, I'd like to warn them about the cause of their problem. So I added a warning message to the version check. It will not abort the build, since most aspects of the build procedure will work just fine anyway.

### DIFF
--- a/make/check-node-version.js
+++ b/make/check-node-version.js
@@ -6,8 +6,15 @@ v = v.split(".");
 v = v.map(function(s){
   return parseInt(s);
 });
-if (v[0] == 0 && v[1] < 12) {
+var a = v[0], b = v[1], c = v[2];
+if (a == 0 && b < 12) {
   console.error("Node 0.12 or later required. Version " +
                 process.version + " found");
   process.exit(1);
+}
+if (a > 6 || (a == 6 && (b > 2 || (b == 2 && c >= 1)))) {
+  console.error("Node versions starting at 6.2.1 have a bug,");
+  console.error("likely causing the “forbidden” check to fail.");
+  console.error("See https://github.com/nodejs/node/issues/7308");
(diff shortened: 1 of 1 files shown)


addaleax:
Just fyi, the next Node.js v6 release will be 6.3.0 which is [a security release](https://nodejs.org/en/blog/vulnerability/june-2016-security-releases/), so it won’t contain any other changes than those security-related ones and very certainly not a bug fix for this.

kranich:
@addaleax Thanks for letting us know.

kranich:
@gagern So maybe we should print the warning for node version >= 6.2.1 until a fix is released?

gagern:
@kranich Yes. Will update accordingly.

make/check-node-version.js
```diff
@@ -11,3 +11,9 @@ if (v[0] == 0 && v[1] < 12) {
                 process.version + " found");
   process.exit(1);
 }
+if (v[0] == 6 && v[1] == 2 && v[2] == 1) {
```
kranich: This should be `v[0] == 6 && ((v[1] == 2 && v[2] == 1) || v[1] > 2)`.

gagern: You are of course right, I can't expect node to read the text of the warning message and perform the comparison based on its wording instead of the code I wrote…

kranich:
@gagern I still hope that the bug will be fixed in the next major release, but OK.

Issue #320, opened 2016-04-12
Conics by contact pairs


gagern:
It might make sense to add additional geometric operations to define a conic using what I call a contact pair: a point on a conic and a line through that point tangent to the conic. (Is there a more established term for this concept?) Possible cases include:

1. Conic by contact pair and three more points
2. Conic by contact pair and three more lines
3. Conic by two contact pairs and one more point
4. Conic by two contact pairs and one more line

All of these apparently yield unique results, which could also be obtained from a limit process. https://github.com/CindyJS/CindyJS/pull/280#issuecomment-198760620 suggests applications and some code for 3. and the special case of 4. where the line is the line at infinity, thus giving a parabola by two contact pairs. Today I was reminded that I did some work on 1. for [a Math Stack Exchange post](http://math.stackexchange.com/a/429858/35416) some years ago, again for the special case where the contact pair includes the line at infinity, such that the result is a parabola.

I'm unsure how to handle the concept of a contact pair in our code, though. One can represent the point and the line as two distinct elements, and allow arbitrary garbage as output if those two are not incident. While this would probably be acceptable for CindyJS as it is, an interactive construction environment like Cinderella or CindyJS with its createtool additions might want to ensure incidence upon construction. Any better suggestions?

gagern:
At today's team meeting it was suggested to consider pole/polar pairs as generalizations of what I called a contact pair. In the case they do coincide, the two concepts are the same, but a pole/polar pair would be defined even if its elements are not incident. It remains to be seen whether the resulting conics would still be uniquely defined. Experiments in this regard still pending…

elkins0:
Yes, that is definitely the way to go. There are some special pole/polar pairs for conics. Each focus and its corresponding directrix  are such a pair which is in fact how `geoOps.ConicParabolaPL` defines its parabola. As you know, hyperbolas and ellipses can also be defined this way using their foci. For the cases where the point lies on the line, I already have the following implementations:
```javascript
// Given ["A", "a", "B", "C", "D"] where "a" is a line thru "A" and "B", "C", "D" are points on the conic
// produces the conic given the tangency (a tangent point on a line) and three points on the conic
geoOps.ConicBy1t3p = {};
geoOps.ConicBy1t3p.kind = "C";
geoOps.ConicBy1t3p.signature = ["P", "L", "P", "P", "P"];
geoOps.ConicBy1t3p.updatePosition = function(el) {
    var A = csgeo.csnames[(el.args[0])].homog;
[...]

PR #382, opened 2016-06-17
Implement conic by polar pairs and incident flats

gagern:
Given (*A*, *a*, *B*, *b*, *C*), compute conic such that 1. (*A*, *a*) and (*B*, *b*) are pole-polar pairs and 2. *C* is incident with the conic. This is one of four algorithms requested by #320. I found a computation which avoids all removable singularities.

It would be easy to adapt this to the dual case of two pairs and one tangent line, but then I'd have a singularity if all three lines were concurrent. I'd like to investigate a bit closer whether that singularity can be avoided.

### How this was found

I'll also document how I found this, in case the approach is applicable to other algorithms.

If one assumes the entries of the desired symmetric matrix as variables, then the characterizing conditions are polynomial equations which are linear in these variables. The conditions are essentially “*C* lies on conic” (i.e. *C*⋅*M*⋅*C*=0, one equation), “*a* is polar of *A*” (i.e. *a*×(*M*⋅*A*)=0, three equations) and “*b* is polar of *B*”. The three equations from each of the latter constraints are linearily dependent; it suffices to take two each, and after removing singularities (i.e. eliminating common factors) the choice doesn't matter. So one obtains a 5×6 system of equations, and the 5×5 minors of it can serve as a solution. Remove gcd and that's the desired matrix.

Using the polynomials obtained from the approach just described, I could experiment in Cinderella and observe that moving *B* (or *b*) led to a family of conics (drawing traces helped) which all passed through a single point on the line *AC*. Since computing the *other* intersection (other than *C*) is a linear operation, the formula for that point could be computed and simplified to *D*=⟨*a*,*A*⟩⋅*C* − 2⟨*a*,*C*⟩⋅*A*. Using (*B*, *b*) in pace of (*A*, *a*) led to *E*. So at that point I had three (fairly simple) points known to lie on the conic in question.

Next came the search for suitable simple matrices to combine. The choice that worked was the pairs (*AC*, *AE*), (*BC*, *BD*) and (*AC*, *BC*), each written as a column times row rank 1 matrix. I looked for a non-trivial linear combination where these three and the desired result would combine to zero. This led to six homogeneous equations (one for each matrix element not implied by symmetry) in four variables. Taking any three equations and using minors as above led to the required coefficients. And I was happy to notice that the coefficient of the desired matrix had degree zero, certifying the absence of removable singularities. The final formula was ⟨*A*,*b*⟩⋅*M*<sub>1</sub> + ⟨*B*,*a*⟩⋅*M*<sub>2</sub> − 2⟨*A*,*b*⟩⟨*B*,*a*⟩⋅*M*<sub>3</sub> where *M*<sub>1</sub> through *M*<sub>3</sub> are the rank-one matrices in the order given above.


src/js/libgeo/GeoOps.js
```diff
@@ -1226,6 +1226,46 @@ geoOps.ConicBy2Foci1P.updatePosition = function(el) {
...
+    var a = csgeo.csnames[(el.args[1])].homog;
+    var B = csgeo.csnames[(el.args[2])].homog;
+    var b = csgeo.csnames[(el.args[3])].homog;
+    var C = csgeo.csnames[(el.args[4])].homog;
+
+    var sp = List.scalproduct;
+    var sm = List.scalmult;
+    var sub = List.sub;
+    var rm = CSNumber.realmult;
+    var dc = geoOps._helper.degeneratePairOfJoiningLines;
```
elkins0: Applause for a very fine result indeed!
Here the dc variable goes unused an can be removed:
```javascript
    //var dc = geoOps._helper.degeneratePairOfJoiningLines; 
```

src/js/libgeo/GeoOps.js
```diff
@@ -1226,6 +1226,46 @@ geoOps.ConicBy2Foci1P.updatePosition = function(el) {
...
+    var asList = List.turnIntoCSList;
+
+    // D = ⟨a,A⟩⋅C − 2⟨a,C⟩⋅A, E = ⟨b,B⟩⋅C − 2⟨b,C⟩⋅B
+    var D = sub(sm(sp(a, A), C), sm(rm(2, sp(a, C)), A));
+    var E = sub(sm(sp(b, B), C), sm(rm(2, sp(b, C)), B));
+    var AC = asList([List.cross(A, C)]);
+    var BC = asList([List.cross(B, C)]);
+    var M1 = General.mult(transpose(AC), asList([List.cross(A, E)]));
+    var M2 = General.mult(transpose(BC), asList([List.cross(B, D)]));
+    var M3 = General.mult(transpose(AC), BC);
```
elkins0: Also General.mult always ends up being handled by List.productMM so we can write:
```javascript
    var mm = List.productMM;
    var M1 = mm(transpose(AC), asList([List.cross(A, E)]));
    var M2 = mm(transpose(BC), asList([List.cross(B, D)]));
    var M3 = mm(transpose(AC), BC);
```

gagern:
Thanks @elkins0 for suggesting these improvements. I incorporated them, and named you author of that commit.

I also found a solution for one polar pair and three incident points. It's less beautiful than the other one, and I had some trouble finding the appropriate conics for this. If anyone wants to try for a simpler solution, perhaps they could build on this Sage output where I investigated formulas of appropriate degrees, looking for linear dependencies between them:

```
goal                  1                                                            
[ABC]*[ABD]*(a,CD)       1                                                         
[ABC]*[ACD]*(a,BD)          1                                                      
[ABC]*[BCD]*(a,AD)             1                                                   
[ABD]*[ACD]*(a,BC)                1                                                
[ABD]*[BCD]*(a,AC)                   1                                             
[ACD]*[BCD]*(a,AB)                      1                                          
[ABC]*<a,A>*(BD,CD)                        1                                       
[ABC]*<a,B>*(AD,CD)                           1                                    
[ABC]*<a,C>*(AD,BD)                              1                                 
[ABC]*<a,D>*(AB,CD)                                 1                              
[ABC]*<a,D>*(AC,BD)                                    1                           
[ABC]*<a,D>*(AD,BC)                                       1                        
[ABD]*<a,A>*(BC,CD)                                          1                     
[ABD]*<a,B>*(AC,CD)                                             1                  
[ABD]*<a,C>*(AB,CD)                                                1               
[ABD]*<a,C>*(AC,BD)                                                   1            
[ABD]*<a,C>*(AD,BC)                                                      1         
[ABD]*<a,D>*(AC,BC)                             -1    -1 -1          -1 -1         
[ACD]*<a,A>*(BC,BD)   1 -1 -1    -1       -1                -1                     
[ACD]*<a,B>*(AB,CD)                                                         1      
[ACD]*<a,B>*(AC,BD)                                                            1   
[ACD]*<a,B>*(AD,BC)  -2  1  1     1          -1 -1             -1    -1 -1    -1   
[ACD]*<a,C>*(AB,BD)                                            -1 -1 -1    -1 -1   
[ACD]*<a,D>*(AB,BC)   2 -1 -1    -1              1 -1     1     1     1  1  1  1   
[BCD]*<a,A>*(AB,CD)                                                               1
[BCD]*<a,A>*(AC,BD)  -1  1  1       -1 -1  1                   -1          -1 -1 -1
[BCD]*<a,A>*(AD,BC)   1    -1 -1        1        1           1  1     1  1  1  1  1
[BCD]*<a,B>*(AC,AD)     -1     1     1        1                 1                  
[BCD]*<a,C>*(AB,AD)        -1 -1        1        1              1  1  1     1  1   
[BCD]*<a,D>*(AB,AC)  -2  1  1       -1 -1           1  1       -1          -1 -1   
```

The row labels are candidate formulas. The first column describes *one* possible solution, since it has a one in the row named “goal”. The other columns may be added to that to obtain different solutions. I boiled this down somewhat by concentrating on solutions which are symmetric under cyclic rotation of *B,C,D*, but the conic by five operation is a good example demonstrating that asymmetric solutions may be simpler. So perhaps there is a simpler solution to be found here as well.

elkins0:
Very nice simplification for the one polar pair case! 
@gagern wrote initially:
> It would be easy to adapt this to the dual case of two pairs and one tangent line, but then I'd have a singularity if all three lines were concurrent. I'd like to investigate a bit closer whether that singularity can be avoided.

Singularities also occur when any three points are collinear or any three lines are parallel. The `ConicBy2Pol1L` dual of  `ConicBy2Pol1P` works just fine shown here using the same pattern as the one polar pair case:
```javascript
// Compute the conic matrix by two polar pairs and one incident point
geoOps._helper.conic2Pol1Inc = function(A, a, B, b, C) {
    var sp = List.scalproduct;
    var sm = List.scalmult;
    var sub = List.sub;
    var mm = List.productMM;
    var rm = CSNumber.realmult;
    var transpose = List.transpose;
    var asList = List.turnIntoCSList;

    // D = ⟨a,A⟩⋅C − 2⟨a,C⟩⋅A, E = ⟨b,B⟩⋅C − 2⟨b,C⟩⋅B
    var D = sub(sm(sp(a, A), C), sm(rm(2, sp(a, C)), A));
    var E = sub(sm(sp(b, B), C), sm(rm(2, sp(b, C)), B));
    var AC = asList([List.cross(A, C)]);
    var BC = asList([List.cross(B, C)]);
    var M1 = mm(transpose(AC), asList([List.cross(A, E)]));
    var M2 = mm(transpose(BC), asList([List.cross(B, D)]));
    var M3 = mm(transpose(AC), BC);
    var Ab = sp(A, b);
    var Ba = sp(B, a);
    // M = Ba * M1 + Ab * M2 - 2 * Ab * Ba * M3
    var M = List.add(sm(Ba, M1), sm(Ab, M2));
    M = sub(M, sm(rm(2, CSNumber.mult(Ab, Ba)), M3));
    M = List.add(M, transpose(M));
    return M;
};

// Given (A, a, B, b, C), compute the conic matrix such that
// 1. (A, a) and (B, b) are pole-polar pairs and
// 2. point C is incident with the conic
geoOps.ConicBy2Pol1P = {};
geoOps.ConicBy2Pol1P.kind = "C";
geoOps.ConicBy2Pol1P.signature = ["P", "L", "P", "L", "P"];
geoOps.ConicBy2Pol1P.updatePosition = function(el) {
    var A = csgeo.csnames[(el.args[0])].homog;
    var a = csgeo.csnames[(el.args[1])].homog;
    var B = csgeo.csnames[(el.args[2])].homog;
    var b = csgeo.csnames[(el.args[3])].homog;
    var C = csgeo.csnames[(el.args[4])].homog;

    var M = geoOps._helper.conic2Pol1Inc(A, a, B, b, C);
    M = List.normalizeMax(M);
    M = General.withUsage(M, "Conic");
    el.matrix = M;
};

// Given (A, a, B, b, c), compute the conic matrix such that
// 1. (A, a) and (B, b) are pole-polar pairs and
// 2. line c is tangent to the conic
geoOps.ConicBy2Pol1L = {};
geoOps.ConicBy2Pol1L.kind = "C";
geoOps.ConicBy2Pol1L.signature = ["P", "L", "P", "L", "L"];
geoOps.ConicBy2Pol1L.updatePosition = function(el) {
    var A = csgeo.csnames[(el.args[0])].homog;
    var a = csgeo.csnames[(el.args[1])].homog;
    var B = csgeo.csnames[(el.args[2])].homog;
    var b = csgeo.csnames[(el.args[3])].homog;
    var c = csgeo.csnames[(el.args[4])].homog;

    var M = geoOps._helper.conic2Pol1Inc(a, A, b, B, c);
    M = List.adjoint3(M);
    M = List.normalizeMax(M);
    M = General.withUsage(M, "Conic");
    el.matrix = M;
};
```

src/js/libgeo/GeoOps.js
```diff
@@ -1266,19 +1266,8 @@ geoOps.ConicBy2Pol1P.updatePosition = function(el) {
...
-geoOps.ConicBy1Pol3P.kind = "C";
-geoOps.ConicBy1Pol3P.signature = ["P", "L", "P", "P", "P"];
-geoOps.ConicBy1Pol3P.updatePosition = function(el) {
-    var A = csgeo.csnames[(el.args[0])].homog;
-    var a = csgeo.csnames[(el.args[1])].homog;
-    var B = csgeo.csnames[(el.args[2])].homog;
-    var C = csgeo.csnames[(el.args[3])].homog;
-    var D = csgeo.csnames[(el.args[4])].homog;
-
+// Conic by one polar pair and three incident flats
```
elkins0: Change 'flats' to 'points'

src/js/libgeo/GeoOps.js
```diff
@@ -1309,6 +1298,42 @@ geoOps.ConicBy1Pol3P.updatePosition = function(el) {
     M = add(oneCombination(D, B, C), M);
     M = List.add(M, transpose(M));
     M = List.normalizeMax(M);
+    return M;
+};
+
+// Given (A, a, B, C, D), compute conic such that
+// 1. (A, a) is a pole-polar pair and
+// 2. B, C, D are incident with the conic
```
elkins0: Add 'points' for clarity:
```javascript
// 2. points B, C, D are incident with the conic 
```

elkins0:
@gagern It looks like you may have coined a new phrase *incident flats*. Did you mean to? The English literature typically uses *lines tangent to a curve* or *tangents of a curve* and the like. When the British refer to a flat it is what an American refers to as an apartment. And you might have a flat spot on your flat tire so in that sense it is understood what is meant. In German *flat* is *flach* and *Fläche* refers to a planar surface but both words are more relevant  in 3D than 2D. Just sayin' ... :smile:

kortenkamp:
Well, in projective geometry a flat is a subspace, so @gagern just uses a common word for point, line or (in higher dimensions) plane.

elkins0:
Interesting. I do find a *Wikipedia* article [Flat (geometry)](https://en.wikipedia.org/wiki/Flat_(geometry)) which at least corroborates that terminology although it is marked *This article needs additional citations for verification.* Also there is a one line entry in *MathWorld*:
[Flat](http://mathworld.wolfram.com/Flat.html)
A set in R^d formed by translating an affine subspace or by the intersection of a set of [hyperplanes](http://mathworld.wolfram.com/Hyperplane.html).

So flat is a good way to refer to points and lines in 2D in a dissertation.

gagern:
> ~~Singularities~~ No solutions also occur when any three points are collinear or any three lines are parallel. Note that duals of three concurrent or three parallel lines both plot as collinear points. The `ConicBy2Pol1L` dual of `ConicBy2Pol1P` works just fine shown here using the same pattern as the one polar pair case[…]

I've been using the term “singularities” to describe *removable* singularities: configurations where the given algorithm would return an all-zero matrix but that could be avoided by canceling some suitable factor from the polynomials. The adjoint of the `ConicBy2Pol1P` matrix has [*ABC*]² as a factor. So while `ConicBy2Pol1P` itself has no removable singularities, its dual matrix has a singularity where *A,B,C* are collinear (of which being parallel is just a special case, projectively speaking). That singularity is removable, and I'm pretty certain that in this degenerate case the dual matrix would factor into the pair of points *D* and *E* where the line *AB* intersects the conics no irrespective of the position of point *C*.

If we were to use that formula to compute `ConicBy2Pol1L`, as you suggest, that would mean we'd not get a solution if *a,b,c* are concurrent, even though algebraically such a solution would be (or at least could be) well defined. I don't like that. But I don't know how to easily avoid this either. Today I tried the same approach I used for the 1Pol3P problem: I listed all the combinations of suitable degree, and tried to find a linear dependency between them. Alas, I found none:

```
<a,A>*<a,B>*<b,A>*<c,B>*(b,c)
<a,A>*<a,B>*<b,B>*<c,A>*(b,c)
<a,A>*<b,A>*<b,B>*<c,B>*(a,c)
<a,A>*<b,B>*<c,A>*<c,B>*(a,b)
<a,B>*<b,A>*<b,B>*<c,A>*(a,c)
<a,B>*<b,A>*<c,A>*<c,B>*(a,b)
<a,A>*<b,B>*[abc]*(c,AB)
<a,A>*<c,B>*[abc]*(b,AB)
<a,B>*<b,A>*[abc]*(c,AB)
<a,B>*<c,A>*[abc]*(b,AB)
<b,A>*<c,B>*[abc]*(a,AB)
<b,B>*<c,A>*[abc]*(a,AB)
```

The desired solution is not contained in the span of these terms. Which has me really surprised; my gut feeling would have been that the fundamental theorem for projective invariants (which states that every projectively invariant property of a set of points which can be expressed in terms of a polynomial in the coordinates can also be expressed as a bracket polynomial, i.e. a polynomial in determinants) applies to guarantee that the desired formula has to be a combination of these formulas. But apparently the mixture of points and lines breaks this somehow. Or I made a mistake.

I guess that I'll indeed go for the adjoint for now, since writing down the polynomials at the coefficient level is no fun at all. Better to have a solution with a known singularity than not having a solution at all. But I'd still like to find a nice formula for the 2Pol1L case that avoids introducing common factors.

gagern:
I was missing some terms, particularly those which have a double line as their matrix. Adding these it becomes possible to describe the solution of 2Pol1L in a number of ways:

```
goal                           2               
<a,A>*<a,A>*<b,B>*<c,B>*(b,c)     1            
<a,A>*<a,A>*<c,B>*<c,B>*(b,b)    -1            
<a,A>*<a,B>*<b,A>*<c,B>*(b,c)  1 -1            
<a,A>*<a,B>*<b,B>*<c,A>*(b,c)        1         
<a,A>*<a,B>*<c,A>*<c,B>*(b,b)     1 -1         
<a,A>*<b,A>*<b,B>*<c,B>*(a,c)           1      
<a,A>*<b,A>*<c,B>*<c,B>*(a,b)     1    -1      
<a,A>*<b,B>*<b,B>*<c,A>*(a,c)              1   
<a,A>*<b,B>*<c,A>*<c,B>*(a,b)    -1       -1   
<a,B>*<a,B>*<b,A>*<c,A>*(b,c) -1    -1         
<a,B>*<a,B>*<c,A>*<c,A>*(b,b)  1     1         
<a,B>*<b,A>*<b,A>*<c,B>*(a,c) -1       -1      
<a,B>*<b,A>*<b,B>*<c,A>*(a,c)  1          -1   
<a,B>*<b,A>*<c,A>*<c,B>*(a,b) -2     1  1      
<a,B>*<b,B>*<c,A>*<c,A>*(a,b)       -1     1   
<b,A>*<b,A>*<c,B>*<c,B>*(a,a)  1        1      
<b,A>*<b,B>*<c,A>*<c,B>*(a,a)          -1  1   
<b,B>*<b,B>*<c,A>*<c,A>*(a,a)             -1   
<a,A>*<b,B>*[abc]*(c,AB)                      1
<a,A>*<c,B>*[abc]*(b,AB)         -1          -1
<a,B>*<b,A>*[abc]*(c,AB)      -1             -1
<a,B>*<c,A>*[abc]*(b,AB)            -1        1
<b,A>*<c,B>*[abc]*(a,AB)               -1     1
<b,B>*<c,A>*[abc]*(a,AB)                  -1 -1
[abc]*[abc]*(AB,AB)                          -2
```

I guess I'll go for the first column as computed, since adding any of the others doesn't seem to make things any easier. I've unassigned the reviewers, since I want this implemented first and merged as a whole.

gagern:
OK, I have a formula for the 2Pol1L case which avoids the singularity in the concurrent case. Please review, since I don't intend to add more operations to this pull request here.

I also gave 1Pol2P1L and 1Pol1P2L a try, basing them on the helper for ConicBy4p1l, in the 1Pol2P1L situation by using the second points of intersection on the connection between the pole and the incident points, and in the 1Pol1P2L situation by dualizing that. But this is bound to fail in the most interesting special case, namely when the pole-polar pair is a contact pair. Then the additional computed points will coincide (with the pole and therefore with one another), leading to undefined solutions.

Since I'd like to avoid case distinctions, I'd rather not switch between this and what @elkins0 wrote, but instead try to find some code which seamlessly caters for all input configurations. This is a larger piece of work, as my current machinery isn't geared for non-linear problems yet. So I'd like to get what I have so far into master, and tackle the remaining multi-valued cases separately.

elkins0:
@gagern Very nice simplification! I have used *Mathematica* to symbolically check that the *JavaScript* for `geoOps.ConicBy2Pol1L` meets the requirement that “a is polar of A wrt. M” (i.e. `a×(M⋅A)=0`), “b is polar of B wrt. M” (i.e. `b×(M⋅B)=0`) and c is tangent to M  (i.e. `c⋅adjoint(M)⋅c=0`) which of course is as expected.

The expanded `List.adjoint3` *JavaScript* found in *CindyJS* can be simplified to:
```javascript
    List.adjoint3 = function(a) {
        var r0 = a.value[0];
        var r1 = a.value[1];
        var r2 = a.value[2];
        var cp = List.cross;
        return List.transpose(List.turnIntoCSList([cp(r1, r2), cp(r2, r0), cp(r0, r1)]));
    }
```
which allows me to symbolically check `geoOps.ConicBy2Pol1L` in *Mathematica* using the following code:
```Mathematica
With[{A = {Ax, Ay, Az}, a = {ax, ay, az}, B = {Bx, By, Bz}, b = {bx, by, bz}, c = {cx, cy, cz}, 
  adjoint3 = Transpose[{Cross[#[[2]], #[[3]]], Cross[#[[3]], #[[1]]], Cross[#[[1]], #[[2]]]}]&}, 
 Block[{aA = a.A, aB = a.B, bA = b.A, bB = b.B, cA = c.A, cB = c.B, v, M}, 
  v = {(bA*cB)*a - (aB*cA)*b};
  M = Transpose[{(bA*aB)*c}].{(aA*cB - aB*cA)*b + (bB*cA - bA*cB)*a -
                 Det[{a, b, c}]*Cross[A, B]} + Transpose[v].v; 
  M = Transpose[M] + M;
  Simplify[{ 
   "a is polar of A wrt. M" -> Cross[a, M.A] == {0, 0, 0}, 
   "b is polar of B wrt. M" -> Cross[b, M.B] == {0, 0, 0}, 
   "c is tangent to M" -> c.adjoint3[M].c == 0}]]]
```
And *Mathematica* gives this confirmation:
```Mathematica
{"a is polar of A wrt. M" -> True, "b is polar of B wrt. M" -> True, "c is tangent to M" -> True}
```

gagern:
Thanks @elkins0 for verifying this. I was quite confident in the sage computation, and the example looked good as well, but it never hurts to have things formally verified in an independent way.

I'm a bit surprised that Mathematica doesn't ship a built-in function to compute the classical adjoint of a matrix, even though there is a short definition in the *Applications* section of the [`Minors` documentation](http://reference.wolfram.com/mathematica/ref/Minors.html) as [pointed out on Mathematica Stack Exchange](http://mathematica.stackexchange.com/a/37927/8395). I'm glad I'm on Sage by default…

I guess that I'd add one final test, namely one to verify that the degree of the greatest common (polynomial) divisor is equal to zero. That has caused me the most grief, and it's hardest to verify just from looking at the examples.

elkins0:
@gagern wrote:
>I guess that I'd add one final test, namely one to verify that the degree of the greatest common (polynomial) divisor is equal to zero.

Right, so I just added `PolynomialGCD @@ Flatten[M]` to the output list to show the actual value. `NumericQ` could used to do that check on the value, but I wanted to see what it was.
```Mathematica
With[{A = {Ax, Ay, Az}, a = {ax, ay, az}, B = {Bx, By, Bz}, b = {bx, by, bz}, c = {cx, cy, cz}, 
  adjoint3 = Transpose[{Cross[#[[2]], #[[3]]], Cross[#[[3]], #[[1]]], Cross[#[[1]], #[[2]]]}]&}, 
 Block[{aA = a.A, aB = a.B, bA = b.A, bB = b.B, cA = c.A, cB = c.B, v, M}, 
  v = {(bA*cB)*a - (aB*cA)*b};
  M = Transpose[{(bA*aB)*c}].{(aA*cB - aB*cA)*b + (bB*cA - bA*cB)*a -
                 Det[{a, b, c}]*Cross[A, B]} + Transpose[v].v; 
  M = Transpose[M] + M;
  Simplify[{ 
   "a is polar of A wrt. M" -> Cross[a, M.A] == {0, 0, 0}, 
   "b is polar of B wrt. M" -> Cross[b, M.B] == {0, 0, 0}, 
   "c is tangent to M" -> c.adjoint3[M].c == 0,
   "GCD of matrix" -> PolynomialGCD @@ Flatten[M]}]]]
```
which gives:
```Mathematica
{"a is polar of A wrt. M" -> True, "b is polar of B wrt. M" -> True, 
 "c is tangent to M" -> True, "GCD of matrix" -> 2}
```

elkins0:
A similar symbolic *Mathematica* check applied `geoOps.ConicBy2Pol1P` looks like this:
```Mathematica
With[{A = {Ax, Ay, Az}, a = {ax, ay, az}, B = {Bx, By, Bz}, b = {bx, by, bz}, C = {Cx, Cy, Cz}},
 Block[{D, E, AC, BC, M1, M2, M3, Ab, Ba, M},
  D = (a.A) C - 2 (a.C) A;
  E = (b.B)*C - 2 (b.C)*B;
  AC = {Cross[A, C]}; BC = {Cross[B, C]};
  M1 = Transpose[AC].{Cross[A, E]};
  M2 = Transpose[BC].{Cross[B, D]};
  M3 = Transpose[AC].BC; Ab = A.b; Ba = B.a;
  M = Ba*M1 + Ab*M2 - 2*Ab*Ba*M3;
  M = Transpose[M] + M;
  Simplify[{
   "a is polar of A wrt. M" -> Cross[a, M.A] == {0, 0, 0},
   "b is polar of B wrt. M" -> Cross[b, M.B] == {0, 0, 0},
   "C lies on conic" -> C.M.C == 0,
   "GCD of matrix" -> PolynomialGCD @@ Flatten[M]}]]]
```
which gives:
```Mathematica
{"a is polar of A wrt. M" -> True, "b is polar of B wrt. M" -> True,
 "C lies on conic" -> True, "GCD of matrix" -> 2}
```

elkins0:
The symbolic *Mathematica* check applied to `geoOps.ConicBy1Pol3P` looks like this:
```Mathematica
With[{A = {Ax, Ay, Az}, a = {ax, ay, az}, B = {Bx, By, Bz}, C = {Cx, Cy, Cz}, D = {Dx, Dy, Dz}},
 Block[{ABC = Det[{A, B, C}], BD = {Cross[B, D]}, AD = {Cross[A, D]}, BC = {Cross[B, C]},
  aA = a.A, aB = a.B, aD = a.D, v, M, f}, 
  v = {Cross[C, aA*D - 2*aD*A]};
  M = ABC*Transpose[BD].v; 
  f = Det[{B, C, D}]*aA - 2*(Det[{A, C, D}]*aB + ABC*aD); 
  M += f*Transpose[AD].BC; M += Transpose[M]; 
  Simplify[{
    "a is polar of A wrt. M" -> Cross[a, M.A] == {0, 0, 0}, 
    "C lies on conic" -> C.M.C == 0,
    "D lies on conic" -> D.M.D == 0, 
    "GCD of matrix" -> PolynomialGCD @@ Flatten[M]}]]]
```
which gives:
```Mathematica
{"a is polar of A wrt. M" -> True, "C lies on conic" -> True, 
 "D lies on conic" -> True, "GCD of matrix" -> 1}
```
However checking `geoOps.ConicBy1Pol3L` with the same code but adding `adjoint3` and using
`PolynomialGCD @@ Flatten[adjoint3[M]]` instead is taking to long...

So instead I opted to use `Table[Head[Factor[adjoint3[M]][[i, j]]], {i, 3}, {j, 3}]` which shows that there are no factors, only sums of symbols (one would see `Times` if there were factors):
```Mathematica
{{Plus, Plus, Plus}, {Plus, Plus, Plus}, {Plus, Plus, Plus}}
```

strobelm:
Very nice, code looks fine. Merged.

PR #384, opened 2016-06-21
Fix mouseupbug #249 on iOS

### LINKED ISSUE #249: mouseup does not trigger draw on iOS
If you change variables your drawscript depends upon in the mouseup script, these changes will not be visible on iOS (tested on iOS9.2.1 and iOS8.4.1) until e.g. the next mousedown.

Note that this bug does not occur on OSX or when using a Cindy.js from last July.

**Example**
`doit = true;` as mouseup script and `if(doit, draw([0,0]));` as draw script

sthoch:
This fixes #249. All it required was a reording of function calls in touchUp, such that the function body equaled that of mouseUp.

I hope I got this PR-thingy correctly.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -265,9 +265,9 @@ function setuplisteners(canvas, data) {
         mouse.down = false;
         cindy_cancelmove();
         stateContinueFromHere();
-        updateCindy();
         cs_mouseup();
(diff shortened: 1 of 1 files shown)


gagern:
Thanks for fixing this.

PR #385, opened 2016-06-22
Switch from unzip to yauzl

gagern:
The node package “[unzip](https://github.com/EvanOxfeld/node-unzip)” appears to be pretty much [unmaintained](https://github.com/EvanOxfeld/node-unzip/issues/95), and it has [problem with the more recent versions of node](https://github.com/EvanOxfeld/node-unzip/issues/97).  The package “[yauzl](https://github.com/thejoshwolfe/yauzl)” (Yet Another UnZip Library) appears to be the most suitable [replacement](https://github.com/EvanOxfeld/node-unzip/issues/98).

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -417,7 +417,7 @@ module.exports = function build(settings, task) {
         "validation-api-1.0.0.GA-sources",
     ];
     var gwt_jars = gwt_parts.map(function(name) {
-        return "download/gwt-" + settings.get("gwt_version") + "/" +
+        return "gwt-" + settings.get("gwt_version") + "/" +
             name + ".jar";
     });
     var gwt_modules = glob.sync("src/java/cindyjs/*.gwt.xml")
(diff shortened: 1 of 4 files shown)


kranich:
@gagern Thanks for fixing this.

kranich:
@gagern I hope you considered this ready for review and merge. Maybe we should not assign pull requests before the pull request is ready for review. (The push after the assignment was a bit unexpected, but was included in my review.)

gagern:
@kranich: Yes, this was ready. When I assign a reviewer, I mean that *at that moment* I consider the branch complete. Nevertheless, occasionally I think of some related stuff that would fit in nicely with the merge in question, and update the branch. If it doesn't arrive in time, that's no problem since I'd just open another request for that. If it got included in a merge without recieving due review, that would be a bit of bad luck, though, so perhaps I should make an effort to do this less often.

PR #386, opened 2016-06-22
Timer

gagern:
Instead of calling `updateCindy` directly when something changed, we now make sure that there is a requested animation frame which will run `updateCindy`. This ensures that we have consistent call stacks when drawing, reducing the risk of unexpected interactions between unrelated code.

This removes the D3 timer.  As far as I can tell from reading the code, all it did *in our setup* was [delegate](https://github.com/CindyJS/CindyJS/blob/cb4026161e9e6f8a71f9703b1e3771bf2335be3d/src/js/Timer.js#L86) to [`requestAnimationFrame`](https://github.com/CindyJS/CindyJS/blob/cb4026161e9e6f8a71f9703b1e3771bf2335be3d/src/js/Timer.js#L45-L47), so we might as well do that ourselves and avoid all that extra machinery we don't use.

This is a breaking change: the CindyScript [`repaint` command](https://github.com/CindyJS/CindyJS/blob/12929c46942649fd7f6f9c298a664f1364007553/src/js/libcs/OpDrawing.js#L1386-L1389) no longer causes an *immediate* [execution of the `draw` script](https://github.com/CindyJS/CindyJS/blob/12929c46942649fd7f6f9c298a664f1364007553/src/js/libgeo/Tracing.js#L110-L111).  So code which executes `repaint` and then relies on observing results generated in the `draw` script would break.  This would be a pretty hackish approach, so most sane widgets should be safe.


gagern:
I'd like some performance evaluation and general OK from @richter-gebert first, and some code review from @kranich or @strobelm afterwards.

gagern:
We recently decided to give this a try, and revert to the D3 timer if any serious problems crop up.

strobelm:
Looks fine, thanks @gagern!

Fixes issue  #419.

PR #389, opened 2016-06-24
ComplexCurves plugin

kranich:
Complex Curves is a [JavaScript library](https://github.com/kranich/ComplexCurves) and [website](http://complexcurves.org) for visualization of complex plane algebraic curves.

It is based on Chapter 6 of my dissertation,

Stefan Kranich, Continuity in Dynamic Geometry: An Algorithmic
Approach, PhD thesis, Technical University of Munich, 2016,
[http://nbn-resolving.de/urn/resolver.pl?urn:nbn:​de:bvb:91-diss-20160418-1281075-1-5](http://nbn-resolving.de/urn/resolver.pl?urn:nbn:de:bvb:91-diss-20160418-1281075-1-5).

The ComplexCurves plugin makes the core functionality of the [website](http://complexcurves.org) in its current state accessible via a CindyJS plugin.

There are still many rough edges, known bugs, and performance is suboptimal. I will try to find time to fix these issues. However, I think that the plugin in its current form is already better than nothing.

Bug reports and pull requests in the [main repository](https://github.com/kranich/ComplexCurves) are very welcome.


gagern:
I gave the submodules a try: https://github.com/CindyJS/CindyJS/compare/master...gagern:complex-curves-plugin. But I'm not happy with this approach yet: I can't make the submodule shallow, so even though I'm referencing the nocache branch, I still have to fetch all that expensive history. [This SO post](http://stackoverflow.com/a/17692710/1468366) suggests that https://github.com/git/git/commit/fb43e31f2b43076e7a30c9cd00d0241cb8cf97eb (i.e. git 2.8 or later) should allow shallow submodules, but perhaps the GitHub server isn't up to this mode of operation. I'll think and experiment some more, but if that fails switch to zip download or similar.

gagern:
@strobelm: I think I'd like your opinion on this. Please compare this PR here, which adds the relevant files to our repository, with https://github.com/CindyJS/CindyJS/compare/master...gagern:complex-curves-plugin which downloads the ZIP instead. The main benefit of the ZIP approach is that it's easy to update to a new version by simply changing the commit id. This one here requires a bit more work, which I'm sure Stefan is willing to provide for the forseeable future, but I'd prefer not to depend on that.

On the other hand, that ZIP file machinery, and the up-to-date checks for it in particular, add quite a bit of overhead and are not always elegant or intuitive. Mainly because the corresponding directory is a mixture between downloaded resource (since we download the ZIP), generated resource (since unpacking auto-generates the files from the ZIP) and source file (since we use these as input for other stages). The way the plugin refers to the library code using `import {…} from "../../lib/ComplexCurves/src/js/ComplexCurves"` made encoding the commit id into the directory name problematic, so I left the files at `lib`.

Does it make sense to add all that complexity to the project, or should we just import the files as this PR here requests? Which one will be easier to maintain in the future?

kranich:
> This one here requires a bit more work, which I'm sure Stefan is willing to provide for the forseeable future, but I'd prefer not to depend on that.

I can promise to maintain the plugin as long as I maintain the Complex Curves project itself.

Updating the plugin is relatively hassle-free. I added the plugin sources to `plugins/ComplexCurves/lib/ComplexCurves` as follows:

```
git remote add ComplexCurves git@github.com:kranich/ComplexCurves.git
git fetch ComplexCurves
git read-tree --prefix=plugins/ComplexCurves/lib/ComplexCurves -u ComplexCurves/nocache
```

And then an update boils down to:

```
git fetch ComplexCurves
git merge -X subtree=plugins/ComplexCurves/lib/ComplexCurves --squash ComplexCurves/nocache
```

I learned this strategy from https://medium.com/@porteneuve/mastering-git-subtrees-943d29a798ec#.z4iaxkiu0

kranich:
As for the ZIP approach, it might make sense to always download the latest `nocache` branch that I will rebase onto the changes made to `master` when they are ready.

Coming to think of it, maybe it makes more sense to remove the large binary files from the ComplexCurves repository altogether and just put them elsewhere (GitHub large file storage, maybe?).

Currently, the only reason for the `nocache` branch is to remove these binary files for the plugin, in order to keep the plugin lean. Having to rebase it onto master seems awkward however.

@gagern What do you suggest?

gagern:
Having the large files out of the way would indeed make things *a lot* easier. (It would even make the submodule approach worth considering again.) You'd have to retroactively clean the history. I guess you could place the files in question on the GitHub Pages for the project. Perhaps you can even tweak the project in such a way that it loads the resources from that location no matter where the libraries themselves are located.

As for referencing the tip of the `nocache` branch: I think that would be too restrictive. It would mean that you can never break API compatibility, because doing so would break the interoperation with the plugin in the CindyJS repository. I also prefer perfect repsorucibility of builds, which means all the essential build tools (sources, libs, compiler) should be fixed to a well-kown version in the repo. We should make updating easy, though.

kranich:
>Having the large files out of the way would indeed make things a lot easier. (It would even make the submodule approach worth considering again.) You'd have to retroactively clean the history.

I'll see what I can do.

kranich:
I have interactively rebased onto the root of [kranich/ComplexCurves master](https://github.com/kranich/ComplexCurves/master), removing all large binary files from the history and referring to http://complexcurves.org/models/ where appropriate.

The models are now part of [kranich/ComplexCurves gh-pages](https://github.com/kranich/ComplexCurves/gh-pages).

The ComplexCurves plugin is now based on [kranich/ComplexCurves master](https://github.com/kranich/ComplexCurves/master) and I deleted the `nocache` branch.

`git clone git@github.com:kranich/ComplexCurves.git --branch master --depth 1 ComplexCurves` now only receives 911.37 KiB, what an improvement!

@gagern How shall we proceed from here?

gagern:
> How shall we proceed from here?

I'd still like a second opinion on whether we should include the sources or use the ZIP, with using a submodule as yet another viable approach now. I guess we'll discuss this at our next video conference.

kranich:
Superseded by #521.

PR #393, opened 2016-06-28
conic4p1l: add tracing and remove random factor

### LINKED ISSUE #101: ConicBy4p1l doesn't trace
`geoOps._helper.ConicBy3p2l` doesn't do any tracing, so its outputs may get swapped arbitrarily.

gagern:
This improves the ConicBy4p1l and ConicBy1p4l operations by removing a random factor and adding tracing code. I'm a bit surprised how stable the operation is in the absence of tracing, but nevertheless I found (after much searching) a configuration which reliably exhibited flipping with the old code, and thus allowed me to verify tracing for the new code. I added a button to the example to make that configuration available.

This fixes #101 (assuming that one really was about 4P1L).

I would have liked to check my algorithm for avoidable singularities, but when I tried to eliminate the square root using a resultant, I ran out of memory. So I still don't know whether this implementation has any common factors.

*Note:* For some reason the conics coming out of this example appear to be particularly problematic for the current conic drawing algorithm. I see a lot of rather coarse polygonal approximations. Might be worth remembering when evaluating #280.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -961,11 +961,12 @@ geoOps._helper.ConicBy4p1l = function(el, a, b, c, d, l) {
     var a2 = List.cross(List.cross(b, d), l);
     var k1 = List.scalmult(r1, a1);
     var k2 = List.scalmult(r2, a2);
-    var x = List.add(k1, k2);
-    var y = List.sub(k1, k2);
-    var t1 = geoOps._helper.ConicBy5(el, a, b, c, d, x);
-    var t2 = geoOps._helper.ConicBy5(el, a, b, c, d, y);
-    return [t1, t2];
+    var x = List.normalizeMax(List.add(k1, k2));
+    var y = List.normalizeMax(List.sub(k1, k2));
+    var xy = tracing2(x, y);
+    var t1 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[0]);
+    var t2 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[1]);
+    return [List.normalizeMax(t1), List.normalizeMax(t2)];
 };
 
 geoOps.ConicBy4p1l = {};
@@ -984,6 +985,7 @@ geoOps.ConicBy4p1l.updatePosition = function(el) {
     el.results = erg;
 
 };
+geoOps.ConicBy4p1l.stateSize = tracing2.stateSize;
 
 
 geoOps._helper.ConicBy3p2l = function(a, b, c, g, h) {
@@ -1126,6 +1128,7 @@ geoOps.ConicBy1p4l.updatePosition = function(el) {
     el.results = erg;
 
 };
+geoOps.ConicBy1p4l.stateSize = tracing2.stateSize;
 
 geoOps.ConicParabolaPL = {};
 geoOps.ConicParabolaPL.kind = "C";
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -946,26 +946,25 @@ geoOps.SelectConic.updatePosition = function(el) {
 
 // conic by 4 Points and 1 line
 geoOps._helper.ConicBy4p1l = function(el, a, b, c, d, l) {
+    var al = List.scalproduct(a, l);
+    var bl = List.scalproduct(b, l);
+    var cl = List.scalproduct(c, l);
+    var dl = List.scalproduct(d, l);
+    var bcd = List.det3(b, c, d);
+    var abd = List.det3(a, b, d);
+    var acd = List.det3(a, c, d);
+    var abc = List.det3(a, b, c);
+    var mul = CSNumber.mult;
+    var r1 = CSNumber.sqrt(mul(mul(bl, dl), mul(bcd, abd)));
+    var r2 = CSNumber.sqrt(mul(mul(al, cl), mul(acd, abc)));
     var a1 = List.cross(List.cross(a, c), l);
     var a2 = List.cross(List.cross(b, d), l);
-    var b1 = List.cross(List.cross(a, b), l);
-    var b2 = List.cross(List.cross(c, d), l);
-    var o = List.realVector(csport.to(100 * Math.random(), 100 * Math.random()));
-
-    var r1 = CSNumber.mult(List.det3(o, a2, b1), List.det3(o, a2, b2));
-    r1 = CSNumber.sqrt(r1);
-    var r2 = CSNumber.mult(List.det3(o, a1, b1), List.det3(o, a1, b2));
-    r2 = CSNumber.sqrt(r2);
-
     var k1 = List.scalmult(r1, a1);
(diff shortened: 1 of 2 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -946,27 +946,27 @@ geoOps.SelectConic.updatePosition = function(el) {
...
-    var t1 = geoOps._helper.ConicBy5(el, a, b, c, d, x);
-    var t2 = geoOps._helper.ConicBy5(el, a, b, c, d, y);
-
-    return [t1, t2];
+    var x = List.normalizeMax(List.add(k1, k2));
+    var y = List.normalizeMax(List.sub(k1, k2));
+    var xy = tracing2(x, y);
+    var t1 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[0]);
+    var t2 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[1]);
+    return [List.normalizeMax(t1), List.normalizeMax(t2)];
```
elkins0: `List.scalproduct`, `List.cross`, `List.det3` and `List.normalizeMax` are used four times each. References to these `List` functions could all be localized like so:
```javascript
	var sp = List.scalproduct;
	var cp = List.cross;
	var det3 = List.det3;
	var nm = List.normalizeMax;
    var al = sp(a, l);
    var bl = sp(b, l);
    var cl = sp(c, l);
    var dl = sp(d, l);
    var bcd = det3(b, c, d);
    var abd = det3(a, b, d);
    var acd = det3(a, c, d);
    var abc = det3(a, b, c);
    var mul = CSNumber.mult;
    var r1 = CSNumber.sqrt(mul(mul(bl, dl), mul(bcd, abd)));
    var r2 = CSNumber.sqrt(mul(mul(al, cl), mul(acd, abc)));
    var a1 = cp(cp(a, c), l);
    var a2 = cp(cp(b, d), l);
    var k1 = List.scalmult(r1, a1);
    var k2 = List.scalmult(r2, a2);
    var x = nm(List.add(k1, k2));
    var y = nm(List.sub(k1, k2));
    var xy = tracing2(x, y);
    var t1 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[0]);
    var t2 = geoOps._helper.ConicBy5(el, a, b, c, d, xy.value[1]);
    return [nm(t1), nm(t2)];
```

elkins0:
@gagern I'm curious as to why tracing was added to the computed points `x` and `y` in `geoOps._helper.ConicBy4p1l` prior to passing them to `geoOps._helper.ConicBy5`. Tracing will be done again later in `geoOps.SelectConic`. This seems needless here.

elkins0:
Using the following setting for the input geometry, one of the solution points `x` and `y` in `geoOps._helper.ConicBy4p1l` experiences chaotic behavior as the line position is jiggled:
```javascript
			  {name:"A", type:"Free", pos:[2,5,1]},
			  {name:"B", type:"Free", pos:[2,3,1]},
			  {name:"C", type:"Free", pos:[4,4,1]},
			  {name:"D", type:"Free", pos:[0.44,3,1]},
			  {name:"X", type:"Free", pos:[6,5,1]},
			  {name:"Y", type:"Free", pos:[6,0,1]},
``` 
In *Mathematica* I see that one of the solution points is `[0,0,0]` with this positioning by using the following rationalized input:
```Mathematica
Block[{A = {2, 5, 1}, B = {2, 3, 1}, C = {4, 4, 1}, D = {11/25, 3, 1},
   l =Cross[{6, 5, 1}, {6, 0, 1}], r1, r2, k1, k2}, 
 r1 = Sqrt[(B.l)*(D.l)*Det[{B, C, D}] Det[{A, B, D}]]; 
 r2 = Sqrt[(A.l)*(C.l)*Det[{A, C, D}] Det[{A, B, C}]]; 
 k1 = r1*Cross[Cross[A, C], l]; k2 = r2*Cross[Cross[B, D], l]; 
 Print[k1 == -k2];
 If[Last@# == 0, #, #/Last@#] & /@ {k1 + k2, k1 - k2}]
```
The resulting output is:
```Mathematica
True
{{0, 0, 0}, {6, 3, 1}}
```
One of the solutions is `[0,0,0]` because `k1 == -k2`. The expected solutions are `[6,6,1]` and `[6,3,1]`. Using `D={0.44,3,1}` instead shows a chaotic result due to floating point noise:
```Mathematica
{{4.`+0.`I, 0.`+0.`I, 1.`+0.`I}, {6.000000000000001`+0.`I,  3.0000000000000004`+0.`I, 1.`+0.`I}}
```

gagern:
@elkins0 wrote:

> Tracing will be done again later in `geoOps.SelectConic`.

No, it won't. `SelectConic` will use the location given in the element description to initially choose the appropriate index. After that, it will just select the conic by index, it contains no tracing state of its own. It was decided to put tracing functionality into the multi-valued geometric operations, not into the selction operations. One reason is that it's often easier to trace there, like in this case here where one has to trace a pair of points (three complex coordinates each) instead of a pair of conics (six coordinates each). Another reason is that this way it is easier to keep things concistent: if two select operations select two distinct outputs, they will always refer to these distinct outputs, and can't accidentially end up referring to the same output just because the mover changed in some degenerate situations where the outputs could not be distinguished.

Thanks for this problematic test case. I see that we still have singularities to avoid in this operation. And I'm not surprised. I've improved my machinery for simplifying conic operations, and put it to use in #399, so I'll see whether it is up to this task now. Last time I tried, I couldn't get the required computations done in reasonable time. Anyway, this pull request here does already fix an existing bug, so I'd like to see it merged even though we know that there are still things worth improving.

elkins0:
It turns out that if the positions of points `B` and `C` are exchanged, both valid solutions are produced:
```Mathematica
Block[{A = {2, 5, 1}, C = {2, 3, 1}, B = {4, 4, 1}, D = {11/25, 3, 1},
   l =Cross[{6, 5, 1}, {6, 0, 1}], r1, r2, k1, k2}, 
 r1 = Sqrt[(B.l)*(D.l)*Det[{B, C, D}] Det[{A, B, D}]]; 
 r2 = Sqrt[(A.l)*(C.l)*Det[{A, C, D}] Det[{A, B, C}]]; 
 k1 = r1*Cross[Cross[A, C], l]; k2 = r2*Cross[Cross[B, D], l]; 
 Print[k1 == -k2];
 If[Last@# == 0, #, #/Last@#] & /@ {k1 + k2, k1 - k2}]
```
The resulting output is:
```Mathematica
False
{{6, 545/89, 1}, {6, 3, 1}}
```
Trying all possible permutations never produces any results other than these four:
```Mathematica
{{0, 0, 0}, {6, 3, 1}}
{{6, 3, 1}, {0, 0, 0}}
{{6, 3, 1}, {6, 545/89, 1}}
{{6, 545/89, 1}, {6, 3, 1}}
```

strobelm:
Thanks @gagern! If you are interested i could run the sage (or mathematica?) code here on machine (with 16gb ram). Merged anyway.

PR #395, opened 2016-06-29
Bind development webserver to localhost only

gagern:
This was made possible by https://github.com/isaacs/st/pull/74 and is good for security since it reduces the risk of accidentially exposing sensitive information.


strobelm:
Good idea! Thanks @gagern.

PR #396, opened 2016-06-29
Position tool buttons in front of the widget

gagern:
This places buttons in front of the widget in a way compatible with Cinderella. A modifier can be used to direct buttons into HTML elements outside the main CindyJS widget, as demonstrated in one example.


strobelm:
Looks fine to me. Merged.

PR #397, opened 2016-06-29
Implement live build mode

gagern:
In this mode, sources are watched for changes, triggering a build.
If such a build is successful, then browser windows reloaded.

I've been wanting to do this for quite some time now, and finally got round to it. Or in other words, been sufficiently annoyed by the old state of affairs. Ideally, you'd just need two windows: editor and browser. If everything works as expected, then you can ignore the console window after executing `node make live` since it will take care of everything in the background.

I'm using the [BrowserSync](https://www.browsersync.io/) to serve files and handle the reload machinery, and [chokidar](https://github.com/paulmillr/chokidar#readme) to watch for changes. Both of these are very powerful and used in a number of projects. We might find more lightweight solutions that satisfy our needs, but I doubt they'd see the same kind of maturity as these two heavily used packages.

I do notice that BrowserSync apparently does not provide an option to just bind to localhost. So I guess I'll make a similar request for that project as I did for st.

I'd like to see the live mode in action for a few weeks before documenting it in the README.

I also haven't included the `live` target in the `Makefile`. Perhaps I should? I'd welcome opinions on this. One problem is that the combination of `live` with other targets won't work with make since `live` is in fact more of a switch than an actual target, and make will delegate targets to the build system one at a time anyway. So `make live …` and `node make live …` would behave fundamentally different for non-empty `…` which might be somewhat confusing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,7 @@ fwdtargets = \
 	forbidden \
 	jshint \
 	katex \
+	live \
 	nodetest \
 	proxy \
 	ref \
--- a/make/README.md
+++ b/make/README.md
@@ -52,9 +52,14 @@ Contributors are encouraged to run these before creating a pull request.
 Travis CI will run the same tests on every pull requests,
 so addressing any issues before will save some time for everybody involved.
 
-There is also ca task called **clean** which removes the `build` directory.
+There is also a task called **clean** which removes the `build` directory.
 It is somewhat special because it will get executed first,
 no matter the order of tasks on the command line.
+
+Another special task is called **live**.
+It will open a browser at startup,
+rebuild the named targets on every change to one of the source files,
+and reload all connected browser windows in case of a successful build.
 
 ## Internals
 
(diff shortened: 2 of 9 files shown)


strobelm:
Hi Martin,

this looks very promising! 

I think there is a bug: if i change some source file the live mode captures that very well and rebuilds but it looks like that this only happens once, so if i change the file again rebuilding is not triggered.

Furthermore my ancient node version on fedora 23 won't work, i'll probably update soon but i think it's a good idea to add live mode to our makefile in order to use the downloaded node version.

gagern:
Thanks for testing, @strobelm. I just added a forwarding rule to the Makefile. Can you please try `make NODE_VERSION=6.2.0`? If that works, we can add another rule to change the default version of Node we use to 6.2.0. If we can figure out what versions of node work and what do not, we could even add a warning to the live build mode if the current version is unsuitable.

strobelm:
Unfortunately doesn't change the behavior.

-- 
Michael Strobel, M.Sc.

strobelm:
I tested chokidar-cli which works fine on my machine.

versions used:
chokidar-cli: 1.2.0
chokidar: 1.6.0


Am 14.07.2016 um 14:14 schrieb Martin von Gagern:
> Thanks for testing, @strobelm <https://github.com/strobelm>. I just
> added a forwarding rule to the Makefile. Can you please try |make
> NODE_VERSION=6.2.0|? If that works, we can add another rule to change
> the default version of Node we use to 6.2.0. If we can figure out what
> versions of node work and what do not, we could even add a warning to
> the live build mode if the current version is unsuitable.
> 
> —
> <https://github.com/CindyJS/CindyJS/pull/397#issuecomment-232648656>, or
> mute the thread
> <https://github.com/notifications/unsubscribe/ANhz4oL-JnwR7lL8JeBlLdXrH_WVeudMks5qVigPgaJpZM4JBWy7>.
> 

-- 
Michael Strobel, M.Sc.

gagern:
I tested in a Docker, and it works there, too. We should have a closer look at this together, to see what's special about your setup…

PR #398, opened 2016-06-30
Show animation controls if requested

gagern:
Show animation controls in an overlay in front of the canvas if the `animcontrols` setting is set to true.

This commit cleans up CindyJS.scss somewhat to avoid excessive nesting and to allow re-using CindyJS-specific classes outside the CindyJS widget area.

This is also the start for using SVG button icons.  The icons are served from a single SVG file in order to allow sharing styles and other things between icons, and to keep the number of transferred files low.  The approach of using CSS and the `:target` selector to control layer visibility was inspired by https://css-tricks.com/svg-fragment-identifiers-work/.

This commit conflicts with #396, but the intention is that the SCSS selector `%toggleButtonGrid` should be used by both animation controls and toolbars. I've just merged both branches to resolve the conflicts, and the result is in https://github.com/gagern/CindyJS/commit/71e1053a4b422d5aa8daf1fb0954b3f601067580. So when one of these two issues gets merged, I'd fast-forward-merge that resolution into the other unless you beat me to it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,6 +20,7 @@
 *.py            text
 *.scss          text
 *.sh            text
+*.svg           text
 *.taskpaper     text
 *.ttf           -text
 *.txt           text
--- a/examples/animcontrols.html
+++ b/examples/animcontrols.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<link rel="stylesheet" href="../build/js/CindyJS.css">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script type="text/javascript">
+
+var cdy = CindyJS({
+  ports: [{id: "CSCanvas", width: 500, height: 500}],
+  scripts: "cs*",
+  language: "en",
+  animcontrols: true,
+  defaultAppearance: {
+  },
+  geometry: [
+    {name:"A", type:"Free", pos:[-5,2], color:[1,.5,.5]},
+    {name:"B", type:"Free", pos:[2,1], color:[1,.8,0], size:9}
+  ],
+  behavior: [
+    {type:"Environment", deltat:.2, accuracy:10},
+    {type:"Mass", geo:["A"], vy:.7},
+    {type:"Sun", geo:["B"]}
+  ]
+});
+
+</script>
+</head>
+
+<body style="font-family:Arial;">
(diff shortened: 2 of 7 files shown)


gagern:
Current Safari has problems with icons in an SVG icon stack, not only on iOS but on OS X as well. Which is probably a showstopper here 😢. I guess for deployment we should either have a bunch of separate (but quite similar) SVG files, or use a big sprite with distinct positions for the different icons as a CSS background. I'll investigate the latter option, see if I can get that automated in the build process somehow.

gagern:
I think I found a beautiful way to keep the SVG stack approach while making sure that things work on Safari as well. So this should be ready for review now.

strobelm:
Hi Martin,

it looks like the play button is not working on Firefox (47) there is no log message on the console, could you verify that?

Furthermore it's interesting (and certainly no show stopper) how different browser treat the svg images:

control-ff
control-chrome

While Firefox (top) looks a bit over smoothed Chrome (bottom) could use some more AA.

gagern:
Yes, I can verify the Firefox problem. Strange, wouldn't have expected that. Thanks for catching it, investigating now…

gagern:
Found the problem: I was adding the click event handler to the image, not the button. Fixed while rebasing.

The buttons currently use filter effects to achieve that 3D look, and filter effect rendering appears to be highly implementation-specific. Perhaps we should simply use a bunch of polygonal shapes with suitable gradients instead of these effects, in order to achieve a more consistent look across browsers. Not a showstopper, though.

One thing I noticed: in Chrome on Android at least, filter effects would not scale when zooming. So the embossed border of the symbols would stay at approximately one or two device pixels no matter the zoom setting, making the overall appearance vary wildly. That, too, would be fixed by doing the 3D lighting manually instead of using filter effects.

kortenkamp:
Actually all this fuss seems to be a reason to use pixel graphics instead of SVG…

gagern:
@kortenkamp Then we'd have the fuss of generating the icons at various pixel densities, plus a different kind of fuss to combine them into sprites and use CSS backgrounds instead of `` elements. Or the cost of ransferring a larger number of icons for a given widget offering several tools.

And I'm very strongly in favor of using SVG for the design, so that there *is* a vector version available for reference (and for printed documents). It also makes it possible to change design decisions (like e.g. the color for curves) without manually editing all the files, thanks to CSS stylesheets. I have done some work along these lines already. Right now I don't know a suitably robust cross-platform way to rasterize SVG as part of our build process. So I'd stick with SVG for the moment unless someone objects strongly.

PR #399, opened 2016-07-03
Multiple conics by polar pair and mixed flats

gagern:
This is an extension of #382, implementing the multi-valued operations suggested by @elkins0 in https://github.com/CindyJS/CindyJS/issues/320#issuecomment-226967432. I'm not happy yet with the way the matrices of the 1Pol1P2L operation in particular are expressed. I included the linear dependencies in the corresponding commit message, and invite everyone to try find a more compact or otherwise more elegant formulation of that matrix.

I'd also ask @strobelm to please double-check that indeed we may use `tracing2core` and the `List.projectiveDistMinScal` it builds upon in a situation where the vectors have six elements each. I could see no reason against doing so, but I'm also pretty sure we didn't envision that use case initially.

Note that this pull requests includes the commits from #382, so if this gets merged, that one is superseded. Conversely, if that one gets merged, this one here becomes a lot thinner and easier to review, since then it will concentrate on the two new operations only.

### DIFF
--- a/examples/ConicBy1Pol1P2L.html
+++ b/examples/ConicBy1Pol1P2L.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Conic by one polar pair, one point and two lines</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script type="text/javascript">
+
+var cdy = CindyJS({ // See ref/createCindy documentation for details.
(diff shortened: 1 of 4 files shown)


gagern:
I managed to re-create the diff here (by force-pushing an older commit from this branch and then pushing the tip again). So now that #393 got merged (thanks for that!), this one here has become a lot leaner and easier to review.

strobelm:
Looks fine and provides conic tracing. Thanks @gagern!

PR #406, opened 2016-07-14
Node 6.3 fixed string_decoder bug

gagern:
So we adjust the warning to only warn about 6.2.x for x >= 1.
We also revert the Travis instructions to use the latest version again.


kranich:
@gagern Thanks for following up on the string_decoder bug.

PR #409, opened 2016-07-15
Move animation settings to their own subobject

gagern:
Now we have `animation.autoplay` and `animation.controls` (i.e.  properties of a newly introduced `animation` subobject) instead of just `autoplay` and `animcontrols` as options in the CindyJS invocation.
This adds structure, keeping related settings together.
It is expected that a new setting `speed` will be added to that object soon.

The old settings are still being used if the `animation` object is absent, so this is not a breaking change.


strobelm:
Code looks fine, examples work. Merged.

PR #411, opened 2016-07-20
Avoid isZero check if divisor is not of type number

gagern:
@strobelm, this fixes the exception you reported. We forgot to think about that possibility in #177.


strobelm:
Great thanks Martin! The branch had some problems running the examples, worked only after a rebase to master.

PR #412, opened 2016-07-21
Improve validity of HTML in examples

gagern:
These examples contain invalid HTML which gets detected when trying to include them into our website. We should fix them. And perhaps at one point add a Travis check to ensure that examples are valid.


montaga:
I checked the examples and they still look good. Thank you for validating the HTMLs!

PR #413, opened 2016-07-21
Deploy releases

gagern:
So far we only have code to deploy to the snapshot directory and branch. With this code in place, one can create an annotated (and preferably signed) tag of the form `v0.1.23` or similar, and the corresponding build will be deployed as a release.

I've tested a modified version of this, where I changed the `.travis.yml` to respond to my own repo, and `travis-deploy.sh` to not actually do the deployment but most of the other logic. Things look god so far. So please have a look at the code, and if that looks sound, let's give it a try using all the steps.


gagern:
I'll merge this myself since this is blocking some other work. Still I'll value input on whether the changes here are understandable or should be made more maintainable in some way.

PR #418, opened 2016-07-29
mousemove in setup.js

strobelm:
Hi currently scripts in mousemove aren't executed since compilation of the script is omitted. This should fix the problem.


kranich:
@strobel Thanks for the fix.

PR #423, opened 2016-08-01
Support for video

montaga:
With this pull request movies can be loaded using the image-Attribute. To me and Jürgen this seems reasonable as it is basically the same datastructure in CindyJS.

Loaded movies can be played and paused through CindyScript using the `playvideo` and `pausevideo` command.

2 examples (without and with CindyGL) are included showing a trailer of the Sintel movie. (4.4 MB)

According to https://durian.blender.org/sharing/, the movie, licensed under the Creative Commons Attribution 3.0 license, of the Blender Foundation can be freely reused if includes the entire credits scroll.


src/js/Setup.js
```diff
@@ -323,7 +323,14 @@ function createCindyNow() {
...
     if (typeof obj === "string") {
-        img = new Image();
+        var ext = obj.split('.').pop().toLowerCase();
+        if (ext === "mp4" || ext === "webm" || ext === "ogg") {
+            img = document.createElement("video");
+            img.preload = "auto";
+            img.loop = true; //loop videos as default
+        } else {
+            img = new Image();
+        }
```
gagern: I'm *slightly* unhappy with the fact that you base the decision on whether it's a video or a still image on the file extension. In the web file extensions are just a convention, not a requirement. MIME types are the thing that really counts. Which means that there may be quite a number of URLs out there which do not contain such an extension, or any extension at all, and which are still video.

Would it make sense to support `videos` as a parallel data structure to `images`, with the understanding that the former contains only videos, the latter only images, and if the creator of the widget doesn't know which one it is it's his responsibility to figure it out somehow? Internally it would all end up in the same collection of images.

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,19 @@ evaluator.cameravideo$0 = function() {
...
+    return nada;
+};
+
+evaluator.pausevideo$1 = function(args, modifs) {
+    var img = imageFromValue(evaluateAndVal(args[0]));
+    if (img.live) {
+        img.img.pause();
+    }
+    return nada;
+};
```
gagern: I'd like to see some duck type check on whether `img.img.play` and `img.img.pause` actually exist before calling them. For example, a texture containing the result of processing some WebGL video would be live as well, in the sense that it should not be cached but must be redrawn. Nevertheless, such a texture doesn't have `play` or `pause`. So simply do `if (img.live && img.img.play)`.

.gitattributes
```diff
@@ -11,6 +11,7 @@
 *.json          text
 *.js            text
 *.md            text
+*.mp4		-text
 *.pages         -text
```
gagern: Please indent consistently.

gagern:
Can you please either look for some example video which is *a lot* smaller, or reference some off-site video? I don't think adding 4+ MiB of pretty much incompressible data to the repository just to demonstrate this feature is warranted. Please use `git rebase -i master` to squash your commits and ensure that the big file is actually gone from the history of this pull request. If you absolutely want a beautiful video in a location we control, we can think about adding it to cindyjs.org somehow, but I'd hope that's not neccessary.

montaga:
Now I have replaced the video with an much smaller video from http://techslides.com/sample-webm-ogg-and-mp4-video-files-for-html5 ("Licence": Feel free to use the following files for whatever you need)

montaga:
Now https://github.com/bfred-it/iphone-inline-video will be also automatically included as lib by installing it as npm-dependency and adding its file `dist/iphone-inline-video.browser.js` as source. Please check whether this has been done properly.

package.json
```diff
@@ -30,7 +30,8 @@
     "st": "^1.2.0",
     "touch": "^1.0.0",
     "whole-line-stream": "^0.1.0",
-    "yauzl": "^2.6.0"
+    "yauzl": "^2.6.0",
+    "iphone-inline-video": "^1.9.3"
```
gagern: I've kept this list sorted alphabetically for easier maintainance. So please insert the new line appropriately. (This has the benefit of this PR just modifying one line, not two.)

gagern:
The way the library was included looks good to me. I'm a bit unhappy about the `global.navigator` setting all over the unit tests, but I can't think of a better solution either.

gagern:
Note that according to the [Notes about iOS 10](https://www.npmjs.com/package/iphone-inline-video#notes-about-ios-10) we should add an additional attribute to our video tags:

    img.setAttribute("webkit-playsinline", "");

Probably this affects camera as well. Can't test just now.

montaga:
I also don't have an iphone to test this. If you have access to one, please test it and tell whether it works. Hence I have added @richter-gebert

gagern:
@montaga This branch has diverged from master and will cause conflicts. Can you please either rebase on current master or merge current master?

montaga:
Do not merge yet. Still there are unresolved conflicts...

montaga:
I don't know how to make travis happy :( On my machine, `make alltests` throws no error.

gagern:
I can reproduce the Travis issue, by doing a `git clean -idx` first to clean up the `node_modules` in particular. The problem is with the iphone-inline-video@1.9.6 code, namely [these lines](https://github.com/bfred-it/iphone-inline-video/blob/v1.9.6/index.js#L8-L9). They got introduced in https://github.com/bfred-it/iphone-inline-video/commit/abd749d591602e33b531e830a5445529bc029325 and affect version 1.9.6 and up.

So we might fix the version to 1.9.5 to avoid the issue. Actually fixing a version like this has the benefit that it makes our code more reproducible: the same commit should always lead to the same result, even several years from now. If we want to update the code we should make that a conscious decision, and ensure that everything works with the new version in question.

montaga:
Thank you, Martin, for isolating the issue. :+1: 
It turned out that there still is some problem with version 1.9.5 (Here `window` is accessed). But v1.9.4 works.
Before merging this, still someone with iphone/ipad should test this.

PR #425, opened 2016-08-01
Support mouseclick event script

gagern:
While #418 fixed mousemove, we still don't have the mouseclick event which Cinderella supports. Here it is.


strobelm:
Oh i missed that when adding mousemove, thanks @gagern!

PR #426, opened 2016-08-01
playtone and playmelody

gagern:
This edit adds support for the [`playtone`](http://cindyjs.org/ref/MIDI_Functions.html#playtone$1) and [`playmelody`](http://cindyjs.org/ref/MIDI_Functions.html#playmelody$1) functions, using a plugin which gets loaded automatically if either of these functions is being used. It builds on a patched version of MIDI.js, althoug I envision using something different in the long run.

Sound fonts have to be loaded before midi becomes available. The implementation chose to simply ignore requests which cannot be satisfied (yet). So widgets which were simply exported from Cinderella may not play the sounds as expected, although if they continue issuing play commands eventually the sound fonts should arrive and actual playback should start. IN order to avoid this problem, the midi plugin introduces a new function called `loadinstruments` which accepts one or more instrument numbers and starts loading the corresponding instruments. It returns true if all the requested instruments have been loaded already.


gagern:
I just assigned this to myself since I'd like to get rid of MIDI.js before this enters master.

montaga:
Right now some of the examples do not work, because the soundfonts are not uploaded to cindyjs.org . For example https://cindyjs.org/extras/midi-js-soundfonts/MusyngKite/percussio gives a 404.

montaga:
This PR has been moved to #652.

PR #428, opened 2016-08-11
evaluate angles

strobelm:
CindyJS doesn't parse angles the same Cinderella does, e.g.
```
α0/180°*pi;
```
leads to unexpected behavior.  I modified the evaluator function in order to parse angles, which is a crucial part of CindyJS and needs thorough review (including side effects this might have).


strobelm:
And here is the updated version using evaluateAndVal.

PR #429, opened 2016-08-11
decrease plot step size

strobelm:
This decreases the step size in plot. Although we wanted to rewrite plotting anyway this small patch should increase the user experience while we are working on it. The pictures show the example from PR #428.

before:
plot2
after:
plot1


kranich:
@strobelm Fair enough. Could you please create and reference a follow up issue for the function plot rewrite?

PR #431, opened 2016-08-11
respect isvisible property for texts

strobelm:
isvisible is currently ignored for texts, one example is 146_angleFromScript.html


kranich:
Thanks, @strobelm for fixing this.

PR #432, opened 2016-08-16
rename setTimeout

### LINKED ISSUE #427: setTimeout in current Cindy.js
I just tried to update all my Cindy.js files to the current version. I get the following error:
`Called undefined function settimeout (as settimeout$2)`

Note that I, of course, called the function as `setTimeout(2, ...);`

Something seems to be wrong with capitals in function names...

strobelm:
This renames setTimeout to lower case and should resolve #427. I think there are no other functions which require an update.


kranich:
Thanks.

PR #433, opened 2016-08-17
csv support

strobelm:
Hi,

i partially rewrote the csv support. 

This will supersede #302.


src/js/libcs/Operators.js
```diff
@@ -4530,3 +4530,68 @@ evaluator.dropped$0 = function() {
...
+    var mcon = evaluateAndVal(modifs.autoconvert);
+    if (mcon.ctype === "boolean") aucon = mcon.value;
+
+    var delim = "";
+    var md = modifs.delimiter;
+    if (md && md.ctype === "string" && md.value.length === 1) delim = md.value;
+
+    var nl = "";
+    var mnl = modifs.newline;
+    if (mnl && mnl.ctype === "string" && mnl.value.length === 2) nl = mnl.value;
```
gagern: Why this length? I assume that the intention is to allow CRLF, but do we really benefit from disallowing a bare CR (Mac OS 9 style)? If the only valid value (other than the default) is CRLF, I'd prefer `mnl.value === "\r\n"` to make that explicit.

ref/String_Operators.md
```diff
@@ -194,6 +194,29 @@ defines the function `f(x)` to be `sin(x)+cos(x)`.
...
+    > abc,5.0
+    > asd,1.3
+    > sdf,4.3
+    > lala,2.3
+    > Barrr,1.7
+    > Baz,4.7");
+    < [[Foo, 1], [Bar, 2.3], [abc, 5], [asd, 1.3], [sdf, 4.3], [lala, 2.3], [Barrr, 1.7], [Baz, 4.7]]
+
+Numbers and Booleans are converted to their respective CindyScript counterparts, if this is not the desired behavior use the `autoconvert` modifier and set it to `false`.
+The modifiers `delimiter` and `newline` can be used to set the separating and linebreak characters.
```
gagern: What are the defaults? [RFC 4180](https://tools.ietf.org/html/rfc4180) uses CRLF but points out that “implementors should be aware that some implementations may use other values”. Common sense tells me that while *parsing* CSV it would make sense to allow all CRLF, CR and LF as end of line terminators (except inside a quoted string).

src/js/libcs/Operators.js
```diff
@@ -4530,3 +4530,68 @@ evaluator.dropped$0 = function() {
...
 };
+
+evaluator.parsecsv$1 = function(args, modifs) {
+    // more strict float parsing
+    var filterFloat = function(value) {
+        if (/^(\-|\+)?([0-9]+(\.[0-9]+)?|Infinity)$/
+            .test(value))
+            return Number(value);
+        return NaN;
+    };
```
gagern: What is the intention here? In particular, is it deliberate to disallow `1e17`, `0.`, `.5` or `NaN`?

Furthermore, is there any reason not to write this conversion the same as the other autoconversions, i.e. as `if ((/…/).test(itm) && aucon) { data[i][j] = … }` in the case distinctions below? Would make things more consistent in my opinion.

ref/String_Operators.md
gagern: Do we need another modifier to control the quoting character? How about escape sequences? I have the MySQL [`LOAD DATA INFILE`](http://dev.mysql.com/doc/refman/5.6/en/load-data.html) construct in mind here, which does offer this level of control. I'm far from certain it's actually needed here, though, so this is really just a comment, not a request to add more modifiers for this PR here.

tools/prepare-deploy.js
```diff
@@ -37,6 +37,7 @@ var handlers = {
     "ours.js": false,
     "ours.js.map": false,
     "pako.min.js": true,
+    "papaparse.min.js": true,
```
gagern: This feels to me like a pretty heavy dependency for what should be an easy task. Unless I'm mistaken, it should be possible to obtain the next field from a string using a regular expression like

    /"(?:[^"]+|"")*"|[^",\r\n]*/

As you apparently do auto-conversion manually, what exactly does papaparse give us in addition to tokenizing the string according to this or a similar regular expression? Is the gain worth the hassle of including an extra dependency?

src/js/libcs/Operators.js
```diff
@@ -4530,3 +4530,68 @@ evaluator.dropped$0 = function() {
...
+    var mnl = modifs.newline;
+    if (mnl && mnl.ctype === "string" && mnl.value.length === 2) nl = mnl.value;
+
+    var papaconfig = {
+        delimiter: delim,
+        newline: nl
+    };
+
+    var str = evaluateAndVal(args[0]).value;
+    var parsed = window.Papa.parse(str, papaconfig);
```
gagern: This will throw an exception if `Papa` is not available, e.g. because the `parsecsv` call was not part of one of the standard scripts but instead used the script of a button or passed to the widget via `evokeCS`. Unlikely scenarios, but in my opinion not unlikely enough to avoid handling them since exceptions can have pretty drastic effects in CindyScript evaluation.

src/js/libcs/Operators.js
```diff
@@ -4530,3 +4530,68 @@ evaluator.dropped$0 = function() {
...
+    var nl = "";
+    var mnl = modifs.newline;
+    if (mnl && mnl.ctype === "string" && mnl.value.length === 2) nl = mnl.value;
+
+    var papaconfig = {
+        delimiter: delim,
+        newline: nl
+    };
+
+    var str = evaluateAndVal(args[0]).value;
```
gagern: This assumes that the argument will evaluate to a string. Passing other values may lead to unexpected results.

src/js/libcs/Operators.js
```diff
@@ -4530,3 +4530,68 @@ evaluator.dropped$0 = function() {
...
+    // convert to CS*
+    var itm, pitm;
+    for (var i = 0; i < data.length; i++) {
+        for (var j = 0; j < data[i].length; j++) {
+            itm = data[i][j];
+            pitm = filterFloat(itm);
+            // Numbers
+            if (!isNaN(pitm) && aucon) data[i][j] = CSNumber.real(pitm);
+            // Bools
+            else if ((/[Tt]rue/).test(itm) && aucon) {
```
gagern: This will match any item *containing* the text `true`, even if there is more text before and after that.

examples/141_csv.html
```diff
@@ -0,0 +1,89 @@
...
+Bar;False
+abc;true
+asd;false
+sdf;4.3
+lala;2.3
+Barrr;1.7
+Baz;4.7";
+
+println(bools);
+cbool = parseCSV(bools);
```
gagern: Shouldn't this be using the `delimiter` modifier?

gagern:
Personally I'd prefer the library-less approach from #434 over this one here, so please have a look and let me know what you think.

PR #435, opened 2016-08-18
Detect widget resizes even if not caused by window resize

gagern:
This adapts the resize sensor from the CSS Element Queries project, written by Marc J. Schmidt and others.  Source of the adapted code: https://github.com/marcj/css-element-queries/blob/bfa9a7f/src/ResizeSensor.js

The idea is that although there is no resize event on the DOM box level in HTML 5, it is possible to (ab)use the scroll event instead.  An extremely large box scrolled to only show its bottom right portion will change its scroll position if its parent container expands, in order to not leave a gap.  Conversely a box sized 200% of its parent will be scrolled if its parent shrinks since it shrinks twice as fast.  Combined these two will detect size changes using some hidden DOM elements nested inside the already existing positioning container that also contains our canvas.

This is probably the best solution to https://github.com/CindyJS/website/issues/13.

### DIFF
--- a/examples/resize.html
+++ b/examples/resize.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Responsive widget resizing</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<style type="text/css">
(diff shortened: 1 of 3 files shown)


strobelm:
Works well (tested on Linux (Chromium) and Android (Firefox)). Thanks Martin!

PR #439, opened 2016-08-22
Implement algorithm and other geometry interaction

gagern:
This implements the algorithm function which returns the name of the geometric algorithm defining a geometric object.  The returned names are designed to be compatible with Cinderella as far as possible.

Remember: also tick item in #219 once this gets merged.


gagern:
@kortenkamp wrote per mail that he'd value compatibility between `algorithm` and `create` over compatibility with Cinderella. I don't think we want to make `create` completely compatible with Cinderella, since we made a number of fundamentally different design decisions in CindyJS so that's not merely a matter of renaming. So I guess we should return the algorithm in the CindyJS fashion, perhaps allowing for a `compatibility->"cinderella"` modifier or some such to obtain the other representation. Stand by while I rewrite the pull request…

gagern:
OK, the scope of this PR has increased somewhat: while editing the documentation (in order to mark `algorithm` as supported) I wanted to add a test case. I noticed however that this whole file was currently skipped during tests (because it had only inputs no outputs). Changing that made a number of other tests run and fail. Instead of disabling them, I set out to fix them as well. I think the result is a great step forwards in terms of compatibility and feature completeness.

We now have `inputs$1`, `moveto$2`, more compatible functionality in `create$3`, the `.slope` accessor readable for all lines and writable for `Through`, less chances of type mismatch when setting properties via accessors, and the nice feature of providing a geometric context for reftests. If this is a bit much to view as a combined diff, you may want to concentrate on the individual commits, which separate these issues quite nicely. Each of them should be fairly easy to check.

strobelm:
Thanks @gagern for the huge effort! Code looks fine, example behaves well. Merged.

gagern:
Thanks for the really quick review here, @strobelm!

PR #440, opened 2016-08-22
Implement text operator to convert argument to string

gagern:
In order to be compatible with Cinderella, the code does cast its argument to a value if it is a geometric object.  This is in sync with current Cinderella behavior, but the code there is annotated with a comment, so this might change again in the future.

Contrary to Cinderella behavior, we always return a string, even if the argument is nada.  In many cases this makes no difference, but in those cases where it does make a difference, this behavior is arguably the more useful one, so using it avoids extra case distinctions.

Both these decisions are of course open to discussion.

### DIFF
--- a/ref/String_Operators.md
+++ b/ref/String_Operators.md
@@ -27,10 +27,30 @@ For this see [Arithmetic Operators](Arithmetic_Operators.md).
 
 #### Conversion to string: `text(‹expr›)`
 
-**Not available in CindyJS yet!**
-
 **Description:**
 The operator `text(‹expr›)` evaluates the expression `‹expr›` and converts the result to a string representation.
+
+    > text(7.2)
+    < "7.2"
+    > text([true, "foo", (;)])
+    < "[true, foo, ___]"
+
+Undefined input yields undefined output in Cinderella:
+
+    - only Cinderella
+    > text(;)
+    < ___
+
+In CindyJS the output is still a string, in order to avoid corner cases.
+
+    - only CindyJS
+    > text(;)
+    < "___"
+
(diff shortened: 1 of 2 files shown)


kortenkamp:
At first I thought that it is reasonable to return a string even if the argument is Nada. But just while writing this I suddenly doubt it - calling a function on nada should map to nada, usually. This avoids distinctions in CindyScript code, and that's where it should be easy, shouldn't it?

gagern:
I guess it depends on use cases.

For example: in Cinderella `err(text(a)+b)` will print `___` if both `a` and `b` are undefined. It will print `___3` if `b="3"` but it will print `___` if `b=3`. That's because the `+` between nada and string is defined, but between nada and number is not. This feels weird to me: after all, I *told* the system to turn the first variable into a string, so why don't I see that concatenated with the second variable?

Most internal functions that expect strings will use `niceprint` internally, thus converting nada to `"___"`. The addition between string and nada is an example of this. It feels very weird that the one function explicitely designed to convert its input to text is the one not to do so in this special case.

Do you have a use case in mind where having `text` return undefined is actually useful, @kortenkamp?

I would agree with the “calling a function on nada should map to nada” concept *if* that were also true for all the essential arithmetic operators, and all the functions that expect strings and auto-convert their arguments to string otherwise. But since those convert nada to `"___"` automatically, I consider the pattern broken already.

kortenkamp:
I agree with "broken already". I never thought it is a good idea to output ___ in any case... so, ok, convert Nada to "".

gagern:
Why `""` instead of the `"___"` used by everything else?

strobelm:
Thanks for implementing this!

PR #441, opened 2016-08-23
fix List.almostequals

strobelm:
Hi,

we have broken recursion in List.almostequals. This should fix the bug.

### DIFF
--- a/ref/Lists_and_Linear_Algebra.md
+++ b/ref/Lists_and_Linear_Algebra.md
@@ -57,8 +57,33 @@ This is sometimes convenient to have a more mathematical appealing notation in t
 A three-dimensional vector may then be written as follows:
 
 `(7.3,9.3,-14.3)`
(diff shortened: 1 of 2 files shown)


gagern:
Thanks a lot! Could you add a reftest to this, to exhibit the previous bug and serve as a regression test?

strobelm:
Done. I'm note quite sure if the ref tests are positioned at the best location, feel free to move them around.

gagern:
I've moved the reftests to the *Boolean Operators* chapter. We definitely should have the info there, since that's where I'd start looking for details on the behavior of a boolean operator, and that is also where we have code in place which is useful for testing the approximate comparison.

Nevertheless, having a pointer in *Lists and Linear Algebra* would be useful as well. But reading that page I found a number of possible improvements, so I think I'll make that a pull request in its own right.

PR #442, opened 2016-08-23
Support newlines in text

gagern:
This adds support for newlines in text, for drawtext calls, Text elements and Button elements, both with and without the KaTeX plugin.

There are a number of unresolved issues yet:

1. The line height factor was measuered from Cinderella, but might be font-dependent, in which case we might want to measure some specific text in an attempt to approximate that value in a better way.

2. Buttons are aligned by their last row of text, where Cinderella aligns them by their first row.

3. The actual height of a TeX formula isn't taken into account when advancing to the next row, making the combination of KaTeX and multi-line text unsuitable for formulas much higher than the regular run of text.

All of these are mentioned in comments inside this code. Fixing them might require considerable additional work, and very likely some new ideas as well, so I propose merging this as it is, and leaving further improvements for future commits.

Should we document the issues with vertical alignment of buttons somewhere, so that we can change that in the future without breaking guaranteed behavior? I fear not, because on the one hand I don't know of any applicable documentation users would actually read, and on the other hand I fear that even then if they wanted multi-line buttons they'd still use a case distinction and risk breakage by a future version. So I think if we get to do vertical alignment the Cinderella way we should do that in a major version bump. The TeX formula height might be a different matter because affected formulas will likely look severely broken. But there, too, people might do case distinctions and add additional newlines, so we should probably do the same major version bump for that as well. Are you comfortable with this?

### DIFF
--- a/plugins/katex/src/js/katex-plugin.js
+++ b/plugins/katex/src/js/katex-plugin.js
@@ -257,7 +257,6 @@
             if ((i & 1) === 0) {
                 if (text.indexOf("\n") !== -1) {
                     var rows = text.split("\n");
-                    console.log(rows);
                     element.appendChild(document.createTextNode(rows[0]));
                     for (var j = 1; j < rows.length; ++j) {
                         element.appendChild(document.createElement("br"));
--- a/plugins/katex/src/js/katex-plugin.js
+++ b/plugins/katex/src/js/katex-plugin.js
@@ -148,22 +148,21 @@
 
     var firstMessage = true;
 
-    function katexRenderer(ctx, text, x, y, align) {
-        var fontSize = /(?:^| )([0-9]+)px(?:$| )/.exec(ctx.font);
-        fontSize = fontSize ? +fontSize[1] : 16;
-        var key = fontSize + ":" + text;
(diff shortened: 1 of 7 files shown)


strobelm:
Hi, just for comparison: should the output look like this? Especially for the C5 in the non-katex version looks misplaced. 

text

gagern:
Yes, that's as it should be. The point here is that there is a newline between `$C` and `_4$`. Which should count as a newline if we don't have enabled any special handling for TeX formulas, but which should be treated as ignorable whitespace in TeX formatting. So the two versions should behave differently, as they do.

plugins/katex/src/js/katex-plugin.js
```diff
@@ -235,10 +253,23 @@
         while (element.firstChild)
             element.removeChild(element.firstChild);
         for (var i = 0; i < n; ++i) {
+            var text = parts[i];
             if ((i & 1) === 0) {
-                element.appendChild(document.createTextNode(parts[i]));
+                if (text.indexOf("\n") !== -1) {
+                    var rows = text.split("\n");
+                    console.log(rows);
```
strobelm: Probably a forgotten console.log

strobelm:
No further remarks. Merged.

PR #443, opened 2016-08-23
Reduce default line overhang to 1

gagern:
The overhang of `1.1` will be useful only in a minority of situations, where setting it explicitely would be the right thing to do.  In general people clipping their lines will expect to see said lines drawn as segments with no overhang.  This is also the default Cinderella uses, and which it assumes when exporting, too.

The code change is trivial, so this is mostly a political decision since it involves a breaking change.


gagern:
Since we will likely break backwards compatibility for the sake of #467, we might pick up this one here along the way so we just have to bump version once and won't have to worry whether this here is fixing a bug or changing a feature. So now I'm looking for actual review on this one line change. 😃

strobelm:
Let's create precedents (perhaps we should think about a career in politics?).

PR #444, opened 2016-08-24
Update Closure Compiler to the latest version (v20160822)

kranich:
This is the upcoming release of Closure Compiler.

`--rewrite_polyfills` is now on by default. This caused some problems with CindyGL that I could fix by explicitly deactivating `--rewrite_polyfills` for CindyGL.

I have tested some examples of CindyJS core, Cindy3D, and CindyGL, all of which seem to work now.

### DIFF
--- a/make/Settings.js
+++ b/make/Settings.js
@@ -18,7 +18,7 @@ module.exports = function Settings() {
         closure_urlbase: "http://dl.google.com/closure-compiler",
         closure_language: "ECMASCRIPT5_STRICT",
         closure_level: "SIMPLE",
-        closure_version: "20160713",
+        closure_version: "20160822",
         verbose: "true",
         logprefix: "true",
         c3d_closure_level: "ADVANCED",
--- a/make/build.js
+++ b/make/build.js
@@ -363,6 +363,7 @@ module.exports = function build(settings, task) {
             dependency_mode: "LOOSE",
             create_source_map: "build/js/CindyGL.js.map",
             compilation_level: this.setting("cgl_closure_level"),
+            rewrite_polyfills: false,
             warning_level: this.setting("cgl_closure_warnings"),
             source_map_format: "V3",
             source_map_location_mapping: [


gagern:
Can you be specific about the kinds of problems polyfill rewrites caused for you? So that we know what to look for in the future to see whether these issues have been fixed? I tried the code without the flag, and could't see any issues immediately.

kranich:
>Can you be specific about the kinds of problems polyfill rewrites caused for you? So that we know what to look for in the future to see whether these issues have been fixed? I tried the code without the flag, and could't see any issues immediately.

For example, in [examples/cindygl/08_julia.html](https://github.com/kranich/CindyJS/blob/e8849fc5e15ed52c9913e61f2cddac25328a59de/examples/cindygl/08_julia.html) (probably in all CindyGL examples), the following errors are thrown in Firefox:

```
TypeError: invalid 'in' operand ba CindyGL.js:2:408
Plugin CindyGL not found Cindy.js:465:1
Called undefined function colorplot (as colorplot$2) Cindy.js:203:474
```

The last one many, many times. And in Chromium (possibly a bit more helpful):

```
polyfill] :36Uncaught TypeError: Cannot use 'in' operator to search for 'String' in undefined(anonymous function) @ polyfill] :36(anonymous function) @ ShaderProgram.js:177
Operators.js:3899 Plugin CindyGL not found
809Setup.js:754 Called undefined function colorplot (as colorplot$2)
```

kranich:
Maybe @montaga wants to investigate this further?

gagern:
I looked at the code. The relevant portion can be cleaned up to something like this:

```js
var ba;
if ("undefined" != typeof window && window === this) ba = this;
else if ("undefined" != typeof global) ba = global;
else ba = this;
var ca = ["String", "prototype", "repeat"];
for (da = 0; da < ca.length - 1; da++) {
    var ea = ca[da];
    ea in ba || (ba[ea] = {});
    ba = ba[ea]
}
var fa = ca[ca.length - 1];
var ga = ba[fa];
var ha = ga ? ga : function(a) { … }
```

So this is an attempt to install the `String.prototype.repeat` method, but it fails because our wrapper does not maintain `this`. So out `this` is undefined, not the global object. I'll propose an alternate PR shortly, addressing this issue.

kranich:
Superseded by #450.

Issue #279, opened 2016-02-26
almostequals fails on nested lists


elkins0:
The `comp_notalmostequals`, `comp_almostequals`, and `List.almostequals` functions all fail on nested lists with an exception:
```
Object doesn't support property or method 'comp_almostequals'
```

1. The `comp_notalmostequals` function calls the `comp_almostequals`  function which calls the `List.almostequals` function which calls an undefined `List.comp_almostequals` function. The  `List.almostequals` function should call the `List.almostequals` function instead of `List.comp_almostequals`.
2. The parameter passing of modifs is inconsistent in both the `List.almostequals` and the `List.equals` functions. Both should receive and pass on the modifs parameter.
3. There are no test cases for nested lists for any of the failing functions.

gagern:
1. That bug is fixed by #441.

2. Why should these internal functions receive the CindyScript modifiers? What would they use them for? In particular since the fuzzy comparisons are usually called using infix operators, and one cannot pas modifiers to those, I don't see a use case for this.

3. That is still true. I though I had done something along these lines for 4d7c929f779fe7fc16475a2e0f3cf8f5558de128 or #446, but apparently not. I guess I will add it to #446.

elkins0:
@gagern "2.Why should these internal functions receive the CindyScript modifiers?"

I only see that the `modifs` parameter  is lost if `comp_almostequals` calls `List.almostequals` which could call `comp_almostequals` again. If that is not a possibility then there is no problem.

```javascript
function comp_almostequals(args, modifs) {
...
        var erg = List.almostequals(v0, v1); // modifs not passed
...
}

List.almostequals = function(a1, a2) {
...
            erg = erg && comp_almostequals([av1, av2], []).value; // modifs not available to be passed on
...
}
```

gagern:
Thanks for the clarification, @elkins0! `comp_almostequals` doesn't make use of its `modifs` argument. That is listed only for the sake of consistency. So there is no harm in not passing it along. Thus 2 is not a bug, just slightly confusing code. With 1 addressed and a fix for 3 included in 1bd084bbcc139a810e983652cd550e9e2d796586 for #446, I think having #446 merged should close this one.

elkins0:
[...]

PR #446, opened 2016-08-24
Documentation improvements

gagern:
Starting with the documentation for lists, I found a number of things worth improving not only there but throughout the documentation, so this PR here has a pretty wide scope now. If there are some changes that need more discussion, I can remove them from this PR and open a separate one for them. The first commit also contains some code changes, improving Cinderella compatibility.


gagern:
With the latest addition of examples, this also fixes #279.

PR #447, opened 2016-08-24
Compile examples during build

### LINKED ISSUE #445: CindyScript 3.0 compilation errors in examples
examples/Life2.html throws `Error compiling init script: Function call in indexing construct must be enclosed in parentheses at 15:6`.

This is a breaking change introduced by CindyScript 3.0 and should be fixed.

@gagern Can we automatically test whether all examples compile without such errors? If so, I think we should!

gagern:
This fixes #445 and implements the suggestion made there about compiling all the examples as part of our usual test suite.

I'm slightly unsure about the split between `commands.js` and `build.js` here. Usually the idea is that `commands.js` implements some reusable command, and `build.js` instantiates this by feeding it some parameters. This leads to a very readable high-level `build.js`: you read a few lines and understand the intended action. On the other hand, it has the drawback that people need to look in two places if they want to fully understand the operation of a given task.

Now in this case, I couldn't think of a way to fit the new `excomp` task into a reusable pattern. The combination of uncached module loading and file iteration together with the specific messages and counting behavior felt pretty special. On the other hand, having all of that functionality in `build.js` would have hidden the high-level overview behind a huge amount of technical details. So I split the functionality anyway, making sure to name all the involved files and the crucial loop over the scripts in `build.js`, but leaving everything else in `commands.js`. Hope this makes sense.

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -121,6 +121,26 @@ module.exports = function build(settings, task) {
     });
 
     //////////////////////////////////////////////////////////////////////
+    // Make sure all examples compile
+    //////////////////////////////////////////////////////////////////////
+
+    task("excomp", [], function() {
+        this.excomp(
+            "examples/**/*.html",
+            "src/js/libcs/Parser.js",
+            function(html, parser) {
+                var re = /<script[^>]*type *= *['"]text\/x-cindyscript['"][^>]*>([^]*?)<\/script>/g;
+                var match, count = 0;
+                while (match = re.exec(html)) {
+                    ++count;
+                    var res = parser.parse(match[1]);
+                    if (res.ctype === "error") throw res;
+                }
+                return count;
+            });
+    });
+
+    //////////////////////////////////////////////////////////////////////
     // Run test suite from reference manual using node
     //////////////////////////////////////////////////////////////////////
 
@@ -131,6 +151,7 @@ module.exports = function build(settings, task) {
     task("tests", [
         "nodetest",
         "unittests",
+        "excomp",
     ]);
 
     //////////////////////////////////////////////////////////////////////
--- a/examples/Life2.html
+++ b/examples/Life2.html
@@ -18,18 +18,21 @@ m=zeromatrix(n,n);
      );
    );
 
+mo(x):=mod(x-1,n)+1;
+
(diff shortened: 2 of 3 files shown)


make/build.js
```diff
@@ -121,6 +121,26 @@ module.exports = function build(settings, task) {
...
     //////////////////////////////////////////////////////////////////////
+    // Make sure all examples compile
+    //////////////////////////////////////////////////////////////////////
+
+    task("excomp", [], function() {
+        this.excomp(
+            "examples/**/*.html",
+            "src/js/libcs/Parser.js",
+            function(html, parser) {
+                var re = /<script[^>]*type *= *['"]text\/x-cindyscript['"][^>]*>([^]*?)<\/script>/g;
```
kranich: @gagern The character group matching inside the script tags seems erroneous. I suppose it should be `[^<]`?

kranich:
Other than my comment above, this looks fine. Thank you, @gagern.

I am on holiday for the rest of the week. Will look into it on Sunday evening or Monday again. Or you can merge this yourself if you see fit.

make/build.js
gagern: No, the `[^]` is intentional; it's meant to match any character at all, including newlines. JavaScript uses `/foo[^]bar/` where other languages might be using `/foo.bar/s` to make the `.` match newlines as well.

I'm using the non-greedy quantifier `*?`, which matches the shortest possible string. So as long as the script tags are properly formatted, this will ensure that we don't capture portions of the next script. Looking at [the HTML spec for end tags](https://html.spec.whatwg.org/multipage/syntax.html#end-tags), the tag name might be using a different case, and there might be whitespace between that and the closing `>`. We could adjust the regexp for that, but I think that the way it is, we'd very likely get a false positive instead, since the combined code from both scripts and including the tags between them is unlikely to be valid CindyScript. So that error would just remind us to format the `</script>` properly, which [we need for the website](https://github.com/CindyJS/website/blob/4732edc4eb54d24d6942ec22088d528c4886d0d5/lib/examples.js#L29), too.

gagern:
I see no hurry, so take the credit for doing the review by doing the merge yourself. Thanks, @kranich!

kranich:
Thanks @gagern, ready to merge.

PR #448, opened 2016-08-24
Polygons in gslp

gagern:
This subsumes #173 and fixes various appearance aspects to make it compatible with what Cinderella currently exports. It also implements `TransformPolygon` as a geometric operation.
You can compare this aginst [the Cinderella file](https://github.com/CindyJS/CindyJS/files/435067/poly.zip) (renamed to `*.zip` to keep GitHub happy) from which the updated example was exported.

@strobelm I'm happy with those of your changes which I left in place. If you are happy with my changes, I think this can go in. Therefore I'm posting this as a pull request against master, not ayour feature branch. If this gets merged, #173 should be marked as merged, too.


strobelm:
Nice, thanks Martin! It seems like we have a drawing bug within shape and polygon drawing:

lineend

The bug is caused by the canvas lineJoin setting. I think there are three Options here:

1) increase the meterLimit, which makes most of the cases look fine but jumps in the case of a degenerate triangle. Perhaps we could handle this special case separately 

2) lineJoin = "round"
round

3) lineJoin = "bevel"
bevel

Any thoughts on that?

gagern:
I'd say this behavior is as Cinderella does it. Both Cinderella and this PR here use miter joins for polygons. The default miter limit is 10 both for Java `BasicStroke` and for CanvasRenderingContext2D, and as far as I can tell Cinderella doesn't change that anywhere. I deliberately went for miter joins in 60a7756f0b9e79b6825811bc41fb9f4f2f80717c.

I'd say that if we want to change this, we should agree to change it consistently for both Cinderella and CindyJS. Which means some more discussion, so I'd postpone that and make it a separate issue.

strobelm:
Since this is intentional, i'm fine with discussing the topic later. The rest look good, thanks Martin for extending my PR. This should also fix issue  #166.

PR #450, opened 2016-08-25
Update Closure Compiler to v20160822

gagern:
This addresses the issue identified in https://github.com/CindyJS/CindyJS/pull/444#issuecomment-242285587 and therefore works with active `--rewrite_polyfills` flag, at least in case of the julia set example. I haven't performed any other tests, though, so I think this should be tested some more before merging. At least to make sure that the examples that Stefan checked for #444 still work even with polyfills in place.


kranich:
@gagern All examples I checked for #444 work with f2a1bc6.

kranich:
This supersedes #444.

kranich:
@gagern Thanks for fixing this.

PR #451, opened 2016-08-25
Implement drawtable function

### LINKED ISSUE #193: drawtable
We should implement [the `drawtable` operator](http://cindyjs.org/ref/Texts_and_Tables.html#drawtable$2) in a way compatible to Cinderella.

gagern:
Defaults and size factors are taken from Cinderella.

This fixes #193 (although not in the way discussed there).
Don't forget to tick the function in #219 as well.

### DIFF
--- a/ref/Texts_and_Tables.md
+++ b/ref/Texts_and_Tables.md
@@ -150,8 +150,6 @@ It returns a boolean value.
 
 #### Drawing tables: `drawtable(‹vec›,‹list›)`
 
-**Not available in CindyJS yet!**
-
 One- and two-dimensional lists can be easily drawn in a geometric view as tables.
 In the simplest form one has only to provide the list and a position where the table should be drawn.
 Modifiers can be used to fine tune the appearance of the table.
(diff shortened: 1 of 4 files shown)


ref/Texts_and_Tables.md
```diff
@@ -150,18 +150,15 @@ It returns a boolean value.
...
 Modifiers can be used to fine tune the appearance of the table.
 
 **Example:**
 The following code produces the picture below:
 
-    - skip test: drawtable not implemented yet
-    > x=1..10;
-    > table=apply(x,(#,#^2,#^3,#^4));
+    > x=1..4;
+    > table=apply(x,(#,#^2,#^3));
```
montaga: This code does not produce build/ref/img/Table1X.png

montaga:
The code and the examples look fine to me. Nethertheless, the image in the references should match the code in the reference.

ref/Texts_and_Tables.md
gagern: I guess that I had thought about adding draw command tests here, and wanted to keep the number of commands low, but then decided against it again. By the way, this just made me think of #460, an idea which I've been toying with for quite some time now.

gagern:
I restored the previous code to match the figure.

PR #452, opened 2016-08-29
Document how to create a release

gagern:
Having a document like this may come handy at some point.


strobelm:
Great, thanks for writing this down!

PR #453, opened 2016-08-30
Transformation of Segments

strobelm:
Hi,

i finished the first implementation of segment transformation (projective and moebius). It's using the midpoint ansatz which becomes very handy in this situation. I haven't touched point on segment yet, since we need point on arc here (which is next on the todo list).

In geoOps.TransformS.updatePosition we are currently altering the farpoint according to the transformation. Is this really what we want to do here, or should we rather recalc the farpoint?

This will supersede #149.

### DIFF
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -619,6 +619,36 @@ Render2D.drawline = function(homog) {
     }
 };
 
+// draws a segment consisting of 2 rays
+Render2D.drawRaySegment = function(A, B, C, sflip) {
+    var ptA = eval_helper.extractPoint(A);
+    var ptB = eval_helper.extractPoint(B);
+    var ptC = eval_helper.extractPoint(C);
+    if (!ptA.ok || !ptB.ok || !ptC.ok) {
+        return;
+    }
+
+    var dx = sflip * (ptA.x - ptB.x);
+    var dy = sflip * (ptA.y - ptB.y);
+    var norm = Math.sqrt(dx * dx + dy * dy);
+
+    // get points outside canvas (at "infinity")
+    var sc = csport.drawingstate.matrix.sdet;
+    var farAway = 25000 / sc; // 25000px in user coordinates
+    var factor = farAway / norm;
+    dx = dx * factor;
+    dy = dy * factor;
+
+    Render2D.drawsegcore(ptA, {
+        x: ptA.x + dx,
+        y: ptA.y + dy
+    });
+    Render2D.drawsegcore(ptC, {
+        x: ptC.x - dx,
+        y: ptC.y - dy
+    });
+};
+
 Render2D.dashTypes = {
     "solid": [],
     "dashed": [10, 10],
--- a/src/js/libcs/Render2D.js
+++ b/src/js/libcs/Render2D.js
@@ -619,17 +619,16 @@ Render2D.drawline = function(homog) {
     }
 };
 
-// draws a segment consisting of 2 rays
-Render2D.drawRaySegment = function(A, B, C, sflip) {
+// draws a segment through infinity, consisting of 2 rays
+Render2D.drawRaySegment = function(A, B) {
     var ptA = eval_helper.extractPoint(A);
     var ptB = eval_helper.extractPoint(B);
-    var ptC = eval_helper.extractPoint(C);
-    if (!ptA.ok || !ptB.ok || !ptC.ok) {
+    if (!ptA.ok || !ptB.ok) {
         return;
     }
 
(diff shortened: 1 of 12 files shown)


gagern:
It's enough to have *either* the farpoint *or* the midpoint. Somehow when I wrote this, having the farpoint felt more natural to me, but I have no strong feelings there, so you might as well use the midpoint and drop the farpoint.

One minor benefit of having the farpoint is that it makes it easier to define the permissible range of values for the cross ratio: one can restrict the cross ratio to [0,1] which makes for a nicely finite range of values. Using [0,∞] as the range is much more awkward in terms of numerics, I think. Not sure whether computing the farpoint is the best solution here.

As I said in some video conference, I have a gut feeling that we may benefit from using just two vectors to represent all segments, by scaling the representatives so that their sum is the midpoint and their difference is the farpoint. I never got round to trying this out in code, but I think this might be a good opportunity now, particularly since you added rendering for line segments through infinity. I'll give that a try, and if I feel it works out I will probably create a PR against your branch here.

kortenkamp:
> Am 30.08.2016 um 14:36 schrieb Martin von Gagern <notifications@github.com>:
> 
> As I said in some video conference, I have a gut feeling that we may benefit from using just two vectors to represent all segments, by scaling the representatives so that their sum is the midpoint and their difference is the farpoint. I never got round to trying this out in code, but I think this might be a good opportunity now, particularly since you added rendering for line segments through infinity. I'll give that a try, and if I feel it works out I will probably create a PR against your branch here.
> 
> 

Only one question: I usually don't like adding or subtracting vectors in projective space. Can you make sure that this is stable against changing signs or scales of the originating (defining) vectors?

gagern:
> I usually don't like adding or subtracting vectors in projective space.

When creating a new Segment from its two endpoints, the previous three-point approaches would either compute the midpoint or the farpoint, or both, and store that with the object. That computation would be done by bringing both vectors to the same *z* coordinate and then adding or subtracting them. **So at that point we already have that operation.** If the line got transformed, the midpoint or farpoint would get transformed accordingly.

The only thing new here is that we kind of defer the addition resp. subtraction. We do it for each operation that requires it. The way to ensure consistency is by *always scaling both vectors together*. In the implementation the vectors form a 2×3 matrix which is normalized as a whole before being split into a pair of vectors. That way the sum and the difference scales with the same factor, and thus still represents the same point.

> Can you make sure that this is stable against changing signs or scales of the originating (defining) vectors?

If in a segment [*A*, *B*] the vector *A* were to be replaced by *λA*, then the endpoints would be computed as *B*<sub>*z*</sub>⋅*λA* and *λA*<sub>*z*</sub>⋅*B*, which is equial to the original definition except for a factor of *λ* common to all six coordinates, and very likely exactly equal to the original definition after joint normalization of both vectors together. Any common factor would be included in both midpoint and farpoint, but that's only changes in representatives.

src/js/libcs/Render2D.js
```diff
@@ -619,6 +619,36 @@ Render2D.drawline = function(homog) {
     }
 };
 
+// draws a segment consisting of 2 rays
+Render2D.drawRaySegment = function(A, B, C, sflip) {
+    var ptA = eval_helper.extractPoint(A);
+    var ptB = eval_helper.extractPoint(B);
```
gagern: `B` might actually be at infinity, turning `ptB` into numeric garbage (Infinities, NaN and the likes). For certain scenarios it's even likely to have the midpoint mapped to infinity, so this is not just a corner case with probability zero.

src/js/libcs/Render2D.js
```diff
@@ -619,6 +619,36 @@ Render2D.drawline = function(homog) {
...
+        return;
+    }
+
+    var dx = sflip * (ptA.x - ptB.x);
+    var dy = sflip * (ptA.y - ptB.y);
+    var norm = Math.sqrt(dx * dx + dy * dy);
+
+    // get points outside canvas (at "infinity")
+    var sc = csport.drawingstate.matrix.sdet;
+    var farAway = 25000 / sc; // 25000px in user coordinates
```
gagern: This is less than ideal, but better than what we have now. We should keep in mind that it would be nice to have a better solution for this one day. Particularly for infinite line segments whose endpoints happen to be more than 25000px outside the current viewport area, so that we don't see a ray that should be visible in theory.

gagern:
I've written two comments for the code, but one becomes obsolete if you merge strobelm#6 and the other is just a thing to keep in mind, not a showstopper. So I'm ready to merge this once strobelm#6 is reviewed and included.

strobelm:
I'm fine with the proposed changes, although the way is a bit unorthodox. Thanks Martin for the PR!

PR #454, opened 2016-08-31
Simulation time and speed

gagern:
In Cinderella, one animation frame advances the internal simulation time by a fixed amount, proportional to the selected simulation speed. The next frame is scheduled 20ms after the previous frame finished its computation. This means that if the time required for the computation of a single animation frame is not significantly less than these 20ms, then the apparent speed (as seen by the user) depends on that computation. This is what I call “framerate-controlled speed”, since the per-frame time advancement is fixed but the speed seen by the user depends on the frame rate.

Doing the same in CindyJS can be problematic. While `requestAnimationFrame` *might* result in approximately 50 calls per second (i.e. the inverse of 20ms), there is no guarantee to this effect. It may be far less on slow devices. It may be more on fast devices, and particularly so on future devices, which might conceivably run their display at e.g. 200Hz and be prepared to render animations at that same speed. So we have essentially three alternatives:

1. Don't use `requestAnimationFrame` for animations, but use `setTimeout` with those same 20ms between frames, to closely resemble Cinderella.
2. Use whatever framerate the browser supports, so that the same animation will run at very different apparent speeds depending on the device being used.
3. Use (a suitably scaled version of) actual elapsed time, so that simulations will run at the same speed no matter the device.

A problem with 3. is that animations may take computation time proportional to the span of simulation time to be simulated, thanks to refinements. So a complicated animation on a slow device may well end up taking longer for a computation than the span of wall time for which it is doing the computation. That way a time dept can be accumulated, conceivably causing the simulation to freeze.

The solution proposed here is a mixture of 2. and 3. For devices capable of computing 20fps or more, the approach is 3, to provide device-independent simulation speed. For slower devices, the actual time between frames is capped at 1/20 of a second, so that if the frame rate drops below 20fps, the animation speed drops with it in order to avoid total deadlock.

The benefit of devide-independent simulation speed comes with some drawbacks. One is reduced compatibility with Cinderella due to different semantics. For example, as long as the 20fps barrier isn't reached, simulations won't slow down in situations requiring more computation. Another possible problem is reproducibility: with simulation time increments derived from elapsed wall time, two runs of some theoretically deterministic simulation from the same starting position might lead to different outcomes due to different numeric behavior depending on the step sizes being used. This can be observed in the ball example included in this pull request.

At the moment I don't see how Cinderella's internal units map to those of CindyJS. Looking at the sources it would seem that Cinderella has a frame approximately every 20ms (at least with low computation overhead), and advances the internal simulation time by 5.0 times the current simulation speed (which defaults to 1.0 for new documents). So that would mean approximately 250 internal time units per second. The `simulationtime` function then divides this internal time by 360, so there would be 25/36=0.6944 script-visible simulation time units per second. But measuring the behavior of a falling ball, I'm seeing something closer to 22 internal time units per second, and 0.0127 script-visible units per internal one, or 0.2794 per second. Observations from the `150_simulationtime` example in Cinderella show that there are more like 0.3 units per second, so apparently I'm not reading the Cinderella code correctly. But at this time I have no real clue where this difference comes from, and whether my measurements are specific to the gravitational force I used in my setup. I've entered the numbers I observed as constants but I'm far from happy with this. Some further investigation of this aspect might be required.

On the other hand, I think that even as it stands now, this PR is a step in the right direction and should be merged. As it stands, it will add the following value to CindyJS.

* Includes a slider above the animation control buttons allowing users to control the speed
* Honors the speed setting exported by Cinderella
* At least some examples will run at speeds comparable to what Cinderella will exhibit
* Support for the `simulationtime` function

Regarding backwards compatibility: The `deltat` setting on the environment object gets translated into a call to adjust the simulation speed, mapping the old default `deltat: 0.3` to the default `speed: 1.0`. So code making use of `deltat` should work fine. I checked some examples which make use of this and they looked fine to me. Therefore I don't see this as a breaking change even if the speed might change somewhat for some simulations.

### DIFF
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -257,8 +257,17 @@ function createCindyNow() {
             csctx.setLineDash = function() {};
         if (data.animation ? data.animation.controls : data.animcontrols)
             setupAnimControls(data);
-        if (data.animation && isFiniteNumber(data.animation.speed))
-            setSpeed(data.animation.speed);
+        if (data.animation && isFiniteNumber(data.animation.speed)) {
+            if (data.animation.accuracy === undefined &&
+                data.cinderella &&
+                data.cinderella.version &&
+                data.cinderella.version[0] === 2 &&
+                data.cinderella.version[1] === 9 &&
+                data.cinderella.version[2] < 1880)
+                setSpeed(data.animation.speed * 0.5);
+            else
+                setSpeed(data.animation.speed);
+        }
         if (data.animation && isFiniteNumber(data.animation.accuracy))
             simaccuracy = data.animation.accuracy;
     }
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -524,7 +524,7 @@ function setupAnimControls(data) {
 var setSpeedKnob = null;
 
 function setSpeed(speed) {
-    simspeed = Math.max(0, speed); // do we want to allow negative speed?
+    simspeed = speed;
     if (setSpeedKnob) setSpeedKnob(speed);
 }
 
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -9,11 +9,40 @@ var cscompiled = {};
 var csanimating = false;
 var csstopped = true;
 var simtime = 0; // accumulated simulation time since start
-var simspeed = 1.0; // the animation.speed setting
-var simfactor = 0.022; // internal simtime units per millisecond at speed 1
-var simunit = 0.0127; // reported simulationtime per internal simtime
-var simcap = 1000/20; // max. ms between frames for fps-independent sim
+var simspeed = 0.5; // global speed setting, will get scaled for applications
+var simcap = 1000 / 20; // max. ms between frames for fps-independent sim
 var simtick = 0; // Date.now of the most recent simulation tick
+var simaccuracy = 10; // number of sub-steps per frame
+
+var simunit = 5 / 360; // reported simulationtime() per internal simtime unit
+/* Cinderella has a factor 5 for its internal animation clock,
+ * and the division by 360 is in the simulationtime function implementation.
+ */
+
+// internal simtime units per millisecond at simspeed 1
+var simfactor = 0.32 / simunit / 1000 * 2;
+/*              ^^^^ simulationtime per second, observed in Cinderella
+ *                     ^^^^^^^ simulationtime per simtime unit
+ *                               ^^^^ milliseconds per second
+ *                                      ^ default accuracy factor
+ *
+ * Cinderella does timing different from CindyJS, so here are some notes.
+ * The default in Cinderella is speed=1.0, accuracy=2, frames=1 in its terms,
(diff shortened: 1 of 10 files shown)


gagern:
I adjusted timing based on a deeper understanding of what Cinderella does. As a notable change, the default speed is now 0.5, since that's essentially what Cinderella is using by default if the accuracy setting there is at its default of 2 as well. See commits for details.

gagern:
Since this might break existing animations (by changing their speed or due to the clock-based vs. frame-based issue), and since we are planning a 0.8.0 release anyway, let's include this into that release but keep it out of the 0.7 line.

Which means that we might drop some of the code I have in here fore the sake of backwards compatibility. Namely `accuracy` and `deltat` which was in the environment behavior for historic reasons, but which is in the animation element now. Should we just drop them from the environment behavior?

src/js/Setup.js
```diff
@@ -421,6 +500,32 @@ function setupAnimControls() {
...
+    function speedUp(event) {
+        speedDragging = false;
+    }
+
+}
+
+var setSpeedKnob = null;
+
+function setSpeed(speed) {
+    simspeed = Math.max(0, speed); // do we want to allow negative speed?
```
strobelm: That's an interesting comment! I would love to be able to run simulations backwards.

src/js/Setup.js
```diff
@@ -5,8 +5,46 @@ var cslib;
...
+var simtick = 0; // Date.now of the most recent simulation tick
+var simaccuracy = 10; // number of sub-steps per frame
+
+var simunit = 5 / 360; // reported simulationtime() per internal simtime unit
+/* Cinderella has a factor 5 for its internal animation clock,
+ * and the division by 360 is in the simulationtime function implementation.
+ */
+
+// internal simtime units per millisecond at simspeed 1
+var simfactor = 0.32 / simunit / 1000 * 2;
```
strobelm: This seems to be very hardware dependent. On my machine (Intel(R) Core(TM) i7-4712HQ CPU @ 2.30GHz) a factor of 0.32 is way to high. The swarm for example looks like someone feed them some funny pills.

strobelm:
Hi,

the code looks fine in general. I won't merge the code as at least at my hardware the provisions for timing adjustment are not working properly.

src/js/Setup.js
gagern: This appears to be more of a problem with legacy widgets. In particular, when I changed the default speed from 1 to 0.5 I forgot to change the scale factor for backwards compatible time control (via the deltat parameter of the environment). My latest commit should take care of that.

But this only strengths the importance of the question I asked above: do we actually want this kind of backwards-compatibility (and maintain it throughout the 0.8 cycle at the least), or do we want to remove that now and force widgets using 0.8 to set speed in a modern way?

gagern:
> at least at my hardware the provisions for timing adjustment are not working properly.

What exactly do you mean? The fact that the swarm looked way too fast? Or something else?

strobelm:
Yes exactly, as this is fixed now there is no reason not to merge.

PR #455, opened 2016-09-02
Make gravitational proportional to mass

gagern:
At the moment, the force the environment's gravity exerts on each mass is not scaled by that mass. This causes heavy objects to fall more slowly than lighter ones, in the absence of friction. This change corrects that mistake.


strobelm:
Thanks for fixing this!

PR #456, opened 2016-09-04
Fix incorrect font for buttons

gagern:
The text size was set in two places, and since both of them did wrap the text size in a CSNumber after a cast to native number, this led to a text size of NaN which made the whole CSS font style specification invalid.

One notable consequence of this was the Buttons1 example using default fonts for both its buttons, which happened to be different for the push button and for the toggle button.


kranich:
@gagern Thanks for the fix.

gagern:
Thank you for the quick review!

PR #457, opened 2016-09-04
State IO

gagern:
This improves state saving for polygons and text elements, buttons in particular. Since MC<sup>2</sup> makes heavy use of state saving, this is something we urgently need there. The first commit adds a function to help debug this kind of issue, by saving the state and then restarting the instance from that saved state.

As the fix for #456 would conflict with these changes (since the line removed there is directly adjacent to the line where I renamed `pos` to `homog` for text elements), I merged that one here, so this PR is a superset of #456. But that should be a pretty clear bug fix.


src/js/Head.js
```diff
@@ -167,4 +167,22 @@ var CindyJS = (function() {
                 console.log(JSON.stringify(state));
             };
 
+            CindyJS.debugState = function(index) {
```
strobelm: Interesting! Perhaps we should collect how to use these hidden debugging features somewhere.

src/js/libgeo/StateIO.js
```diff
@@ -105,6 +110,10 @@ function saveGeoElement(el) {
         if (val !== null && val !== undefined)
             res[key] = val;
     });
+    if (el.kind === "P" && (!el.movable || el.pinned) && res.color) {
```
strobelm: I'm wondering why we're looking for not movable or pinned here.

strobelm:
Code looks fine, just minor questions from my side. However github reports merge conflicts.

src/js/libgeo/StateIO.js
gagern: This is the counterpart for [the dimming in `pointDefault`](https://github.com/gagern/CindyJS/blob/85374cf4386361c7f7004ba6fa48d532ce81f93e/src/js/libgeo/GeoBasics.js#L77-L79).

gagern:
The merge conflicts reported by GitHub were really strange! Merged on my system without conflicts, so now GitHub is happy as well.

PR #459, opened 2016-09-05
CindyGL: automatically force recompile if generation of function changes

montaga:


plugins/cindygl/src/js/Renderer.js
```diff
@@ -238,11 +245,21 @@ Renderer.prototype.loadTextures = function() {
...
+                console.log(fname + " is outdated; forcing rebuild.");
+                return false;
+            }
+        }
+        return true;
+    }
+    /**
+     * runs shaderProgram on gl. Will render to texture in canvaswrapper
+     * or if argument canvaswrapper is not given, then to glcanvas
+     */
```
gagern: The indentation here looks strange and confusing. Shouldn't all of this be outdented by 4 spaces? And perhaps a newline before the comment?

gagern:
It seems I was wrong about this being incompatible with the extra argument to `drawtext$2`. That's because the third argument here is only being used for infix operators, not for function invocations. Source looks good to me. Behavior is fine as well. Thanks a lot!

PR #461, opened 2016-09-06
Support calculations

gagern:
Calculations, i.e. the GSLP elements one can add with the *f(x)* button in Cinderella, are being properly exported since Cinderella beta build 1873. This adds CindyJS rendering support for these.

### DIFF
--- a/examples/Calculation.html
+++ b/examples/Calculation.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Calculations</title>
+    <link rel="stylesheet" href="../build/js/CindyJS.css">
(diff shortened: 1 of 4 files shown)


strobelm:
It would be great to have an example file here. Is there already one or should i generate one on my own? Code looks good, nevertheless i would like to test a real world example.

gagern:
I think I had an example for this somewhere, I'll see about adding that.

gagern:
The example I just attached originated from [`Calculation.cdy`](https://github.com/CindyJS/CindyJS/files/471976/Calculation.zip) (renamed to `*.zip` to make GitHub happy).

strobelm:
Thanks!
We should really take care of plot, we got some nasty rendering bugs there.

Issue #417, opened 2016-07-29
imagergb only implemented within colorplots


kortenkamp:
Christian Mercat ran into a problem with accessing data from images. He was puzzled by the fact that `imagergb` is only working within `colorplot`, not outside (probably due to CindyGL having it, but not CindyJS).

PR #462, opened 2016-09-07
added imagergb[a] outside of colorplot

montaga:
This should fix https://github.com/CindyJS/CindyJS/issues/417

This pull request implements the imagergb[a] command outside of colorplot.
- imagergb[a]$3 corresponds to the one that is known from Cinderella; both(!) returning a RGBA-vector with entries between 0-255 for the colors, and 0-1 for the alpha-value.
- imagergb[a]$4(‹point1›, ‹point2›, ‹image›, ‹point3›) returns a RGB[A]-vector for color at the coordinate ‹point3› assuming that the left/right lower corner is ‹point1›/‹point2› resp. All entries are in the interval [0,1].

### DIFF
--- a/examples/cindygl/34_cpuimagergba.html
+++ b/examples/cindygl/34_cpuimagergba.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html lang="en">
+   <head>
+      <meta charset="utf-8">
+      <title>Cindy JS</title>
+      <script type="text/javascript" src="../../build/js/Cindy.js"></script>
+			<script type="text/javascript" src="../../build/js/CindyGL.js"></script>
+   </head>
+   <body style="font-family:Arial;">
+      <h1>CindyJS: imagergba4</h1>
+			<script id='csinit' type='text/x-cindyscript'>
+				interpolate = true; repeat = false;
+				t0 = seconds();
+				createimage("test", 4, 3);
+			</script>
+      <script id='csmove' type='text/x-cindyscript'>
+         colorplot(A, B, "test", hue(.1*((seconds()-t0)+|#,A|)));
+				 colorplot(imagergba(A, B, "test", #, interpolate->interpolate, repeat->repeat));
+         col = imagergba(A, B, "test", C, interpolate->interpolate, repeat->repeat);
+         C.color = (col_1, col_2, col_3);
+         C.alpha = .5*col_4+.5;
+      </script>
+
+      <script type="text/javascript">
+         var gslp=[{name:"A", kind:"P", type:"Free", pos:[-8,-8,1]},
+                   {name:"B", kind:"P", type:"Free", pos:[8,-8,1]},
+                   {name:"C", kind:"P", type:"Free", pos:[0,3,1],size:10}];
+
+         cdy = CindyJS({canvasname:"CSCanvas",
+                     scripts: 'cs*',
+                     geometry:gslp,
+										 autoplay: true,
+										 use: ["CindyGL"]
+
+            });
+      </script>
+			<div id="CSCanvas" style="width:500px; height:500px;"></div>
+		  <p>
+		    <input onclick="cdy.evokeCS('interpolate = ' + this.checked);" checked type="checkbox" >interpolate bilinear&nbsp;&nbsp;&nbsp;
+		    <input onclick="cdy.evokeCS('repeat = ' + this.checked);" type="checkbox" >repeat
+		  </p>
+   </body>
+</html>
--- a/examples/cindygl/34_cpuimagergba.html
+++ b/examples/cindygl/34_cpuimagergba.html
@@ -4,18 +4,18 @@
       <meta charset="utf-8">
       <title>Cindy JS</title>
       <script type="text/javascript" src="../../build/js/Cindy.js"></script>
-			<script type="text/javascript" src="../../build/js/CindyGL.js"></script>
+      <script type="text/javascript" src="../../build/js/CindyGL.js"></script>
    </head>
    <body style="font-family:Arial;">
       <h1>CindyJS: imagergba4</h1>
-			<script id='csinit' type='text/x-cindyscript'>
-				interpolate = true; repeat = false;
-				t0 = seconds();
-				createimage("test", 4, 3);
-			</script>
+      <script id='csinit' type='text/x-cindyscript'>
+        interpolate = true; repeat = false;
+        t0 = seconds();
+        createimage("test", 4, 3);
+      </script>
       <script id='csmove' type='text/x-cindyscript'>
          colorplot(A, B, "test", hue(.1*((seconds()-t0)+|#,A|)));
-				 colorplot(imagergba(A, B, "test", #, interpolate->interpolate, repeat->repeat));
+         colorplot(imagergba(A, B, "test", #, interpolate->interpolate, repeat->repeat));
          col = imagergba(A, B, "test", C, interpolate->interpolate, repeat->repeat);
          C.color = (col_1, col_2, col_3);
          C.alpha = .5*col_4+.5;
@@ -29,15 +29,15 @@
          cdy = CindyJS({canvasname:"CSCanvas",
                      scripts: 'cs*',
                      geometry:gslp,
-										 autoplay: true,
-										 use: ["CindyGL"]
+                     autoplay: true,
+                     use: ["CindyGL"]
 
             });
       </script>
-			<div id="CSCanvas" style="width:500px; height:500px;"></div>
-		  <p>
-		    <input onclick="cdy.evokeCS('interpolate = ' + this.checked);" checked type="checkbox" >interpolate bilinear&nbsp;&nbsp;&nbsp;
-		    <input onclick="cdy.evokeCS('repeat = ' + this.checked);" type="checkbox" >repeat
-		  </p>
+      <div id="CSCanvas" style="width:500px; height:500px;"></div>
+      <p>
+        <input onclick="cdy.evokeCS('interpolate = ' + this.checked);" checked type="checkbox" >interpolate bilinear&nbsp;&nbsp;&nbsp;
+        <input onclick="cdy.evokeCS('repeat = ' + this.checked);" type="checkbox" >repeat
+      </p>
    </body>
 </html>
--- a/plugins/cindygl/src/js/General.js
+++ b/plugins/cindygl/src/js/General.js
@@ -292,6 +292,16 @@ function getPixelType() {
     else return gl.UNSIGNED_BYTE;
 }
 
+function toFloat(pixels) {
+    let res = [];
+    for (let i = 0; i < pixels.length; i++) {
+        if (can_use_texture_float) res.push(pixels[i]);
+        else if (can_use_texture_half_float) res.push(decodeFloat16(pixels[i]));
+        else res.push(pixels[i] / 255);
+    }
+    return res;
+}
+
 function smallestPowerOfTwoGreaterOrEqual(a) {
     let ans = 1;
     while (ans < a) ans <<= 1;
--- a/plugins/cindygl/src/js/General.js
+++ b/plugins/cindygl/src/js/General.js
@@ -250,45 +250,45 @@ var toByte = f => f * 255
 
 /**
  * converts a float array to an array encoded in the internal type
- * @param {Array<number>} pixels
+ * @param {Array<number>} samples
  */
-function createPixelArrayFromFloat(pixels) {
-    if (can_use_texture_float) return new Float32Array(pixels);
-    if (can_use_texture_half_float) { //return new Uint16Array(pixels.map(toHalf)); <- does not work in recent safari
-        let newpixels = new Uint16Array(pixels.length);
-        for (let i = 0; i < pixels.length; i++) {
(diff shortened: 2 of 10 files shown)


src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
     });
     return img;
 };
+
+var tmpcanvas; //temporary unvisible canvas.
```
gagern: Make that “invisible”.

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
...
+        var data;
+        if (img.img.getContext) { //img is a canvas
+            var ctx = img.img.getContext('2d');
+            data = ctx.getImageData(x, y, width, height).data;
+        } else { //copy corresponding subimage of img.img to temporary canvas
+            if (!tmpcanvas) {
+                //creating tmpcanvas only once increases the running time
+                tmpcanvas = /** @type {HTMLCanvasElement} */ (document.createElement("canvas"));
+                tmpcanvas.style.display = "none";
+                document.body.appendChild(tmpcanvas);
```
gagern: Do we need to append the canvas? Do you know of any actual problem where drawing to a canvas which is not included in the DOM would fail? Any specification saying so? See also [this Stack Overflow question of mine](http://stackoverflow.com/q/37658330/1468366).

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
...
+    var rgba = readPixelsIndirection(img, x, y, 1, 1);
+    return List.realVector([rgba[0] * 255, rgba[1] * 255, rgba[2] * 255, rgba[3]]);
+};
+
+evaluator.imagergb$3 = evaluator.imagergba$3; //According to reference
+
+/**
+ * imagergba(<point1>, <point2>, ‹image›, <point3>) returns the color at the coordinate
+ * <point3> assuming that the left/right lower corner is <point1>/<point2> resp.
+ */
```
gagern: Please add this to the reference manual, and remove the `not implemented yet` notice for the Cinderella-compatible function.

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
...
+        coord.y = (coord.y % h + h) % h;
+    }
+
+    var rgba = [0, 0, 0, 0];
+
+    if (interpolate) {
+        var i, j;
+
+        var xf = (coord.x % 1 + 1) % 1; //fractional parts
+        var yf = (coord.y % 1 + 1) % 1;
```
gagern: I find this hard to read. I understand that the `(… + 1) % 1` is designed to take care of situations where the initial remainder might have been negative. But since you will be using `Math.floor(coord.x)` in many places, why not make this more explicit:

```js
var xi = Math.floor(coord.x);
var xf = x - xi;
```

That way you can shorten the subsequent expressions and make this assignment here both easier to understand and faster to execute.

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
...
+                for (i = 0; i < 8; i++) pixels[i] = pixels[i + 8];
+            if (Math.floor(coord.y) === h - 1 && yf < 0.5)
+                for (i = 0; i < 8; i++) pixels[i + 8] = pixels[i];
+        }
+
+        //bilinear interpolation for each component i
+        for (i = 0; i < 4; i++)
+            for (j = 0; j < 4; j++) {
+                rgba[i] += pixels[4 * j + i] * ((j < 2) ? (1 - yf) : yf) * ((j % 2 === 0) ? (1 - xf) : xf);
+            }
```
gagern: Personally I think it would be easier to write this without a loop, but I don't want to force you if you disagree.

```js
rgba[i] = (1 - yf) * ((1 - xf) * pixels[i] + xf * pixels[i + 4]) +
    yf * ((1 - xf) * pixels[i + 8] + xf * pixels[i + 12]);
```

src/js/libcs/OpImageDrawing.js
```diff
@@ -385,3 +385,169 @@ evaluator.cameravideo$0 = function() {
...
+    } else {
+        var x = Math.floor(coord.x);
+        var y = Math.floor(coord.y);
+        if (isNaN(x) || isNaN(y)) return nada;
```
gagern: What about infinities? Why is there no such check for the interpolating code path?

montaga:
@gagern Thank you for this detailed review. Now I have fixed the issues you had addressed.

src/js/libcs/OpImageDrawing.js
```diff
@@ -388,3 +388,167 @@ evaluator.cameravideo$0 = function() {
...
+    if (!Number.isFinite(xi) || !Number.isFinite(yi)) return nada;
+
+    var rgba = [0, 0, 0, 0];
+    if (interpolate) {
+        var i, j;
+
+        var xf = coord.x - xi; //fractional part
+        var yf = coord.y - yi;
+
+        var pixels = readPixelsIndirection(img, xi, yi, 2, 2);
```
gagern: I was slightly surprised this does not fail for the WebGL case if the position is outside the picture, but it's [well specified](https://www.khronos.org/registry/webgl/specs/1.0/#READS_OUTSIDE_FRAMEBUFFER). :+1:

src/js/libcs/OpImageDrawing.js
```diff
@@ -388,3 +388,167 @@ evaluator.cameravideo$0 = function() {
...
+
+        //modify pixels for boundary cases:
+        if (repeat) { //read pixels at boundary seperately
+            if (xi === w - 1 || yi === h - 1) {
+                var p10 = readPixelsIndirection(img, (xi + 1) % w, yi, 1, 1);
+                var p01 = readPixelsIndirection(img, xi, (yi + 1) % h, 1, 1);
+                var p11 = readPixelsIndirection(img, (xi + 1) % w, (yi + 1) % h, 1, 1);
+                pixels = pixels.slice(0, 4).concat(p10).concat(p01).concat(p11);
```
gagern: A single call to [`concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) should suffice, as it can take multiple arrays as arguments.

src/js/libcs/OpImageDrawing.js
```diff
@@ -388,3 +388,167 @@ evaluator.cameravideo$0 = function() {
...
+            if (xi === -1 && xf >= 0.5)
+                for (i = 0; i < 4; i++)
+                    for (j = 0; j < 2; j++) pixels[8 * j + i] = pixels[8 * j + i + 4];
+            if (xi === w - 1 && xf < 0.5)
+                for (i = 0; i < 4; i++)
+                    for (j = 0; j < 2; j++) pixels[8 * j + i + 4] = pixels[8 * j + i];
+            if (yi === -1 && yf >= 0.5)
+                for (i = 0; i < 8; i++) pixels[i] = pixels[i + 8];
+            if (yi === h - 1 && yf < 0.5)
+                for (i = 0; i < 8; i++) pixels[i + 8] = pixels[i];
```
gagern: :bomb: This is weird: if you are still inside the image, you clamp to the last pixel, but if you are outside, you interpolate again. Why? Shouldn't a position outside return `[0,0,0,0]` no matter what? How about this:

```js
if (xi === -1 || xi === w - 1) xf = Math.round(xf);
if (yi === -1 || yi === h - 1) yf = Math.round(yf);
```

That will facilitate a round to nearest interpolation approach, and greatly reduce the code complexity here.

src/js/libcs/OpImageDrawing.js
```diff
@@ -388,3 +388,167 @@ evaluator.cameravideo$0 = function() {
...
+
+    if (w0 === nada || w1 === nada || p === nada) return nada;
+
+    //create an orientation-reversing similarity transformation that maps w0->v0, w1->v1
+    var ii = List.ii;
+    var jj = List.jj;
+
+    var m1 = eval_helper.basismap(v0, v1, ii, jj); //interchange I and J,
+    var m2 = eval_helper.basismap(w0, w1, jj, ii); //see Thm. 18.4 of Perspectives on Projective Geometry
+    var p = evaluateAndHomog(args[3]);
```
gagern: I wonder whether it would be a good idea to split the function at this point. Up till here you essentially computed the pixel position. Now you retrieve the actual data. If we at some point decide to support interpolation for the `imagergb$3` call, or to decide `imagergb$5` and so on for more complicated image mapping operations, then having specific position combinations and then sharing the interpolation would be very useful. Modifier handling would go into the second part. What do you think? We can keep this for a later change if you prefer.

ref/Image_Manipulation_and_Rendering.md
```diff
@@ -314,6 +312,35 @@ It plots a point with the corresponding color and opacity and by this creates a
 
 ------
 
+#### Getting pixel data: `imagergb(‹imagename›,‹int›,‹int›)`
+
+**Description:**
+This function does the same as `imagergba(‹imagename›,‹int›,‹int›)`. In particular, it also delivers the alpha value in order to preserve backward compatibility.
+
```
gagern: I wonder whether it would make sense to have a modifier to enable more modern behavior instead of this compatibility burden. Probably not, though, as `_(1..3)` is shorter than any modifier `, modern->true`.

examples/cindygl/34_cpuimagergba.html
```diff
@@ -0,0 +1,43 @@
...
+         var gslp=[{name:"A", kind:"P", type:"Free", pos:[-8,-8,1]},
+                   {name:"B", kind:"P", type:"Free", pos:[8,-8,1]},
+                   {name:"C", kind:"P", type:"Free", pos:[0,3,1],size:10}];
+
+         cdy = CindyJS({canvasname:"CSCanvas",
+                     scripts: 'cs*',
+                     geometry:gslp,
+										 autoplay: true,
+										 use: ["CindyGL"]
+
```
gagern: Minor nit: the indentation here looks weird.

src/js/libcs/OpImageDrawing.js
```diff
@@ -388,3 +388,167 @@ evaluator.cameravideo$0 = function() {
...
+/**
+ * imagergba(‹image›,x,y) implements imagergb(‹imagename›,x,y) from Cinderella, i.e.
+ * returns a 4 component vector ranging from (0-255, 0-255, 0-255, 0-1)
+ */
+evaluator.imagergba$3 = function(args, modifs) {
+    var img = imageFromValue(evaluateAndVal(args[0]));
+    var x = evaluateAndVal(args[1]);
+    var y = evaluateAndVal(args[2]);
+
+    if (!img || x.ctype !== 'number' || y.ctype !== 'number') return nada;
```
gagern: :bomb: I think `!img.ready` should be another reason to return `nada`. Both here and in the other result from `imageFromValue`.

ref/Image_Manipulation_and_Rendering.md
montaga: I think, this function only exists because of backward compatibility reasons. Most users should use imagergb$4

src/js/libcs/OpImageDrawing.js
montaga: The intention of not mixing the pixels with [0,0,0,0] is that I tried to model the way as a image is drawn by context.drawImage: No interpolation is done at the outer half pixels at the boundary. This can be verified in `examples/148_ImageRGBA4.html`.
I'll replace the code by yours. Thanks :-)

plugins/cindygl/src/js/General.js
```diff
@@ -292,6 +307,16 @@ function getPixelType() {
...
+function toFloat(pixels) {
+    let res = [];
+    for (let i = 0; i < pixels.length; i++) {
+        if (can_use_texture_float) res.push(pixels[i]);
+        else if (can_use_texture_half_float) res.push(decodeFloat16(pixels[i]));
+        else res.push(pixels[i] / 255);
+    }
+    return res;
+}
+
```
montaga: I choose the push-approach as `samples` might be a `Float32Array` or `Uint16Array`. For typed Arrays, `map` is not implemented for several browsers yet, and would even preserve the type: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/map

src/js/libcs/OpImageDrawing.js
montaga: makes sense

src/js/libcs/OpImageDrawing.js
montaga: sounds reasonable. Done it

montaga:
Thank you for your review. I hope now everything is satisfying.

gagern:
The issue about `!img.ready` is neither handled nor did you comment on it. Apart from that it's fine.

montaga:
Oh, sorry. Forgot about that :( But now it is included in both cases where `imageFromValue` is called within a imagergb[a]

PR #463, opened 2016-09-07
Support load$2 and load$3

gagern:
The unary`load$1` is not feasible in a web environment, since loading has to happen asynchronously.  So we use a callback model: the function doing the loading takes a piece of code to be invoked once loading has succeeded.  A variable will contain the resulting value during execution of that code, with the default of `#` as usual if the variable name is omitted.

I guess once this gets merged, we can tick or remove `load$1` from #219. Perhaps we should bulk-remove all the functions where we know we are not going to support them any time soon. And adjust the manual to state as much.

### DIFF
--- a/examples/152_load.html
+++ b/examples/152_load.html
@@ -1,0 +1,37 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+
+  color = [1, 1, 0];
+  load("http://sws.geonames.org/2921044/about.rdf", // should support CORS
+    if(isundefined(#), color = [1, 0, 0], color = [0, 0.8, 0]));
+
+</script>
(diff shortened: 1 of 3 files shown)


kranich:
>The unary load$1 is not feasible in a web environment, since loading has to happen asynchronously.

Not necessarily, one *can* load files synchronously, blocking execution. This might not be desired, though.

gagern:
Can we? As far as I know, synchroneous load operations were a non-standard feature of some browsers, and are deprecated now, so browsers will likely not support those. The only way I can think of to make this synchroneous is by rewriting all our interpreter to separate the JavaScript call stack from the CindyScript call stack. That's a *huge* effort, for little gain, I'd say.

kranich:
@gagern Method open of XMLHttpRequest has an optional parameter async that can be set to false. I could not find any hint of it being deprecated in the specification.

See https://www.w3.org/TR/XMLHttpRequest/#synchronous-flag

kranich:
@gagern Just found that https://xhr.spec.whatwg.org/#the-open()-method says otherwise. So you are probably right.

strobelm:
Looks fine to me! Examples work well. Merged. Thanks Martin!

PR #467, opened 2016-09-09
Adjust scale of drawimage(‹pos›,‹image›) to match Cinderella

### LINKED ISSUE #466: Image size for drawimage$2
The function `drawimage(‹pos›,‹name›)` behaves differently in Cinderella and CindyJS. In the former, the size of the image is computed with respect to the user coordinate system, so that zooming in to the construction will zoom into the image. The latter renders the image at its natural size, so that zooming (which is not possible dynamically via the current UI anyway) will not change the apparent size. Both of these have benefits and drawbacks. And while compatibility with Cinderella mandates one behavior, compatibility with existing CindyJS content mandates another.

So which of these compatibility expectations do we prefer to break? As @richter-gebert has probably the best overview over who is using what feature of CindyScript, I'd value his input here.

gagern:
While CindyJS sized its images relative to the viewport so far, this modification makes the size relative to the user coordinate system as it is done in Cinderella.

This is a breaking change, since it will very likely break content which made use of the 2-ary `drawimage` function before. Therefore I recommend that we merge this and #443 and then release CindyJS 0.8.0. Unless someone is aware of other breaking changes.

This fixes #466.

### DIFF
--- a/src/js/libcs/OpImageDrawing.js
+++ b/src/js/libcs/OpImageDrawing.js
@@ -159,6 +159,9 @@ evaluator.drawimage$2 = function(args, modifs) {
         var sc = Math.sqrt(xx1 * xx1 + yy1 * yy1) / Math.sqrt(ixx1 * ixx1 + iyy1 * iyy1);
         var ang = -Math.atan2(xx1, yy1) + Math.atan2(ixx1, iyy1);
 
+        var viewScale = csport.drawingstate.matrix.sdet / 72;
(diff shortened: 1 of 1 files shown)


richter-gebert:
Looks ok. Does the current size Agree with the Cinderella Settings (I have not checked this)

gagern:
Yes, the Cinderella size in Euclidean ports is the same. Feel free to merge if you are happy with this.

PR #469, opened 2016-09-13
WIP: Point on Arc

strobelm:
Hi,
i finished the first implementation of PointOnArc. It currently works fine as long the arc is moved but i need some help on implementing the movement of the point itself. 

Somehow i could not manage to do a proper projection to the arc, if the point is moved. I think it's rather a technical problem than a problem with the projection. I'm a bit lost in between getParamFromState, putParamToState, getParamForInput, initialize and updatePosition.  
Could someone help me out how to properly use the tracing machinery in this case?

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -2261,7 +2261,7 @@ geoOps.TrMoebiusArc = {};
 geoOps.TrMoebiusArc.kind = "C";
 geoOps.TrMoebiusArc.signature = ["Mt", "C"];
 geoOps.TrMoebiusArc.signatureConstraints = function(el) {
-    return csgeo.csnames[el.args[0]].isArc;
+    return csgeo.csnames[el.args[1]].isArc;
 };
 geoOps.TrMoebiusArc.updatePosition = function(el) {
     var t = csgeo.csnames[(el.args[0])];
@@ -2564,6 +2564,9 @@ geoOps.ReflectCP.updatePosition = function(el) {
 geoOps.ReflectCArc = {};
 geoOps.ReflectCArc.kind = "C";
 geoOps.ReflectCArc.signature = ["Rc", "C"];
+geoOps.ReflectCArc.signatureConstraints = function(el) {
+    return csgeo.csnames[el.args[1]].isArc;
+};
 geoOps.ReflectCArc.updatePosition = function(el) {
     var t = csgeo.csnames[(el.args[0])];
     var Arc = csgeo.csnames[(el.args[1])];
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -584,6 +584,43 @@ geoOps.OtherPointOnCircle.updatePosition = function(el) {
     el.homog = General.withUsage(pos, "Point");
 };
 
+
+// start and end assumed to be normalized at once!
+geoOps._helper.projectPointToSegmentCR = function(seg, pos, infseg) {
+    var line = seg.homog;
+    var tt = List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero]);
+    var farpoint = List.sub(seg.startpos, seg.endpos);
+    var cr = List.crossratio3(
+        farpoint, seg.startpos, seg.endpos, pos, tt);
+
+    //handle case if we have infinite segment
+    if (infseg) {
+        // 0 <  cr(FP, A, C, D) < 1 -> not in infinite segment 
+        // probably needs tracing 
+        if ((cr.value.real > 0) && (cr.value.real <= 0.5))
+            cr = CSNumber.complex(0, cr.value.imag);
+        if ((cr.value.real < 1) && (cr.value.real > 0.5))
+            cr = CSNumber.complex(1, cr.value.imag);
+    }
+    // normal case
+    else {
+        if (cr.value.real < 0)
+            cr = CSNumber.complex(0, cr.value.imag);
+        if (cr.value.real > 1)
+            cr = CSNumber.complex(1, cr.value.imag);
+    }
+
+    return cr;
+};
+
+geoOps._helper.projectPointToSegment = function(seg, param) {
+    var start = seg.startpos;
+    var end = seg.endpos;
+    var far = List.sub(end, start);
+    var homog = List.add(start, List.scalmult(param, far));
+    return List.normalizeMax(homog);
+};
+
 geoOps.PointOnSegment = {};
 geoOps.PointOnSegment.kind = "P";
 geoOps.PointOnSegment.signature = ["S"];
@@ -595,16 +632,7 @@ geoOps.PointOnSegment.initialize = function(el) {
 };
 geoOps.PointOnSegment.getParamForInput = function(el, pos) {
     var seg = csgeo.csnames[el.args[0]];
-    var line = seg.homog;
-    var tt = List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero]);
-    var farpoint = List.sub(seg.startpos, seg.endpos);
-    var cr = List.crossratio3(
-        farpoint, seg.startpos, seg.endpos, pos, tt);
-    if (cr.value.real < 0)
-        cr = CSNumber.complex(0, cr.value.imag);
-    if (cr.value.real > 1)
-        cr = CSNumber.complex(1, cr.value.imag);
-    return cr;
+    return geoOps._helper.projectPointToSegmentCR(seg, pos);
 };
 geoOps.PointOnSegment.getParamFromState = function(el) {
     return getStateComplexNumber();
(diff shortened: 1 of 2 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+    var cr = List.crossratio3harm(A,C,B,P,List.ii);
+    var m = cr.value[0];
+    var n = cr.value[1];
+    if (!CSNumber._helper.isAlmostZero(m)) {
+        n = CSNumber.div(n, m);
+        m = CSNumber.real(1);
+    } else {
+        m = CSNumber.div(m, n);
+        n = CSNumber.real(1);
+    }
```
gagern: Why not simply `List.normalizeMax(cr)`?

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+ //   Render2D.drawline(diam);
+//    debugger;
+
+    var isec = geoOps._helper.IntersectLC(diam, arc.matrix);
+
+
+    var crA = geoOps._helper.PointOnArcCr(arc, isec[0]);
+    var prodA = CSNumber.mult(crA[0], crA[1]);
+   // console.log("pA",niceprint(prodA));
+    if (crA[0].value.real < 0) prodA = CSNumber.neg(prodA);
```
gagern: I don't get the semantic intention behind this line. Why look at the sign or the real part of the numerator?

I *think* you want to work out whether `isec[0]` is between start and end, right? Which means whether it has a positive cross ratio. Which for real inputs you can check by looking at the sign of the product of numerator and denominator. That is no longer true for complex inputs: if you have the fraction `i/i` then its sign is positive but the product is negative. You have to multiply the numerator with the conjugate of the denominator.

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+
+    
+    var erg;
+    if(prodA.value.real > 0 && prodB.value.real > 0){
+        console.log("hit1");
+        var d1 = List.projectiveDistMinScal(P, isec[0]);
+        var d2 = List.projectiveDistMinScal(P, isec[1]);
+
+        erg = d1 < d2 ? isec[0] : isec[1];
+    }
```
gagern: This part is similar to [`PointOnCircle.initialize`](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/libgeo/GeoOps.js#L469-L490). Not sure it would make sense to somehow share this, though.

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+        erg = isec[1];
+    }
+    else //(prodB.value.real < 0)
+    {
+        console.log("hit4");
+        erg = isec[0];
+    }
+
+
+    return List.normalizeMax(erg);
```
gagern: We do have the cross ratio of the point in question. Perhaps we should return that as well, since that's more suitable as a parameter.

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+geoOps.PointOnArc.kind = "P";
+geoOps.PointOnArc.signature = ["C"];
+geoOps.PointOnArc.isMovable = true;
+geoOps.PointOnArc.initialize = function(el) {
+    var arc = csgeo.csnames[el.args[0]];
+    var p = geoOps._helper.initializePoint(el);
+    p = geoOps._helper.projectPointToArc(arc, p);
+
+    var cr = geoOps._helper.PointOnArcCr(arc, p);
+    cr = CSNumber.mult(cr[0],cr[1]);
```
gagern: Why the product, not the quotient? And why not both quantities independently, since we are using homogeneous coordinates in other cases as well?

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+    putStateComplexNumber(cr);
+};
+geoOps.PointOnArc.getParamForInput = function(el, pos) {
+    var arc = csgeo.csnames[el.args[0]];
+    var cr = geoOps._helper.PointOnArcCr(arc, pos);
```
gagern: I think `pos` should be projected to the circle as well. Oherwise we'll get complex parameter values if the point is moved off the arc.

src/js/libgeo/GeoOps.js
strobelm: Yes that's the intention and you are right i missed that.

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+    return List.normalizeMax(erg);
+};
+
+
+geoOps.PointOnArc = {};
+geoOps.PointOnArc.kind = "P";
+geoOps.PointOnArc.signature = ["C"];
+geoOps.PointOnArc.isMovable = true;
+geoOps.PointOnArc.initialize = function(el) {
+    var arc = csgeo.csnames[el.args[0]];
```
gagern: We need a [`signatureConstraint`](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/libgeo/GeoBasics.js#L232) to ensure this is indeed an arc.

src/js/libgeo/GeoOps.js
gagern: It is slightly confusing to have `A` and `B` denote the two points of intersection here, since in al the other code `A`, `B` and `C` form the projective basis. Perhaps name these points here `P` and `Q`?

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+
+    var cr = getStateComplexNumber();
+    putStateComplexNumber(cr); // copy parameter
+
+    var erg;
+    if( isMover){
+        var point = geoOps._helper.initializePoint(el);
+        console.log(niceprint(List.normalizeZ(point)));
+        erg = geoOps._helper.projectPointToArc(arc,point);
+        // projection should be done here
```
gagern: What is the motivation for the case distinction here? I'd have thought that we could always compute the result based on the arc and the cross ratio.

src/js/libgeo/GeoOps.js
```diff
@@ -625,6 +625,144 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+
+    var ll = List.sub(aciB,dabiC);
+        ll = List.normalizeMax(ll);
+        ll = List.cross(ll,List.ii);
+
+    erg = geoOps._helper.IntersectLC(ll, arc.matrix);
+    var d1 = List.projectiveDistMinScal(erg[0], II);
+    var d2 = List.projectiveDistMinScal(erg[1], II);
+
+    erg = d1 < d2 ? erg[1] : erg[0];
```
gagern: I don't really understand `aciB`, `dabiC` and `ll`, and I fail to understand the comment at first glance. But I dislike having to do a generic line-conic intersection here.

This is how I see it: point `II` sees `A`, `B`, `C` and our desired point `P` under a given cross ratio. So compute lines joining `II` to `A`, `B` and `C`. Scale them such that two representatives form a projective basis. Express the line from `II` to `P` with respect to this basis. So now you have a line and one point on that line (namely `II`), so you want the other point. Which is a *linear*  problem! Ah, I see you compare results against `II`, so I guess you do something similar there but without making use of the one known solution. I see that [`OtherIntersection`](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/libgeo/GeoOps.js#L1612-L1620) does something similar as well. We really should implement a nice function to make use of the essentially linear nature of this problem.

gagern:
Hi Michael, I had a detailed look. I guess the main problem I have with this is that you use the product as parameter value instead of the quotient or the homogeneous coordinates. Is there an explanation for this, or is this wrong?

src/js/libgeo/GeoOps.js
strobelm: Normalizemax would not be equivalent, this would break of half of the arc.

src/js/libgeo/GeoOps.js
strobelm: We don't have a signature for arcs, right? Afaik we have a flag which indicates if a circle is an arc. Should we add this?

gagern: A `signatureConstraint` is a function which can perform arbitrary checks. We currently [use this for `OtherPointOnCircle`](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/libgeo/GeoOps.js#L577-L579) which only makes sense for circles. But we could add a check for the `isArc` flag. Come to think of it, we should add that same check to [`TransformArc`](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/libgeo/GeoOps.js#L2466-L2475) which has the same kind of problem.

Personally I don't like this whole “types are strings” business here, since it doesn't capture inheritance. Every segment is a line. Every circle is a conic. Every arc is a circle. At least in our world. But look at all the places where we have to either make sure to treal lines and segments alike, or check whether a given conic is a circle and recover if it is not. But I think if we are going to change this, it will be in a later commit and on a larger scale, so for now signature constraints are the way to go, I think.

strobelm:
Hi, here is the rewrite. Looks like it's working well in practice (but for sure needs some testing). I'm also handling the case if the arc is an infinite segment. Cinderella neglects that and is also unstable for segments in general. This is probably due to the projection which only works for non-degenerate circles.

I haven't found a linear solution like @gagern suggested but i'm going to look into that tomorrow. This could be merged (i know that some people in Munich are waiting for that for quite long). Or we could postpone the review until i found a linear solution.

src/js/libgeo/GeoOps.js
gagern: I still don't understand, please explain. `crossratio3harm` returns a homogeneous result, namely a pair [*m* : *n*] which represents the cross ratio *m*/*n*, but includes the special case of infinity if *n* is zero. So *any* multiple of this would represent the same point on the circle. And `normalizeMax` does choose one multiple, in such a way that one of the coordinates becomes 1, just as it is in your code. How can that break anything?

strobelm: Yes that's true and i'm not quite sure why this happens. I'm using m and n later on separately (to avoid the infinity case). I'll look into that.

gagern:
Thanks, @strobelm! I'll have another look. Although I must confess that I would have preferred an incremental commit, so I'd only have to look at your recent modifications.

I don't see the lack of a linear solution as a showstopper, but would file a ticket if it's not included by the time we merge this.

src/js/libgeo/GeoOps.js
```diff
@@ -584,6 +584,43 @@ geoOps.OtherPointOnCircle.updatePosition = function(el) {
...
+geoOps._helper.projectPointToSegmentCR = function(seg, pos, infseg) {
+    var line = seg.homog;
+    var tt = List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero]);
+    var farpoint = List.sub(seg.startpos, seg.endpos);
+    var cr = List.crossratio3(
+        farpoint, seg.startpos, seg.endpos, pos, tt);
+
+    //handle case if we have infinite segment
+    if (infseg) {
+        // 0 <  cr(FP, A, C, D) < 1 -> not in infinite segment
```
gagern: I think it's hard to read such a comment if you introduce new labels for the points which are not used anywhere else. I realize you are using them in `_helper.PointOnArcCr` and perhaps others, but that information is out of scope here.

src/js/libgeo/GeoOps.js
```diff
@@ -584,6 +584,43 @@ geoOps.OtherPointOnCircle.updatePosition = function(el) {
...
+// start and end assumed to be normalized at once!
+geoOps._helper.projectPointToSegmentCR = function(seg, pos, infseg) {
+    var line = seg.homog;
+    var tt = List.turnIntoCSList([line.value[0], line.value[1], CSNumber.zero]);
+    var farpoint = List.sub(seg.startpos, seg.endpos);
+    var cr = List.crossratio3(
+        farpoint, seg.startpos, seg.endpos, pos, tt);
+
+    //handle case if we have infinite segment
+    if (infseg) {
```
gagern: What exactly is an “infinite segment”? If it is a projectively transformed segment which passes through infinity, then the fact that we compute `farpoint` as the projectively transformed point at infinity should already take care of this. I think `examples/62_ProjTransform.html` demonstrates this: it has a yellow point on a projected segment, and does the right thing even if said segment should pass through infinity. If you disagree, do you have an example which demonstrates why this would be needed?

src/js/libgeo/GeoOps.js
```diff
@@ -616,15 +644,170 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+
+    var homog = geoOps._helper.projectPointToSegment(seg, param);
     el.homog = General.withUsage(homog, "Point");
 };
 geoOps.PointOnSegment.stateSize = 2;
 
+geoOps._helper.PointOnArcCr = function(arc, P) {
+    var A = arc.startPoint,
+        B = arc.viaPoint,
+        C = arc.endPoint;
```
gagern: Minor nit: I prefer repeating the `var` keyword. Do we want to achieve consistent styling on this?

src/js/libgeo/GeoOps.js
```diff
@@ -616,15 +644,170 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+
+
+geoOps.PointOnArc = {};
+geoOps.PointOnArc.kind = "P";
+geoOps.PointOnArc.signature = ["C"];
+geoOps.PointOnArc.isMovable = true;
+geoOps.PointOnArc.initialize = function(el) {
+    var arc = csgeo.csnames[el.args[0]];
+    var p = geoOps._helper.initializePoint(el);
+    //p = geoOps._helper.projectPointToCircle(arc, p);
```
gagern: Can we drop this comment, or does it serve a purpose?

src/js/libgeo/GeoOps.js
```diff
@@ -616,15 +644,170 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+        m = CSNumber.real(1);
+    } else {
+        m = CSNumber.div(m, n);
+        n = CSNumber.real(1);
+    }
+
+
+    var prod = m.value.real * n.value.real;
+
+    if (prod >= 0 && prod <= 1) {
```
gagern: Why this case distinction here based on the *product*? I think this should be 0 ≤ *m*/*n* ≤ 1, which you could write like this if you only care about the real values. If you want to avoid division you have to multiply everything with *n*² like this: 0 ≤ *m*∙*n* ≤ *n*². Could this be the reason why `normalizeMax` doesn't do what you expect?

src/js/libgeo/GeoOps.js
```diff
@@ -616,15 +644,170 @@ geoOps.PointOnSegment.updatePosition = function(el) {
...
+        var d1 = List.projectiveDistMinScal(P, A);
+        var d2 = List.projectiveDistMinScal(P, C);
```
gagern: I wonder whether we want to take the closer point in terms of the `projectiveDistMinScal` metric, or whether we should use the already computed cross ratio for this. The same probably applies to point on segment as well.

src/js/libgeo/GeoOps.js
strobelm: This is intended to handle the case when a PointOnArc passes through infinity while the defining points of the arc do not. I think this is a different case, i'm note sure if there is easier way.
a1
a2

strobelm: Nevermind. The point does not pass through infinity (was on the wrong track here). Nevertheless the characteristics of the the Segment changes in such a way that projection to the interval [0,1] yields the wrong result.

strobelm: The more i think about it there should be a solution similar to PointOnSegment and without a case distinction.

strobelm: After further investigations i came to the conclusion that this check is probably needed. When we create a segment we implicitly assume that it's finite and encoding far/midpoint accordingly. We can't assume that for arc simply because we don't know in which position the points occur when the arc degenerates. If you have a different opinion @gagern, i'm open to suggestions.

gagern: Suppose you have an arc from *A* through *B* to *C*. Then the point is uniquely identified by its cross ratio with respect to these, as seen from any other point on the circle like e.g. *I*. This means that CR(*A*,*C*,*B*,*P*) should be between 0 and ∞. If it is negative, then we should map to one of the arc endpoints. Perhaps depending on whether the cross ratio is greater or less than −1. Because if *B* plays the role of the midpoint, then its harmonic conjugate (at cross ratio −1) would play the role of the farpoint, and thus tell you which endpoint to choose. All of this should work equally well for the degenerate and the non-degenerate case, I think. Or am I missing something?

I'm slightly concerned by the fact that we have a parameter range [0,∞] here but [0,1] for PointOnLine. So this requires a homogeneous parameter vector, while that one can be encoded as a single scalar parameter. Should we make an effort to make these two more consistent somehow?

strobelm:
Hi,

how do we proceed here? The problem in a nutshell is that if one changes the ordering of the arc points the midpoint of the arc passes through infinity which is not tracked by the way we are handling projective transformations on segments (since arc points do not need to pass through infinity in order to change their order). So somewhere we have to watch out for this case and i don't see a way to substantially improve the code to handle this. I would be very happy to get some input here (or we merge the code as it is).

gagern:
I'm currently thinking about this. Already have one more commit in my local copy of this, want to play through some ideas, will get back to you. (Got side-tracked on another interesting issue, more on that wehn I'm ready.)

strobelm:
Superseded by #527.

PR #470, opened 2016-09-14
Editable Text

gagern:
This adds EditableText as a gslp type to include text fields in the construction.

### DIFF
--- a/examples/EditableText.html
+++ b/examples/EditableText.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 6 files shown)


gagern:
To enable code reuse I had to merge #461 so this is now a superset of that. Better review that one first.

gagern:
I regenerated the diff here so it omits the calculations part. As far as I can see this PR here contains the last parts required to finally close #14.

strobelm:
I had a look on the rebased version. Works well. Merged.

PR #477, opened 2016-09-16
Fix image sizes for fish swarm examples

### LINKED ISSUE #472: Fix image scale in examples
#467 broke `examples/40_Swarm.html` and `examples/40_SwarmExplorer.html`, causing the fishes to become way smaller than intended.

gagern:
This fixes #472.


kranich:
@gagern Thanks for the fix.

PR #478, opened 2016-09-16
Update Closure Compiler to latest version (v20160911)

kranich:
This brings several improvements to Closure Compiler.

I have tested the new version with several examples of CindyJS core, Cindy3D, and CindyGL, all of which worked.

### DIFF
--- a/make/Settings.js
+++ b/make/Settings.js
@@ -18,7 +18,7 @@ module.exports = function Settings() {
         closure_urlbase: "http://dl.google.com/closure-compiler",
         closure_language: "ECMASCRIPT5_STRICT",
         closure_level: "SIMPLE",
-        closure_version: "20160822",
+        closure_version: "20160911",
         verbose: "true",
         logprefix: "true",
         c3d_closure_level: "ADVANCED",


gagern:
Looking at [the release notes](https://github.com/google/closure-compiler/wiki/Releases#september-11-2016-v20160911), most of that doesn't apply to our project.

* “compile time performance improvements” in my opinion means it will compile more quickly, but since we usually only concatenate files during development, that is of little concern.
* “Allow applying input source maps to the generated source map” might make [one step of our build process](https://github.com/CindyJS/CindyJS/blob/master/make/build.js#L89-L92) superfluous. But currently that step works, so I see no urgent need for changing that. If you want to make use of this, feel free to change the build file and make sure the source maps work as they do now.
* The other changes are even less relevant to us.

So I'd say we can skip this update. Thanks Stefan for keeping an eye out on releases there, but personally I don't think we should follow the latest version unless doing so has some actual benefit for us. I think I'd leave this open till monday, see if the rest of the team agrees with this view.

kranich:
> “compile time performance improvements” in my opinion means it will compile more quickly, but since we usually only concatenate files during development, that is of little concern.

I think it would not hurt if Travis builds run more quickly, for example.

> personally I don't think we should follow the latest version unless doing so has some actual benefit for us. I think I'd leave this open till monday, see if the rest of the team agrees with this view.

+1

kranich:
@gagern And we should take into account that many bug fixes do not make it into the release notes.

gagern:
We [decided today](https://github.com/CindyJS/CindyJS/wiki/Meeting-2016-09-19#closure-versionen) to not follow Closure's every update, but only take those with considerable impact for us, or if the last update has been rather long ago (approx. half a year). So we skip this.

PR #483, opened 2016-09-20
Fullsize and fullscreen examples

gagern:
This improves the fullsize example for mobile devices, by specifying the height as `100%` instead of `100vh`. The former is the minimal viewport size, i.e. what you see if the top bar with the browser controls is displayed. The latter is the maximal viewport size, i.e. with the top bar hidden the way it is after scrolling. Since we don't scroll, the former is the right thing to use here.

See https://github.com/bokand/URLBarSizing for a nice summary of current behavior of various browsers.

This also demonstrates how a widget can be run in full screen mode. This may not work on all browsers, mobile browsers in particular, but it has been tested successfully on common desktop browsers.

### DIFF
--- a/examples/fullscreen.html
+++ b/examples/fullscreen.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Widget which can enter fullscreen mode</title>
+<meta charset="UTF-8">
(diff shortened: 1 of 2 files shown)


strobelm:
Works well for me, merged.

PR #485, opened 2016-09-21
Add the ability to read Text coordinates

### LINKED ISSUE #484: Text coordinates should be accessible and writable
I just tried to convert a file of Iritani-san together with @masatakakaneko and it seems to fail because the xy coordinate of a text is not accessible (at least it is not readable, and probably not writable as well). I attach a simple test case in .cdy format.

The script attached to the button is `Text0.xy = Text0.xy+(1,1);` and this gets exported.

[MoveText.cdy.zip](https://github.com/CindyJS/CindyJS/files/484068/MoveText.cdy.zip)

kortenkamp:
This commit will add the ability to read `xy`, `homog`, `x` and `y` coordinates from Text objects.
The code is actually the same than it is for reading point coordinates, so this might be refactored
to include the code only once. 

there is some superfluous if statement at line 287, please ignore this, I will use it soon in another commit (or remove it).

This will fix #484.

### DIFF
--- a/examples/153_MoveText.html
+++ b/examples/153_MoveText.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Construction.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <link rel="stylesheet" href="http://cindyjs.org/dist/v0.7/CindyJS.css">
+    <script type="text/javascript" src="http://cindyjs.org/dist/v0.7/Cindy.js"></script>
+    <script type="text/javascript">
+createCindy({
+  scripts: "cs*",
+  defaultAppearance: {
+    dimDependent: 0.7,
+    fontFamily: "sans-serif",
+    lineSize: 1,
+    pointSize: 5.0,
+    textsize: 12.0
+  },
+  angleUnit: "°",
+  geometry: [
+    {name: "Text0", type: "Text", pos: [3.48, 4.08, 1.0], color: [0.0, 0.0, 0.0], fillcolor: [1.0, 0.29, 0.29], fillalpha: 0.5384615384615384, text: "Move Me!"},
+    {name: "Text1", type: "Button", pos: [3.8496240601503757, -4.0, -0.7518796992481203], color: [0.0, 0.0, 0.0], fillcolor: [1.0, 1.0, 1.0], fillalpha: 0.27272728085517883, script: "Text0.xy = Text0.xy + [1,1];", text: "Move diagonal"},
+    {name: "Text1'", type: "Button", pos: [2.9655172413793105, -4.0, -0.5747126436781609], color: [0.0, 0.0, 0.0], fillcolor: [1.0, 1.0, 1.0], fillalpha: 0.27272728085517883, script: "Text0.xy = [1,1];", text: "Move to [1,1]"},
+    {name: "Text2", type: "Calculation", pos: [4.0, -2.784, -0.8], color: [0.0, 0.0, 0.0], text: "Text0.xy"}
+  ],
+  ports: [{
+    id: "CSCanvas",
+    width: 680,
+    height: 285,
+    transform: [{visibleRect: [-9.06, 9.34, 18.14, -2.06]}],
+    axes: true,
+    grid: 1.0,
+    background: "rgb(168,176,192)"
+  }],
+  cinderella: {build: 1872, version: [2, 9, 1872]}
+});
+    </script>
+</head>
+<body>
+    <div id="CSCanvas"></div>
+</body>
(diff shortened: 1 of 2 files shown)


kortenkamp:
`node make alltests` fails:

```
[beautified  ] git diff --exit-code --name-only
[beautified  ] examples/153_MoveText.html
[beautified  ] src/js/libcs/Accessors.js
git diff --exit-code --name-only exited with code 1
```

But I have no idea why, and the fail message is not helpful. Can anybody help me?

kortenkamp:
Another comment: This adds also accessors for `.homog`, `.x` and `.y` – these are not available in Cinderella.

kranich:
@kortenkamp `node make alltests` failed since you did not adhere to the coding style. Simply run `node make beautify` for code beautification (should have already been invoked during the failing `node make alltests`). So now you need to commit and push the beautification changes.

kortenkamp:
This is exactly what I thought – but I *did* run `node make beautify`.  I says
```
[beautify    ] beautified src/js/libcs/Accessors.js - unchanged
```
and I did commit all the changes
```
PacMook-Bro:CindyJS ulli$ git status
# On branch accessTextXY
nothing to commit, working directory clean
PacMook-Bro:CindyJS ulli$ 
```

Hm. Running `node make alltests` again I get a "Forbidden Pattern Detected" in my example now. Will check this.

kortenkamp:
Ok. I fixed the example and got again the strange error message, but on the second run it went away. Nothing to commit, though. 
```
PacMook-Bro:CindyJS ulli$ git status
# On branch accessTextXY
nothing to commit, working directory clean
PacMook-Bro:CindyJS ulli$ 
```

So let's cross fingers and hope that Travis won't complain again.

kranich:
Otherwise looks fine. I will merge once the Travis build completed successfully. (We have agreed to assign not before we think the pull request is ready to merge, so I suppose you are ready?)

kortenkamp:
Yes, it's ready (I wouldn't assign otherwise)

kranich:
@kortenkamp Thanks for the implementation.

kranich:
@kortenkamp And congratulations for your first (meaningful) pull request! Great you are finding time to contribute!

gagern:
> `node make alltests` fails:
>
>    [beautified  ] git diff --exit-code --name-only
>    [beautified  ] examples/153_MoveText.html
>    [beautified  ] src/js/libcs/Accessors.js
>    git diff --exit-code --name-only exited with code 1

`node make beautify` modifies the code in place. [`node make beautified`](https://github.com/CindyJS/CindyJS/blob/v0.8.0/make/build.js#L200-L204) checks whether `beautify` would do any changes. It does so by verifying that there are no unstaged changes prior to running that, and then ensuring that there were no new changes after the run. The first is what's failed for you: you had unstaged changes. A `git add -u` should usually solve that. The README [suggests as much](https://github.com/CindyJS/CindyJS/tree/v0.8.0#contributing).

kortenkamp:
thanks for the explanation – but I was completely confused by the error message `git diff --exit-code --name-only exited with code 1`. Does it make sense to file an enhancement request to issue a better error message?

gagern:
I don't want to spoil your enthusiasm, but how do these changes interact with docked texts? And what would we expect from Cinderella? I fear this might need an example and some more implementation.

gagern:
A more helpful error message would be a good thing, and having an issue for it would help me keep track of that. Feel free to write an issue on this and assign it to me.

kortenkamp:
@gagern: yes, this is far from complete for all cases, but it is better to move around docked texts in a wrong way than to be unable to move texts at all. I did not find the code for docked texts so far, but I assume that such a docked text will just ignore direct changes to `geo.homog`. Please either point out where I can find the code to understand it better and fix it or, if you like, please fix it yourself. In any case, it won't harm to file a bug report if you can find wrong behavior with docked texts.

strobelm:
Usually this is caused by the usage of CreateCindy instead of CindyJS in the example file. 

Am 21. September 2016 07:58:42 MESZ, schrieb kortenkamp <notifications@github.com>:

kortenkamp:
@strobelm yes, figured that. I just changed Cinderella to use `var cindy = CindyJS` in its CindyJS template.

gagern:
The docking code is [here](https://github.com/CindyJS/CindyJS/blob/v0.8.0/src/js/libgeo/GeoRender.js#L254-L262). And the main problem with the current code here is that it will do things with `geo.homog` which may be undefined in case of docked text. So those operations would throw exceptions, and those exceptions will prevent subsequent script code from getting executed.

I guess that we'll want to refactor the “give me the position of that text element” code to a separate function, and then call that function both when rendering text and when reading the text position. When writing the text position we'll likely want to remove the docking info, I guess. But I haven't checked yet what Cinderella does for both these cases.

kortenkamp:
Ok, I checked the code and for setting the homogeneous coordinates I do not see a problem – and I think it is fine to just ignore setting that coordinate if the text is docked.

For reading – well, what you propose makes sense. I just checked Cinderella, and it returns the coordinates of the element that the Text has been docked to (disregarding the offset). 

Actually, the access to Text elements' positions was implemented in Cinderella by somebody else, and I had the same concerns you have now in CindyJS, but finally we ended up with the half-baken implementation that is at least correct for non-docked elements (and nobody wants to move docked elements using accessors anyway, as the semantics are not clear at all). So, maybe I just file an issue and we'll see.

PR #496, opened 2016-09-27
Perform division without inversion

### LINKED ISSUE #494: Integer division
I noticed today that in current CindyJS, `107 / 107 != 1`. That's because we compute this internally as `107 * (1 / 107)` and round the intermediate results. However it is probably very confusing to users if simple integer computations like this won't produce exact results. Cinderella doesn't have this problem because it computes division in a single step, without inversion.

gagern:
This avoids rounding errors, particularly when dividing integers.

Fixes #494.

### DIFF
--- a/ref/Texts_and_Tables.md
+++ b/ref/Texts_and_Tables.md
@@ -59,7 +59,7 @@ The code
     D fillStyle = "rgba(127,0,127,1)"
     D font = "bold 25px sans-serif"
     D fillText("Text", 375.5, 229.5)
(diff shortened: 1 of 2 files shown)


kortenkamp:
Just out of curiosity - is this provably more stable than the previous implementation? How did you come across this problem?

strobelm:
Looks good. Thanks!

kortenkamp:
As discussed on the mailing list, I would appreciate if this made its way into an official release soon!

PR #497, opened 2016-09-28
Fine-tune Cinderella version check for animation speed

gagern:
The Cinderella export change to include accuracy as part of the speed value was first included in build 1875, so files created from that or later versions should not halve their speed even if the accuracy got manually removed after exportin the file.


kortenkamp:
Makes sense!

PR #498, opened 2016-09-28
Support keyup and keytyped events

gagern:
Things are slightly complicated by the fact that Cinderella uses strange event names for reasons of backwards compatibility: it maps “Key Down” in the GUI to “keytyped” in the export, and “Key Typed” in the GUI to “keytype” in the export. I've submitted a merge request to change that, but I've also included a backwards-compatibility hack to detect and adjust previous exports.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -146,20 +146,38 @@ function setuplisteners(canvas, data) {
 
     if (data.keylistener === true) {
         addAutoCleaningEventListener(document, "keydown", function(e) {
-            cs_keypressed(e);
+            cs_keydown(e);
             return false;
         });
-    } else if (cscompiled.keydown) {
+        addAutoCleaningEventListener(document, "keyup", function(e) {
+            cs_keyup(e);
+            return false;
+        });
+        addAutoCleaningEventListener(document, "keypress", function(e) {
+            cs_keytyped(e);
+            return false;
+        });
+    } else if (cscompiled.keydown || cscompiled.keyup || cscompiled.keytyped) {
         canvas.setAttribute("tabindex", "0");
         addAutoCleaningEventListener(canvas, "mousedown", function() {
             canvas.focus();
         });
         addAutoCleaningEventListener(canvas, "keydown", function(e) {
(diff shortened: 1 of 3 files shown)


src/js/Events.js
```diff
@@ -146,21 +146,39 @@ function setuplisteners(canvas, data) {
 
     if (data.keylistener === true) {
         addAutoCleaningEventListener(document, "keydown", function(e) {
-            cs_keypressed(e);
+            cs_keydown(e);
             return false;
```
strobelm: I'm wondering if there is any purpose of these return statements.

src/js/Events.js
```diff
@@ -491,16 +509,28 @@ function updateCindy() {
...
     var unicode = evtobj.charCode ? evtobj.charCode : evtobj.keyCode;
     var actualkey = String.fromCharCode(unicode);
     cskey = actualkey;
     cskeycode = unicode;
-    evaluate(cscompiled.keydown);
+    evaluate(script);
     scheduleUpdate();
 }
 
+function cs_keydown(e) {
```
strobelm: These functions look a bit redundant.

strobelm:
Only minor remarks here.

src/js/Events.js
gagern: According to [this Stack Overflow answer](http://stackoverflow.com/a/128966/1468366) the effect should be comparable to `preventDefault`. Perhaps we should be using that throughout? Or perhaps we should re-visit whether we want to prevent default behavior at all? I'm far from certain in this regard, particularly since I had to enable default behavior for keydown in order to receive keypressed. And I left default behavior in place for buttons and text boxes, for obvious reasons at least in the latter case since default behavior is typing text.

src/js/Events.js
gagern: Yes, they do to me as well. But they are consistent with existing code patterns, so if we are going to drop these, we should remove all the others as well.

They might help future improvements, though. Since all invocations of event scripts go through these functions, it's easy to add some code there. Adding debug output to one of these can be useful as well during development. So I don't have a strong opinion either way.

strobelm:
No showstoppers here, merged.

PR #499, opened 2016-09-28
Allow currenttext as an alias for the text accessor

gagern:
Apparently Cinderella supports both `.text` (inherited from the basic text class) and `.currenttext` (specific to the `EditableText` type), so we should support both as well.

### DIFF
--- a/src/js/libcs/Accessors.js
+++ b/src/js/libcs/Accessors.js
@@ -120,7 +120,7 @@ Accessor.getField = function(geo, field) {
                 return General.bool(false);
             }
         }
-        if (field === "text") {
+        if (field === "text" || field === "currenttext") {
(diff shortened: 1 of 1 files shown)


kortenkamp:
In Cinderella, `currenttext` has different semantics than `text`. With `currenttext` you can retrieve the current text in the EditableText, even if it has not been confirmed with the enter key. 

Nevertheless, I merge this as it is better to return `text` than to fail if `currenttext` is requested. Actually, I do not know whether it is possible at all to get the currenttext of an EditableField.

gagern:
Thanks for the clarification. I've just created #500 to keep track of this distinction. I think we can go with the current behavior for now. Feel free to merge.

PR #501, opened 2016-09-29
Support printing messages with stack traces

gagern:
I've started this as a debugging aid, since I got swamped by this error message with no clue where it came from. Now the message to the web console will contain a list of all the functions (both built-in and user-defined) called up to that point. I wouldn't want to merge this as is, but instead I'd like to start a discussion.
1. Do we want this included in the main codebase, or just merge it in when debugging something? Since others should be able to use our code, I'd say the former, even if it might have a slight performance impact.
2. Do we want to enable this by default, or only upon request (like e.g. some specific query parameter)? I see no harm in having this enabled in general.
3. Do we want to go through `csconsole`, or directly to `console.err`? The latter tends to causes these messages to appear in red, but I'm not even sure that's a benefit.
4. Do we want to add this to other messages? If so, which ones?
5. Do we want to add top-level labels for button scripts? `evokeCS`? `parse$1`? Anything else?
6. What order should stack frames be printed in? At the moment it's from outside to inside, i.e. the first line describes the event trigger and the last line is the function where the problem occurred.
7. Can anyone think of places where the synthetic top level infix operations I use for labeling might break something? I don't think CindyGL should be seeing those since it will only ever see a portion of a script, never a whole script. Anything else?

At some point I'd like to include the actual code of the function invocation for each stack frame.


kortenkamp:
> Do we want this included in the main codebase, or just merge it in when debugging something? Since others should be able to use our code, I'd say the former, even if it might have a slight performance impact.

I would prefer not to have any performance impacts. Did you try this with highly recursive code, for example? If it is possible to not include this but only on demand, this would be great.

> Do we want to enable this by default, or only upon request (like e.g. some specific query parameter)? I see no harm in having this enabled in general.

see above – I would like to have this by default, but a way to disable it in case it hits performance (don't forget that many mobile phones are still slow).

> Do we want to go through csconsole, or directly to console.err? The latter tends to causes these messages to appear in red, but I'm not even sure that's a benefit.

csconsole.

> Do we want to add top-level labels for button scripts? evokeCS? parse$1? Anything else?

yes

> What order should stack frames be printed in? At the moment it's from outside to inside, i.e. the first line describes the event trigger and the last line is the function where the problem occurred.

I am used to the other order (recent first), but have no objections against changing that.

> Can anyone think of places where the synthetic top level infix operations I use for labeling might break something? I don't think CindyGL should be seeing those since it will only ever see a portion of a script, never a whole script. Anything else?

What about `parse`and possible extensions that would like to manipulate the expression tree? But as this is hypothetical, maybe we shouldn't care.

src/js/libcs/Evaluator.js
```diff
@@ -151,3 +154,29 @@ function analyse(code) {
...
+            var res = evaluate(code);
+            callStack = [];
+            return res;
+        }
+    };
+}
+
+function printStackTrace(msg) {
+    csconsole.err(msg + "\n" + callStack.map(function(frame) {
+        return "  at " + frame.oper;
```
montaga: It would be really nice, if line numbers were also printed. But this would be a quite invasive surgery for the `analyse` function.

montaga: In particular, this could help spot a error that occurs in a list of functions that are concatenated with `;`.

gagern: Line numbers with respect to the script should be doable if we stop [discarding](https://github.com/CindyJS/CindyJS/pull/354) the [position information we have](https://github.com/CindyJS/CindyJS/blob/v0.8.3/src/js/libcs/Parser.js#L308-L314). But as you're usually looking at the HTML file as a whole, those line numbers would have to be offset by the start of the `<script>` tag. I doubt there is an easy [way to obtain that](http://stackoverflow.com/q/16298945/1468366). Perhaps it would be better to print some context around the point of invocation. But as that would require some more work, I'd leave that for a later improvement.

montaga:
Oh sorry, my formulation was not very clear. The output via `csconsole` is pretty good for this purpose.
I meant that `printStackTrace` should be used for more outputs than only non-boolean if conditions.
In the recent commit I have replaced many `console.log`, `console.err`, `csconsole.err` with `printStackTrace`.

montaga:
I cherry-picked the relevant commits in
https://github.com/CindyJS/CindyJS/pull/623

Issue #509, opened 2016-10-10
lines do not support .angle property


kortenkamp:
In Cinderella, a line can be asked for its (Euclidean) angle with respect to the x-axis using the `.angle` property. John Mason is using this in the attached example, which **silently** fails in CindyJS. 
[S3 Perp Bisector Intersections.cdy.zip](https://github.com/CindyJS/CindyJS/files/520320/S3.Perp.Bisector.Intersections.cdy.zip)

gagern:
CindyJS is silent in way too many cases… Nevertheless, there is [code for a `.angle` accessor](https://github.com/CindyJS/CindyJS/blob/v0.8.1/src/js/libcs/Accessors.js#L62-L65). Strange!

gagern:
Looking at the example, that's not the angle of a line, but the angle of a point on circle. That's indeed currently unsupported, should be fixed.

adamjackir:
It seems like you're working with a point on a circle that doesn't support the .angle property, which could be a bit tricky if you're trying to calculate or define angles directly through it. On the other hand, an angle bisector is a line or ray that divides an angle into two equal parts, essentially cutting it down the middle. It’s a key concept in geometry and helps in creating symmetrical structures.

Also, if you're looking for more insights or discussions on geometry, feel free to check out [Global Wiza Hub](https://globalwizahub.com/) – a great place to connect and learn more from others!

PR #512, opened 2016-10-11
Allow setting the fillcolor of buttons

### LINKED ISSUE #510: Textbuttons should support Colors
Buttons are currently gray. In Cindy they have colors.

gagern:
This fixes #510.

I've decided to move the type-specific accessors to GeoOps, so that we don't clutter Accessors with a lot of type-specific functionality. Let me know whether you consider this a good move.

### DIFF
--- a/examples/Buttons2.html
+++ b/examples/Buttons2.html
@@ -2,25 +2,13 @@
 <html>
 <head>
     <meta charset="UTF-8">
-    <title>Buttons2.cdy</title>
-    <style type="text/css">
-        * {
-            margin: 0px;
-            padding: 0px;
-        }
-
-        #CSConsole {
-            background-color: #FAFAFA;
-            border-top: 1px solid #333333;
-            bottom: 0px;
-            height: 200px;
-            overflow-y: scroll;
-            position: fixed;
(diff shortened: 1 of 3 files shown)


strobelm:
I somehow feel uneasy with functions changing the display of an object in geoOps, mixing model and view is usually bad practice. Perhaps i am to pedantic here, any further opinions?

kortenkamp:
I don't think that the color is a property of the view, but of the model, indeed. Quick example: If you change the coordinates of a line, it's view will change as well, but we – hopefully – agree on the fact that the coordinates are a model property. 

The color is a property of text buttons, and how to interpret them is depending on the view, yes. But it belongs into geoOps, I grew with @gagern .

gagern:
When moving the accessors to GeoOps I also had some less appearance-related accessors in mind. Namely #509 requires a read/write `.angle` property for `PointOnCircle` which will likely cloely resemble some of the functionality we have in existing `PointOnCircle` methods. And I guess there will be more accessors like these.

strobelm:
Alright i'm sufficiently convinced. The code looks fine. Merged.

PR #514, opened 2016-10-17
Extended Möbius transformations

gagern:
This changes our framework for Möbius transformations to also contain support for anti-Möbius transformations:

```
x + i*y ↦ ((ar + i*ai)*(x - i*y) + (br + i*bi))/((cr + i*ci)*(x - i*y) + (dr + i*di))
```

Note in particular the minus in `x - i*y` here. An anti-Möbius transformation is a complex conjugation followed by a regular Möbius transformation. Therefore we can get the corresponding 3×3 matrices by negating their second column. The code does this using a function `flip` which is negation in the anti-Möbius case but identity in the regular Möbius case.

One canonical application of anti-Möbius transformations is inversion in a circle. This used to be its own type so far, but now we can employ all the machinery for Möbius transformations, thus obsoleting (and in fact removing) large portions of code from #151.

Another new addition is the inversion of Möbius transformations, which is already geared to also include anti-Möbius transformations as well. Since inversion on the level of 3×3 matrices would be problematic, this pull request introduces another data structure for Möbius transformations which maintains the 8 conceptually real components of the ℂℙ¹ transformation. (Extracting them form the 3×3 matrices is complicated by the fact that both matrices are normalized independently). The code to turn this representation into a pair of matrices is now shared between different operations.


kortenkamp:
I am a bit puzzled. I have not time to check details, but complex conjugation sounds like a non-analytic function introduce different here. Is this safe? Usually, all geometric operations have to be analytic in order to not destroy tracing. 

Please check this!

gagern:
This is a complex conjugation on the “doubly complex” representation, i.e. the form where we perform the steps we usually perform when doing complex arithmetic using real numbers, except we do these steps using complex numbers to represent the components. You can also think of this as 2×2 Hermitian matrices (with complex entries), and “conjugation” just transposes these. (without conjugating the actual components).

kortenkamp:
Ok, fine.

elkins0:
Nice. Intriguing. On first blush I notice that this exact same code fragment:
```JavaScript
List.realVector([Math.random() - 0.5, Math.random() - 0.5, Math.random() - 0.5]);
```
now appears three times in GeoOps.js. Also in `geoOps.RandomPoint` effectively the same code appears:
```JavaScript
    el.homog = List.realVector([100 * Math.random(), 100 * Math.random(), 100 * Math.random()]);
    el.homog = List.normalizeMax(el.homog);
```
Perhaps these could become `List.randomVector(3)`:
```JavaScript
List.randomVector = function(n) {
    var erg = [];
    for (var i = 0; i < n; i++) {
        erg[i] = {
            "ctype": "number",
            "value": {
                'real': Math.random() - 0.5,
                'imag': 0
            }
        };
    }
    return {
        'ctype': 'list',
        'value': erg
    };
};
```

elkins0:
@gagern: In the initial description, I see the word _frameword_ which I assume is supposed to be the word _framework_. Similarly the word _remobving_ is probably meant to be _removing_.

gagern:
@elkins0: Refactoring the creation of random objects into reusable code is part of #299. In its current draft it will introduce `List.getRandComplexVec`. As creation of randomized elements is a core feature there, while the line here only changed indentation, I'd rather track that change there.

elkins0:
In my comment to [#515](https://github.com/CindyJS/CindyJS/issues/515#issuecomment-256112303) I show a way to implement `geoOps.TrMoebiusC.updatePosition` and `geoOps.TrMoebiusL.updatePosition` entirely without the use of randomness which works without quirks as implemented.

examples/154_ReflectC.html
```diff
@@ -0,0 +1,56 @@
...
+  defaultAppearance: {
+    dimDependent: 0.7,
+    fontFamily: "sans-serif",
+    lineSize: 1,
+    pointSize: 5.0,
+    textsize: 12.0
+  },
+  angleUnit: "°",
+  geometry: [
+    {name: "A", type: "Free", pos: [4.0, 3.2, 0.8], color: [1.0, 0.0, 0.0], labeled: true},
```
elkins0: The free point `A` appears nowhere else in the construction. Perhaps it should be put to better use as the center of the circle `C0`. Also a point should be added for the center of the circle `C3` as that is all that remains when `C2's ` radius is reduced to zero. Of course I found that I had to increase  `C2's ` radius from one to two in order to be able to stretch it at all. Also it might be nice to add a bit more color to associate the objects and their reflections in the circle `C0`.

elkins0: The free point `A` appears nowhere else in the construction. Perhaps it should be put to better use as the center of the circle `C0`. Also a point should be added for the center of the circle `C3` as that is all that remains when `C2's ` radius is reduced to zero. Of course I found that I had to increase  `C2's ` radius from one to two in order to be able to stretch it at all. Also it might be nice to add a bit more color to associate the objects and their reflections in the circle `C0`.

gagern:
I'd say #515 should not be included in this PR here, but discussed separately and only merged after this one here, in order to keep possibly controversial stuff out and to keep the scope of this PR here focused. Thanks for the suggestion there, and the cross reference.

examples/154_ReflectC.html
gagern: Valid points, all of them. Will try to address them soon.

src/js/libgeo/GeoOps.js
```diff
@@ -1994,18 +1999,42 @@ geoOps.TrMoebius.updatePosition = function(el) {
...
 
+geoOps.TrInverseMoebius = {};
+geoOps.TrInverseMoebius.kind = "Mt";
+geoOps.TrInverseMoebius.signature = ["Mt"];
+geoOps.TrInverseMoebius.updatePosition = function(el) {
+    var m = csgeo.csnames[el.args[0]].moebius;
+    var neg = CSNumber.neg;
+    var flip = m.anti ? neg : General.identity;
+    el.moebius = {
+        sign: m.sign,
```
elkins0: As far as can be seen, the el.moebius object's `sign` property is not referenced elsewhere. Also el.moebius object's `anti` property is not defined. So perhaps this line should be changed from `sign: m.sign` to `anti: m.anti`.

strobelm:
Looks like this converged, is the current state ready for review?

gagern:
Yes, @strobelm, I'd say this is ready for review. In #517 I see some issues with unexplained sign flipy for the *y* coordinate of Möbius transformations, so there *might* be a bug in here somewhere. I'm currently trying to pinpoint this.

strobelm:
Code looks good, examples too. Merged. Thanks @gagern for the code and @elkins0 for the bug squashing

PR #517, opened 2016-10-17
Compute iterated function systems

gagern:
Building on #514 this is a first shot at IFS done the Cinderella way: continuosly apply transformations to a point and draw the pixel at the resulting position. To keep the browser responsive, the computation is performed in a worker thread. I've done a bit of tight loop optimization here: I'm only dealing with the case of real matrices (since I assume that for complex transformation matrices, the resulting points would mostly be complex, so we shouldn't draw those). The code to do the bit manipulation is written in C++ and compiled to JavaScript (and indeed asm.js for browsers who optimized for that) using emscripten. I've added the transpiled version so we don't depend on emscripten as part of our build process.

Screenshot

There are a couple of things I'd like to discuss and / or investigate before I'm happy with this.
- [x] Currently this only supports Möbius transformations. Projective ones are implemented in the C++ code but not handled in the other part of the worker yet. Easy to add.
- [x] I think I'm creating premultiplied colors, but I haven't yet found a way to tell this to the `createImageBitmap` function or `ImageData` constructor. So perhaps I have to try to generate non-premultiplied color values.
- [x] There is a sign flip in the _y_ coordinate and I don't know where that's coming from. At some point I thought it was a difference in the _y_ orientation convention in GL Textures vs. 2d contexts, but that would swap [0,h] with [h,0], not turn [0,h] into [0,-h]. Strange.
- [ ] There are a number of constants which might need some tweaking. The number of points discaded for warmup, the number of pixels between subsequent images, the suprsampling factor, the color composition alpha value. Should experiment those at some point.
- [ ] Perhaps we should even auto-tune the number of points between frames to optimize for CPU performance. So that on mobile devices we still get reasonable update rates even if there are fewer pixels added between consecutive frames.
- [x] The Cinderella export had to be extended to export the colors and probabilities. And I'm not sure I like the format. Perhaps it would be better to have two arrays `colors` and `probs` instead of _n_ values of the style `ifs.color0` or `ifs.prob1`. The dots in there are a bit ugly, too.
- [x] This doesn't work on Safari yet. At least not the Safari 9.1.3 I have here. It complains about missing `Uint8ClampedArray.fill` which is easy to work around (at some performance hit), but then it complains about lack of a `createImageBitmap` function which is probably a showstopper here.


montaga:
This is a beautiful push request. Both from the result and a technical point of view!
I did not find time to understand how everything is working.
Maybe, it is not to much effort to implement option to have multiple IFS workers running at the same time, (in order to track different orbits, they should be initialized with different seeds) and displaying an overlay of all the generated images. Most modern devices where CindyJS runs will have multiple cores that could be used in parallel.

gagern:
@montaga: The example included in this PR *already* contains two IFS using different probabilities, which are being handled by two concurrent workers. So far this appears to work as intended. *Not* using multiple concurrent workers would in fact be more difficult, because as it is each worker can simply use some global variables to maintain its state.

But I guess you meant running multiple workers for the same IFS, right? My main concern is that prople might not agree with a random web page using up several of their CPU cores, particularly since I doubt we can get the actual number of available cores from within JavaScript. Having three concurrent workers plus the GUI thread on a two-core device would be a serious responsiveness killer. So it's an interesting idea but not uncontroversial I guess.

gagern:
I'm no longer sure it's a good idea to have separate workers for separate IFS. At least in the Necklace example, where they render to the same parts of the canvas, this causes one of them to be in front of the other, while in Cinderella they both get a chance to render to the same area in an interleaved fashion.

I noticed that [Firefox has a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1271504) which causes it to perform alpha premultiplication repeatedly, thus dimming the IFS if it gets drawn more than once. That is a problem while the user moves the mouse, or if there is more than one IFS in the current multi-worker setup, or if there is an animation… I wonder whether we should try to detect this problem and work around it somehow. We can't use the Safari workaround for this, though, since [Firefox can't transfer buffers which are linked into asm.js code](https://bugzilla.mozilla.org/show_bug.cgi?id=1312139). I had a fun evening isolating these bugs…

gagern:
Firefox also appears to have a memory leak somewhere. At least my Firefox on linux is eating up memory *real fast* while running the necklace example. I just reported [yet another Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1312148) about this.

elkins0:
See [Functions and classes available to Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Functions_and_classes_available_to_workers).
Under APIs available in workers, `ImageData` is supported in FireFox since version 25 however there is no support shown for IE, Blink (Chrome and Opera) or WebKit (Safari). I can confirm `ImageData` does not work in the latest IE in Windows 10 when called from `worker.js`:
```
SCRIPT0: Object doesn't support this action
ifs.js (85,1)
```
And for FireFox with NoScript running the worker script is not allowed:
```
SecurityError: Failed to load worker script at "http://localhost:8000/ifs.js" (unknown)
```

gagern:
[The WHATWG living standard](https://html.spec.whatwg.org/multipage/scripting.html#imagedata) marks `ImageData` as exposed in both window and worker contexts, but I guess it's not implemented like this in all browsers yet. Lack of `ImageData` object in the worker is something I should be able to work around, but if it's absent in the worker, it's perhaps absent in the main thread as well. I somehow doubt I'll get all of this working on IE, and would consider requiring Edge to be a valid approach for such widgets.

Not sure what to do about the NoScript scenario. Perhaps I can somehow work out whether that's the likely cause, and if so, ask users to disable NoScript if they want to view this content. I might also be able to fall back to main-thread rendering, but I'd rather avoid that.

elkins0:
OK—the example `156_Necklace4.html` can be made to work in both Edge and Firefox browsers on a Windows 10 system. Copy `156_Necklace4.html` to `build\js` and  change it to the reference `Cindy.js` in the same directory:
```html
24   -    <script type="text/javascript" src="../build/js/Cindy.js"></script>
24   +    <script type="text/javascript" src="Cindy.js"></script>  
```
`Cindy.js` was also changed to reference `ifs.js` with no path:
```javascript
2829 -    var worker = ifs.worker = new Worker(baseDir + "ifs.js");
2829 +    var worker = ifs.worker = new Worker("ifs.js");
```
The resulting generated image matches the one shown in the initial description above.

The latest version of IE in Windows 10 using the same files still reports (only to the console log) :
```
SCRIPT0: Object doesn't support this action
ifs.js (85,1)
```

gagern:
Is the same directory a requirement when serving from a `file:` URI, or also when using the local development HTTP server?

elkins0:
That was for `file:` URI. I just tried it using `http://localhost:8000/156_Necklace4.html` and that also works for Edge and Firefox but not IE. Using only:
```
CD build\js
python.exe -m SimpleHTTPServer 8000
```

gagern:
Note that the [README](https://github.com/CindyJS/CindyJS/blob/master/README.md#examples) already suggests a way to start a simple HTTP server: `npm_modules/.bin/st -l -nc`. And I guess I should add `node make live` to that, as it provides a local server as well. Although the Python command looks reasonably simple, so use whatever you are comfortable with.

gagern:
While there is certainly room for improvements, I think this can be merged now that the data format is as we'd like it to be and the sign error fixed.

gagern:
Note that this branch builds on #580 so reviewing and merging that first might make sense.

montaga:
I am not sure whether this is the best place here, but the Cinderella export has not been adapted accordingly in the recent version. Both https://cdyconvert.cinderella.de/ and my local version (2.9 build 1835) do not export the color and prob arguments for each transformation yet.

montaga:
Within *Chromium* 56 the following example is eating up a lot of my memory:
[SticklerIFS.html.txt](https://github.com/CindyJS/CindyJS/files/871912/SticklerIFS.html.txt)
screenshot from 2017-03-27 11-25-42
Sometimes it might even crash my system. Within the included examples `157_IFS2sim.html` or `156_Necklace4.html` qualitatively similar behavior occurs, but in smaller magnitude. (i.e. they never crashed my system).
If I force to use the Firefox-workaround also for Chromium (by adding a `ff = true;` in `/src/js/ifs/worker.js`) , the memory behavior becomes much better:
screenshot from 2017-03-27 11-34-15

gagern:
The Cinderella export is still not merged into beta, see !160 there.

Chromium eating memory like this is likely an indication of a bug there. I will try to see whether I can reproduce, create a minimal reproducing example and file a bug for this.

gagern:
Filed [Chromium bug \#705718](https://bugs.chromium.org/p/chromium/issues/detail?id=705718) to track this. I guess I should enable the `ff` workaround by default for now, and hope that one day all of these issues get resolved and we can start transferring images instead of byte arrays.

montaga:
Now the bugs are fixed. I have tested the IFS-functionality in recent versions of Safari, Firefox, Chrome and Chromium and the memory consumption does not explode anymore :+1:

PR #518, opened 2016-10-18
Export to PNG

strobelm:
Hi, we need exporttoPNG on Saturday (perhaps it's useful for someone else).

### DIFF
--- a/src/js/libcs/RenderBackends.js
+++ b/src/js/libcs/RenderBackends.js
@@ -868,3 +868,9 @@ globalInstance.exportPDF = function() {
         exportWith(PdfWriterContext, wnd);
     });
 };
+
+globalInstance.exportPNG = function() {
+    var data = csctx.canvas.toDataURL();
+    var wnd = window.open('about:blank', '_blank');
+    wnd.location.href = data;
+};
--- a/src/js/libcs/RenderBackends.js
+++ b/src/js/libcs/RenderBackends.js
@@ -840,7 +840,7 @@ shutdownHooks.push(releaseExportedObject);
 // result in a new tab.  Note that Firefox fails to show images embedded
 // into an SVG.  So in the long run, saving is probably better than opening.
 // Note: See https://github.com/eligrey/FileSaver.js/ for saving Blobs
-function exportWith(Context, wnd) {
+function exportWith(Context) {
     cacheImages(function() {
         var origctx = csctx;
         try {
@@ -851,7 +851,9 @@ function exportWith(Context, wnd) {
             var blob = csctx.toBlob();
             releaseExportedObject();
             exportedCanvasURL = window.URL.createObjectURL(blob);
-            wnd.location.href = exportedCanvasURL;
+
+            downloadHelper(exportedCanvasURL);
+            window.URL.revokeObjectURL(exportedCanvasURL);
         } finally {
             csctx = origctx;
         }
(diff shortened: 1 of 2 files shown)


src/js/libcs/RenderBackends.js
```diff
@@ -868,3 +868,9 @@ globalInstance.exportPDF = function() {
         exportWith(PdfWriterContext, wnd);
     });
 };
+
+globalInstance.exportPNG = function() {
+    var data = csctx.canvas.toDataURL();
+    var wnd = window.open('about:blank', '_blank');
+    wnd.location.href = data;
```
gagern: This is a synchronous operation, so you might as well avoid using the `about:blank` URL first, and instead directly use the URL of the PNG.

src/js/libcs/RenderBackends.js
```diff
@@ -868,3 +868,9 @@ globalInstance.exportPDF = function() {
         exportWith(PdfWriterContext, wnd);
     });
 };
+
+globalInstance.exportPNG = function() {
+    var data = csctx.canvas.toDataURL();
```
gagern: Personally I was wondering how this defines the format to be PNG. I then [read](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL) that PNG is the default format, but I'd rather make that explicit by passing `"image/png"` as an argument to that function. That makes it easier to undertsand what's happening, and also to adapt it to other formats if we ever need those.

All the other exporting functions are using blobs instead of data URLs. I wonder whether there is any reason to prefer one over the other in the situation at hand. A data url is far longer, but also more portable and more permanent. So perhaps fine for this task unless we run into URL length problems in some real world setup.

src/js/libcs/RenderBackends.js
kranich: When I implemented PNG export for ComplexCurves I found it preferable to generate an `<a>` element pointing to the data URL then invoking its `click` event. Unlike opening a new window, this also works with popup blockers.

Moreover, you can specify a file name in the data URL, too.

kranich: Like so:

```JavaScript
function download(name, url) {
    var link = document.createElement("a");
    link.href = url;
    link.download = name;
    link.style.display = "none";
    document.body.appendChild(link);
    link.click();
    document.body.removeChild(link);
}
```

gagern: I haven't seen any problems with popup blockers if the new windows were opened in immediate response to a button event, i.e. from within the event handler. Which is the reason the other export functions create a window with `about:blank` as URL directly from within the handler, even if they only change the location of said window later in in an asynchronous way.

src/js/libcs/RenderBackends.js
```diff
@@ -851,26 +851,36 @@ function exportWith(Context, wnd) {
             var blob = csctx.toBlob();
             releaseExportedObject();
             exportedCanvasURL = window.URL.createObjectURL(blob);
-            wnd.location.href = exportedCanvasURL;
+
+            downloadHelper(exportedCanvasURL);
+            window.URL.revokeObjectURL(exportedCanvasURL);
```
gagern: This is overlapping functionality of `releaseExportedObject`. When I wrote that, I was unsure about when we can revoke an object URL. To err on the side of caution, I decided to only revoke URLs when the next export is underway, so that we have at most one object active at any point, but the most recently exported object stays alive until superseded.

If you are certain that revoking the URL immediately is safe, and perhaps even have some documentation supporting that claim, then you can drop `releaseExportedObject` (both the call and the implementation) and make `exportedCanvasURL` local to this scope here.

src/js/libcs/RenderBackends.js
```diff
@@ -851,26 +851,36 @@ function exportWith(Context, wnd) {
...
+
+
+var downloadHelper = function(data) {
+    var a = document.createElement("a");
+    document.body.appendChild(a);
+    a.style = "display: none";
+    a.href = data;
+    a.download = "CindyJSExport";
+    a.click();
+    document.body.removeChild(a);
```
gagern: Ah, so instead of *showing* the exported content in a *new* tab, we now *download* it in the *current* tab. I missed that aspect (no new tab) when reading the comment by @kranich. Hmm… Probably makes sense for most practical use cases, but will be annoying when developing export functionality, where I want to have a quick look at the generated file without having to download and open it every time. Well, I'll find a way around when I gets to this.

strobelm:
Probably the best solution would be a webworker which removes the objects when downloading is done. I just quickly added a settimeout, which should be fine for our purposes on Saturday. We can discuss a better solution later (so no pressure merging this right now).

gagern:
I think a worker would be complete overkill. A timeout sounds fine. Actually I think that keeping one eported image in RAM would be acceptable as well. We just have to make sure we don't leak arbitrary amounts of resources.

PR #519, opened 2016-10-18
Read webcam in maximal resolution

montaga:
With this the image of the webcam is read in the maximal possible resolution. We require this for saturday.
I am not sure wheather there is a nicer solution, but according to http://stackoverflow.com/questions/27420581/get-maximum-video-resolution-with-getusermedia it seems like there is no better solution.

### DIFF
--- a/src/js/libcs/OpImageDrawing.js
+++ b/src/js/libcs/OpImageDrawing.js
@@ -349,21 +349,34 @@ evaluator.allimages$0 = function() {
 
 evaluator.cameravideo$0 = function() {
     var openVideoStream = null;
-    var constraints = {
+    // As per https://bugs.chromium.org/p/chromium/issues/detail?id=543997#c47,
+    // Chrome 54 doesn't actually honor ideal constraints yet, so we need
+    // to explicitely list some common widths to control resolution selection.
+    var constraints = [320, 640, 1024, 1280, 1920, 2560];
+    constraints = constraints.map(function(w) {
+        return {
+            width: {
+                min: w
+            }
+        };
+    });
+    // We'd like to also minimize aspect ratio i.e. maximize height for a given
+    // width, but Chrome again appears to have a problem with this. See also
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=657145
+    if (false) {
+        constraints = constraints.concat([1.34, 1.59, 1.78].map(function(a) {
+            return {
+                aspectRatio: {
+                    max: a
+                }
+            };
+        }));
+    }
+    constraints = {
         video: {
-            optional: [{
-                minWidth: 320
-            }, {
-                minWidth: 640
-            }, {
-                minWidth: 1024
-            }, {
-                minWidth: 1280
-            }, {
-                minWidth: 1920
-            }, {
-                minWidth: 2560
-            }, ]
+            width: 16000, // ideal dimensions, will
+            height: 9000, // prefer big resolutions
+            advanced: constraints
         },
         audio: false
     };
--- a/src/js/libcs/OpImageDrawing.js
+++ b/src/js/libcs/OpImageDrawing.js
@@ -351,38 +351,36 @@ evaluator.cameravideo$0 = function(args, modifs) {
     var maximal = true; //use maximal as default (if no other modifier is given)
     var constraints = {};
 
+    function makeconstraints(width) {
+        return {
+            video: {
+                width: width,
+                advanced: [{
+                    width: {
+                        max: width, //see below for details
+                        min: width
+                    }
+                }, {
+                    width: {
+                        ideal: width
+                    }
+                }]
+            },
+            audio: false
+        };
+    }
+
     if (modifs.resolution !== undefined) {
-        maximal = false;
         var val = evaluate(modifs.resolution);
         if (val.ctype === 'string' && val.value === 'maximal') maximal = true;
         else {
-            var width = null;
-            var heightorratio = null;
-
             if (val.ctype === 'number') {
-                width = val.value.real;
-            } else if (List.isNumberVector(val).value && (val.value.length === 2)) {
-                width = val.value[0].value.real;
-                heightorratio = val.value[1].value.real;
-            } else return nada;
-
-            constraints = {
-                video: {
-                    width: width,
-                    advanced: [{
-                        width: {
-                            max: width, //ideal or exact seem not be supported by Chrome yet
-                            min: width
-                        }
-                    }, {
-                        width: {
-                            ideal: width
-                        }
-                    }]
-                },
-                audio: false
-            };
-            if (heightorratio) {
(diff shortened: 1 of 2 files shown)


gagern:
Is there any specification for this `optional` key? I haven't found anything on https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia or https://w3c.github.io/mediacapture-main/#idl-def-mediastreamconstraints. So perhaps we should at least allow for some fallback in case some browser does not understand this notation.

Looking at both these resources, it should be possible to do something like this, though:

```js
var constraints = {
    video: { width: { ideal: 16000 }, height: { ideal: 9000 } },
    audio: false
};
```

Which was also mentioned by http://stackoverflow.com/a/38746391/1468366. See https://w3c.github.io/mediacapture-main/#dfn-fitness-distance for details of fitness computation. I'd say the above should cause a selection of maximal dimensions for pretty much all real world devices.

montaga:
Unfortunately
```
var constraints = {
    video: { width: { ideal: 16000 }, height: { ideal: 9000 } },
    audio: false
};
```
works with Firefox but not with Chromium.

gagern:
Relevant links regarding status of this in Chrome:

* [Chromium Bug 543997](https://bugs.chromium.org/p/chromium/issues/detail?id=543997)
* [Chrome platform status](https://www.chromestatus.com/feature/5807737287999488)
indicates this should be in Chrome 52 for Android or 53 for Desktop
* [Stack Overflow: “getUserMedia() video size in Firefox & Chrome differs”](http://stackoverflow.com/q/26415778/1468366)

But I can confirm this still doesn't work in Chrome 53 on my system either. I asked on the above bug report about the current status of the “ideal” property.

What's worse, we can't combine both, since Chrome will complain:

> TypeError: Failed to execute 'getUserMedia' on 'MediaDevices': Malformed constraint: Cannot use both optional/mandatory and specific or advanced constraints. TypeError: Failed to execute 'getUserMedia' on 'MediaDevices': Malformed constraint: Cannot use both optional/mandatory and specific or advanced constraints.

I'm not happy merging something into master which satisfies one browser following an obsolete API for which I can't find a proper specification, while not supporting the current standards which are already supported by other browsers. This should be no showstopper to whatever you do next Saturday, as you likely control the environment there. But if this is going into master, we should at least try to condition it on browser name or similar.

I also had a look at [`adapter.js`](https://github.com/webrtc/adapter). I would have hoped that including that into the page would magically turn Chrome or Chromium into something conforming to the latest specs. But it doesn't work that way, probably because [judging from the sources](https://github.com/webrtc/adapter/blob/06c98d2b47663c0ce6dfc6e1e15d1d372e808f1b/src/js/chrome/getusermedia.js#L33-L46) it turns an `ideal` requirement into an exact optional, with none of the magic best-fit computation of the new spec.

gagern:
Oh, and I'd also like to see some modifiers to control this, so that for applications which don't need this high resolution (and I guess these might well be in the majority), we don't have to ware resources downscaling some immense video. Actually I'm not convinced the default should be maximal resolution. If you think it should be, we probably should start a discussion to get input from the rest of the team.

How about a modifier `resolution` (or `res`?), which accepts the string `"maximal"` (and/or `"max"`?) to indicate maximal resolution, or a pair of integers `[‹int›,‹int›]` to indicate the ideal dimensions (for browsers which support `ideal`), or a single number `‹int›` to specify the width of the image? We could even extend this to `[‹int›,‹number›]` if the second element is not an integer (or smaller than 10) to denote width and aspect ratio. So perhaps I want `[1920, 4/3]`, i.e. an ideal width of 1920px but please at an aspect ratio of 4:3 if possible?

gagern:
I just added a commit, after reading [the spec](https://w3c.github.io/mediacapture-main/#dfn-applyconstraints) some more, as well as [discussing on Chrome bug 543997](https://bugs.chromium.org/p/chromium/issues/detail?id=543997#c47) and [filing Chrome bug 657145](https://bugs.chromium.org/p/chromium/issues/detail?id=657145). Since this is now essentially my code, I'll not review it myself, but I think @montaga may now qualify to test and review this.

But I still think we should have a modifier to control choice of resolution, so please add support for that if you agree with the general approach of my code.

montaga:
Your code seems okay for me and it works both for Firefox and Chromium.

I have implemented an modifier as you proposed.

If the webcam has no native support for the given resolution, also here the browsers handle it different. Firefox finds the cloeses "ideal" resolution, while chrome will scale the images exactly to the given resolution. 
I hope that things will be less cumbersome in future...

gagern:
I'll look into this tomorrow, as the other review took longer than expected.

src/js/libcs/OpImageDrawing.js
```diff
@@ -347,12 +347,102 @@ evaluator.allimages$0 = function() {
...
+        maximal = false;
+        var val = evaluate(modifs.resolution);
+        if (val.ctype === 'string' && val.value === 'maximal') maximal = true;
+        else {
+            var width = null;
+            var heightorratio = null;
+
+            if (val.ctype === 'number') {
+                width = val.value.real;
+            } else if (List.isNumberVector(val).value && (val.value.length === 2)) {
```
gagern: Better use [`List._helper.isNumberVecN`](https://github.com/CindyJS/CindyJS/blob/v0.8.2/src/js/libcs/List.js#L806).

src/js/libcs/OpImageDrawing.js
```diff
@@ -347,12 +347,102 @@ evaluator.allimages$0 = function() {
...
+                width = val.value[0].value.real;
+                heightorratio = val.value[1].value.real;
+            } else return nada;
```
gagern: Usually we ignore modifiers of invalid type, instead of failing the whole function call. Is there a reason to do things differently here?

src/js/libcs/OpImageDrawing.js
```diff
@@ -347,12 +347,102 @@ evaluator.allimages$0 = function() {
...
+
+            constraints = {
+                video: {
+                    width: width,
+                    advanced: [{
+                        width: {
+                            max: width, //ideal or exact seem not be supported by Chrome yet
```
gagern: Grammatically incorrect comment. Perhaps also indicate that there are links with details further down, e.g. “see below for details”.

src/js/libcs/OpImageDrawing.js
```diff
@@ -347,12 +347,102 @@ evaluator.allimages$0 = function() {
...
+            video: {
+                width: 16000, // ideal dimensions, will
+                height: 9000, // prefer big resolutions
+                advanced: constraints
+            },
+            audio: false
+        };
+    }
+
+    console.log(constraints);
```
gagern: Do we want to keep this? Perhaps, for the time being. In the long run it should go away, though.

PR #521, opened 2016-10-24
Complex curves plugin

gagern:
This is an alternative to #389, referencing the complex curves source code by @kranich as a ZIP file download instead of the full-text inclusion envisioned there. I promised to create a PR for this in [our last meeting](https://github.com/CindyJS/CindyJS/wiki/Meeting-2016-09-19). Sorry it took me so long to file a PR for this…

What's in this PR here?
1. Plugin code to glue complex curves and CindyJS together.
2. Examples demonstrating that plugin.
3. Build instructions specific to that plugin.
4. Some code to work with submodules, which is not used any more. Should I drop that?
5. Feature to rename directory while unzipping it.


kranich:
@gagern Thanks for the PR. Seems you missed to add `plugins/ComplexCurves/lib/ComplexCurves.commit`.

strobelm:
Hi, looks like travis crashed at some point. Could we re trigger the tests?

gagern:
The commit file is missing, and I'm currently trying to work out which one to use.

gagern:
I rebased this to mimic the current state of #389, and in particular to use the upstream commit also used there (https://github.com/kranich/ComplexCurves/commit/350b8bfc3ce93fa33f7aee46171c7e4f7fb0a9ba).

While testing this, I found that the examples won't show on my MacBook, with no error message to console. I spent quite some time trying to find a better commit, but I guess I just used my other computer when I tried this first, and indeed the branch as it is now will work there as well. Or it might be due to my upgrade of macOS to Sierra.

@kranich, do you have an idea where I might start looking for causes, in the absence of any console output? It's not a browser issue as it can be reproduced with Firefox and Chrome. It's not a local file system restriction since it can be reproduced with local HTTP server. Could this be a resource exhaustion somewhere?

These issues should not prevent us from merging the plugin, but we should be careful about advertising examples of this if we have to expect some visitors will only see black rectangles.

kranich:
@gagern Could be resource exhaustion, could be graphics card driver issue or unsupported extension (that should cause a warning though). Maybe you could try to start Firefox from the console and see if any debug output is written there.

Unfortunately, I cannot investigate the issue any further soon, since I am not at home for the rest of the week.

It would help me if you could post what webglreport.com shows for the failing platform.

kranich:
@gagern Thanks for implementing ZIP download, etc. for the build system.

Merging this as agreed at #534.

PR #525, opened 2016-10-25
Correctly obtain used fonts from KaTeX box

gagern:
This fixes a bug introduced by 8be4849eb678d3a8e8db6ce2b7bb78cee7ddb29a where we no longer detect which fonts are actually being used by a given formula, thus leading to incorrect rendering while the fonts load, and even incorrect rendering afterwards as results based on wrong metrics get cached.

### DIFF
--- a/plugins/katex/src/js/katex-plugin.js
+++ b/plugins/katex/src/js/katex-plugin.js
@@ -188,8 +188,9 @@
                     }
                 } else {
                     try {
-                        row.push(katex.canvasBox(parts[i], ctx, opts));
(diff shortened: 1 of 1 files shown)


strobelm:
Hi, is there an example where one can observe the bug?

gagern:
I've seen this in http://cindyjs.org/examples/katex1.html but it's a race condition, so exactly reproducing this can be tricky at times.

strobelm:
The examples look fine but i have to admit that i couldn't reproduce the bug. The facts (i.e. the code) looks fine. Merged.

PR #527, opened 2016-10-28
Point on Arc

gagern:
This is intended to supersede #469. Since I reverted quite a number of the changes there, and since I'm still not sure I understood why @strobelm handled segments through infinity the way he did, I'm proposing this as a separate pull request even though it builds on the commits from there (might be squashed, though, if you prefer).

My main motivation was to avoid case distinctions as far as possible:
1. No case distinction to distinguish between arc with finite center and arc which degenerates into a straight segment
2. No case distinction to distinguish segments which pass through infinity from those which don't
3. No circle-line intersection with a case distinction to choose the right solution

Avoiding all of these should help make tracing more stable. There are of course still case distinctions for restricting the point to the extent of the arc. Contrary to the PointOnSegment case, I don't preserve the imaginary part here. There is also a circle-line-intersection followed by a case distinction when projecting the point to the circle. But both of these operations only occur when moving the point along the arc, not when the arc moves from under the point. So it's only affected by free movement, not by dependent movement, and therefore should not impact complex tracing.

@strobelm, please have a look at this, perhaps compare it with what you have, and let me know whether you think I'm simplifying too much. In particular, whether there is some situation related to straight line segments through infinity, which my code doesn't handle correctly.

@kortenkamp, can you please have a look at my use of conjugation here? I would assume this to be all right, but I would feel better if you could confirm that.

In the long run we might want to look into sharing code between the `projectPointToCircle` helper and `PointOnCircle` methods, but I got bogged down when I tried this so I'd like to see the current state merged before investigating that aspect.

### DIFF
--- a/examples/146_pointOnArc.html
+++ b/examples/146_pointOnArc.html
@@ -22,7 +22,7 @@
     <link rel="stylesheet" href="http://cindyjs.org/dist/v0.7/CindyJS.css">
     <script type="text/javascript" src="../build/js/Cindy.js"></script>
     <script type="text/javascript">
-createCindy({
+CindyJS({
   scripts: "cs*",
   defaultAppearance: {
     dimDependent: 0.7,
--- a/examples/146_pointOnArc.html
+++ b/examples/146_pointOnArc.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Konstruktion.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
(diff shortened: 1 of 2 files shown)


strobelm:
Neat idea. I was a bit skeptical at first, the use of a pseudo inverse since these tend to be unstable but this should be negligible in our homogeneous setting. I could not provoke any unstable situation anyway. Have tested the code under transformation? I recently found a bug in the drawing code where arcs could be drawn incorrectly (i should file an issue).

gagern:
I think the alternative to a pseudoinverse here would be choosing two independent non-zero coordinates and finding suitable factors from those. But choosing independent coordinates again entails case distinctions, and due to those it's probably less traceable than this approach using the pseudoinverse.

strobelm:
I really like the code, so thanks very much @gagern. I would be willing to merge this if @kortenkamp and/or @richter-gebert are fine with the complex conjugation

gagern:
Rebased to take the removal of `geoOps.ReflectCArc` from #514 into account.

PR #540, opened 2016-11-23
Check result from fs.writeFile calls in build script

gagern:
This avoids a warning which otherwise gets issued by Node 7:

> DeprecationWarning: Calling an asynchronous function without callback is deprecated.

Another such warning can be avoided by upgrading whole-line-stream to 0.1.1. `npm update --depth 0` should do the trick, although the depth option is not needed for npm >= 2.6.1.

### DIFF
--- a/tools/prepare-deploy.js
+++ b/tools/prepare-deploy.js
@@ -47,6 +47,13 @@ process.once('beforeExit', function() {
     process.exit(exitStatus);
 });
 
+function check(err) {
+    if (err) {
+        console.error(err.stack);
+        exitStatus = 1;
+    }
+}
+
 child_process.execFile("git", ["rev-parse", "HEAD"], function(err, stdout, stderr) {
     if (err) {
         console.log(stdout);
@@ -84,7 +91,7 @@ function subst(name, err, content) {
     if (err) throw err;
     content = content.toString();
     content = content.replace(/\$gitid\$/, head);
-    fs.writeFile(path.join(outDir, name), content);
+    fs.writeFile(path.join(outDir, name), content, check);
 }
 
 var mapKeys = ["version", "file", "sourceRoot", "sources", "sourcesContent", "names", "mappings"];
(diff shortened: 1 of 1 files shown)


gagern:
There is also a deprecation message from `npm`, informing us that our use of the `prepublish` script is deprecated in npm ≥4 and will change in npm ≥5:

> npm WARN prepublish-on-install
>
> As of npm@5, `prepublish` scripts will run only for `npm publish`.
> (In npm@4 and previous versions, it also runs for `npm install`.)
> See the deprecation note in [`npm help scripts`](https://github.com/npm/npm/blob/v4.0.3/doc/misc/npm-scripts.md#prepublish-and-prepare) for more information.

Not sure whether we want to worry about that yet. We could declare a `prepare` script identical to our `prepublish` script, to ensure that `npm install` will run the build process at least once, no matter the version of npm in use. `prepare` was introduced in npm 4, so npm 4 would run the script twice, which should cause no harm as most steps check time stamps so the second run should be near instantaneous.

What do you think?

kranich:
I can add
```
(node:6561) DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.
```
to the list of annoying/worrying warnings (node v7.2.0, npm 4.0.2).

gagern:
@kranich, that sounds like the outdated whole-line-stream I mentioned in the original comment. Can you still reproduce the issue after updating that? If so, please run node with `--trace-deprecation` to get more details.

kranich:
@gagern Sorry, missed that. I started afresh after deleting node_modules, whole_line_stream 0.1.1 got installed and the warning disappeared.

strobelm:
Fedora doesn't offer node 7 in their repositories, so it's a bit of a hassle to test this. Anybody around with a more modern node? Otherwise i'll set up a docker container.

gagern:
Have a look at [NVM](https://github.com/creationix/nvm#readme). It allows you to install different versions of node with little hassle, and without affecting the rest of the system. That's actually what's being used by Travis, too. Should be far easier and more lightweight than a docker setup.

strobelm:
Almost forgot about that. Works fine for me. Merged.

PR #544, opened 2016-11-29
Combine moves of the same mover

### LINKED ISSUE #539: Continuity does not work
In the attached file CindyJS handles the continuity while moving the point Y different than Cinderella: In CindyJS, it seems to be impossible to switch sides after making X Y P being co-linear. In Cinderella we can always switch the sides.
screenshot

[Papierstreifen.zip](https://github.com/CindyJS/CindyJS/files/608883/Papierstreifen.zip)

gagern:
This fixes #539.

Switching the mover breaks complex tracing: as we only move one object at a time, we have to abandon the saved state and make the current state persistent if the mover changes. Up till now, CindyJS would perform this kind of mover change upon mouse up, and around scripted moves. So multiple mouse moves of the same object would not get traced correctly if the mouse was released in a singular situation in between. And multiple script commands moving the same object, or perhaps the same object that was also moved by the mouse, also caused this mover switch behavior.

This commit here reduces the number of such mover switches to when the mover actually changes. One consequence is that it's now possible to release the mouse in a singular situation without loosing tracing information. Only when another point gets moved afterwards the information is lost.

As discussed in https://github.com/CindyJS/CindyJS/issues/539#issuecomment-262991267, Cinderella apparently treats mouse move and script move as independent operations, and will take a checkpoint in between these two if the situation there is not singular. CindyJS, on the other hand, does treat mouse move and script moves as one entity, if they refer to the same object. So the automatic checkpointing will only occur before the mouse move. However, to allow more control to power users, this change introduces a function `continuefromhere()` (alternatively spelled `continue from here()` or `continueFromHere()`) which allows manual checkpointing. The example file demonstrates its use.

### DIFF
--- a/examples/EllipseWithPaper.html
+++ b/examples/EllipseWithPaper.html
@@ -1,0 +1,86 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Drawing an ellipse with a strip of paper</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <link rel="stylesheet" href="../build/js/CindyJS.css">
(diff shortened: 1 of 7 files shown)


montaga:
I have reviewed your Code and it seems okay to me.
The examples work as they should do.

gagern:
As Travis is fine with the added example, and I adapted that myself, I think you can go ahead and merge, @montaga.

PR #546, opened 2016-12-05
Unwrap result from isNumberMatrix as required

gagern:
I noticed that there were two occasions where the result of `isNumberMatrix` was not unwrapped, but used as a native boolean-like value. I also noticed that if the result here is wrapped, we can avoid unwrapping and re-wrapping it in the corresponding CindyScript operators.

Personally I think that using these wrapped types for internal interfaces should be avoided in all cases where a corresponding native type exists. But doing that consistently would affect many more functions, so it should be discussed in some meeting before making any changes.


strobelm:
Thanks for fixing this!

PR #547, opened 2016-12-05
Introduce FreeConic

gagern:
A free conic is not movable by mouse, but it can be set arbitrarily using scripted assignment to its `matrix` property.  The example of the nine-point conic demonstrates how this might be used in practice.


strobelm:
This look great, thanks Martin!

PR #548, opened 2016-12-10
performance of reverse

strobelm:
@richter-gebert drew my attention to the poor performance of reverse. This should improve the execution time.

### DIFF
--- a/src/js/libcs/List.js
+++ b/src/js/libcs/List.js
@@ -209,9 +209,9 @@ List.consecutive = function(a) {
 };
 
 List.reverse = function(a) {
-    var erg = [];
-    for (var i = a.value.length - 1; i >= 0; i--) {
-        erg.push(a.value[i]);
+    var erg = new Array(a.value.length);
+    for (var i = a.value.length - 1, j = 0; i >= 0; i--, j++) {
+        erg[j] = a.value[i];
     }
 
     return {


src/js/libcs/List.js
```diff
@@ -209,9 +209,9 @@ List.consecutive = function(a) {
 };
 
 List.reverse = function(a) {
-    var erg = [];
-    for (var i = a.value.length - 1; i >= 0; i--) {
-        erg.push(a.value[i]);
+    var erg = new Array(a.value.length);
```
kranich: Why not simply `var erg = a.value.reverse();`?

kranich: OK, I see, reverse mutates, so we would have to copy first.

kranich: Like `var erg = a.value.slice().reverse()`.

@strobelm Maybe you can compare the performance of that?

strobelm: I did some tests beforehand:
```

var n = 25000000;
var a = new Array(n);
for(let i = 0; i < n; i++) a[i] = i;

var t = Date.now();
var erg = [];
for (let i = a.length-1 ; i >= 0; i--) {
        erg.push(a[i]);
    }
console.log("time push", Date.now() -t);

t = Date.now();
var erg = new Array(a.length);
for (let i = a.length-1, j = 0; i >= 0; i--, j++) {
        erg[j]= a[i];
    }
console.log("time native", Date.now() -t);

t = Date.now();
erg = a.slice().reverse();
console.log("time slice reverse", Date.now()-t);
```
which resulted in:
```
time push 414
time native 169
time slice reverse 406
```
on Chromium. Would be great to have some iOS device tested.

gagern: I must say I'm surprised. Even accepting that the slice and reverse does touch each item of data twice instead of once, the time is still more than twice that of the native code, suggesting people would be better off reimplementing these functions instead of using the built-in versions. Strange! Too bad jsPerf isn't available.

strobelm:
It's getting even more strange. Firefox and some browsers on Windows 7 showed different results:

Firefox
time push 525  
time native 544 
time slice reverse 131

Internet Explorer
time push 7109  
time native 4660 
time slice reverse 212

Opera
time push 537
time native 662 
time slice reverse 508

I'll do some performance tests with an iPad / Macbook when i'm back in the office.

elkins0:
@strobelm: I found that using F12 in IE 11 to bring up the console starts with the debugger connected which slows everything down by 10x.

strobelm:
Here are the results for Safari on iOS:

```
time push	                 175	203	198
time native	                 109	81	79
time slice reverse	         255	252	260
```

Issue #552, opened 2016-12-14
Cindy3D fails to draw large number of triangles


gagern:
If the number of triangles to be drawn in a 3D setup is *really* large, the result looks weird.

Commit 6ee113c26428cc56547b88d399beb6c1f6481e7d switched from [`drawArrays`](https://www.khronos.org/opengles/sdk/docs/man/xhtml/glDrawArrays.xml) to [`drawElements`](https://www.khronos.org/opengles/sdk/docs/man/xhtml/glDrawElements.xml) which uses an index into the array of attributes to reference each primitive. But these indices have to be of type `UBYTE` or `USHORT`, so they can index at most 2<sup>16</sup> bytes. As each triangle is represented using 144 bytes of attribute data, representing more than 455 triangles becomes impossible.

It would be nice if Cindy3D could offer a way to draw more triangles, using multiple such calls. Not sure whether it would be feasible to use multiple buffers for this purpose, or whether a single buffer would be re-written for each of these calls. On devices that support it, using [the `OES_element_index_uint` extension](https://www.khronos.org/registry/webgl/extensions/OES_element_index_uint/) should be a simple alternative, and a good starting point.

PR #553, opened 2016-12-14
Draw larger number of primitives using uint indexing

gagern:
This requires the OES_element_index_uint extension.  If that is not available, nothing will be drawn at all since the drawElements call is invalid in that case.  So this is a partial solution to #552.

### DIFF
--- a/plugins/cindy3d/src/js/PrimitiveRenderer.js
+++ b/plugins/cindy3d/src/js/PrimitiveRenderer.js
@@ -236,7 +236,7 @@ PrimitiveRenderer.prototype.addPrimitive = function(attributes) {
         this.indexByteCount = 4;
         this.itemTotalByteCount = this.itemAttribByteCount +
           this.numElements * this.indexByteCount;
-        this.maxCount = Math.floor((1 << 31) / this.itemAttribByteCount);
+        this.maxCount = Math.floor(Math.pow(2, 32) / this.itemAttribByteCount);
       }
     }
     nd = new ArrayBuffer(c * this.itemTotalByteCount);
--- a/plugins/cindy3d/src/js/PrimitiveRenderer.js
+++ b/plugins/cindy3d/src/js/PrimitiveRenderer.js
@@ -4,9 +4,6 @@ let vec4Length = 4;
 /** @constant @type {number} */
 let float32ByteCount = 4;
 
-/** @constant @type {number} */
-let indexByteCount = 2;
-
 /**
  * @param {Array.<string>} attributes  names of vec4-typed attributes
  * @param {Array.<number>} elements  order of vertices in the drawn elements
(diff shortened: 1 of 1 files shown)


plugins/cindy3d/src/js/PrimitiveRenderer.js
```diff
@@ -219,9 +229,22 @@ PrimitiveRenderer.prototype.addPrimitive = function(attributes) {
...
   if (this.count == this.capacity) {
     let c = this.capacity*2, nd, nda, ndi, i, j, k, o, e = this.elements;
+    if (c > this.maxCount) {
+      if (this.capacity < this.maxCount) {
+        c = this.maxCount;
+      } else {
+        this.indexByteCount = 4;
+        this.itemTotalByteCount = this.itemAttribByteCount +
+          this.numElements * this.indexByteCount;
+        this.maxCount = Math.floor((1 << 31) / this.itemAttribByteCount);
```
montaga: `1 << 31` seems dangerous to me as it will be interpreted as signed, thus negative, number. But Anyway, we are sure that `this.itemAttribByteCound` is sufficiently large :)
Actually, for uint32 we would like to have something like `(1<<32-1)` (which is different from -1 in javascript) but I guess this is more hypothetical as these numbers are rather beyond our scope and `this.maxCount` will never be used after that.

plugins/cindy3d/src/js/PrimitiveRenderer.js
```diff
@@ -23,7 +20,8 @@ function PrimitiveRenderer(attributes, elements) {
   this.itemLength = vec4Length * numVertices * numAttributes;
   this.vertexByteCount = vec4Length * float32ByteCount * numAttributes;
   this.itemAttribByteCount = tmp = numVertices * this.vertexByteCount;
-  this.itemTotalByteCount = tmp + numElements * indexByteCount;
+  this.itemTotalByteCount = tmp + numElements * this.indexByteCount;
+  this.maxCount = Math.floor((1 << 16) / tmp);
```
montaga: We also need a declaration like `PrimitiveRenderer.prototype.maxCount;`

plugins/cindy3d/src/js/PrimitiveRenderer.js
montaga: Actually, we would like to have `2^32` here. With the division we will use doubles anyway.
So we could simply use `Math.pow(2,32)`, or if you like bit-shifts `2*((-1^(1<<31))+1)` ;-)

kortenkamp:
Isn't the maxCount trying to find out how many things can be accommodated? So why not shift by the base-2-log of the number of items, i.e. shift both until number of items is 0?

plugins/cindy3d/src/js/PrimitiveRenderer.js
gagern: That is included in line 99 of the modified file.

plugins/cindy3d/src/js/PrimitiveRenderer.js
gagern: I think `Math.pow` is the appropriate thing to use, with `(1<<16)*(1<<16)` as an alternative I'd find agreeable as well.

gagern:
@kortenkamp: It's maximal capacity divided by item size. The former is a power of two, the latter is 156 or something like that. Doing that division using bit shifts is more trouble than it's worth. And rouding 156 up to the next power of two would be quite wasteful.

PR #554, opened 2016-12-19
Make colorplot and reference manual match

montaga:
This completes the reference manual such that it fits the implemented colorplot command.
Furthermore, colorplot now also supportes running varaibles different from # as it is documented.


strobelm:
Looks fine to me! Thanks!

PR #555, opened 2016-12-20
Conic drawing based on boundary crossings

gagern:
This is an alternative to #280, mostly based on https://github.com/CindyJS/CindyJS/pull/280#issuecomment-266962274 and some code by @elkins0 from that issue. The number of epsilon-based case distinctions has been reduced considerably. Instead of maintaining a list of pairs of intersections, sorted by one coordinate, we now have a list of boundary crossings sorted in cyclic order. Subsequent pairs of boundary points are to be connected by an arc, in an alternating fashion. Figuring out which ones to connect took some work, so here are the rules:

* For zero boundary crossings, do the segmentation of an ellipse as described below. If all four segmentation points are numerically well defined and within the rectangle, draw a closed ellipse.
* For two boundary crossings, these two need to be connected by an arc.
* For four boundary crossings, check whether diagonal pairs of points are on different branches of a hyperbola. That's the case if the line joining these crosses the line at infinity inside the conic. If it's different branches, the points to connect have the boundary between them inside the conic, else the boundary between them is outside. This combination of two signs is robust even if the conic degenerates to a pair of lines, so that “inside” and “outside” are arbitrary definitions. If the point of intersection is off-canvas, this will lead to the right result.
* For more than four crossings, it has to be an ellipse and the points to connect always have an outside boundary between them, not an inside one.

For ellipses, just connecting the boundary points doesn't always work. Adaptive refinement may pick the wrong arc there, i.e. the one outside the visible rectangle. So if the sign of the discriminant indicates an ellipse, we also identify the points of vertical or horizontal tangency. We mix them with the boundary points and sort everything by angle.


elkins0:
@gagern: This addresses drawing a conic arc using a rational B-spline curve as a guide.

It occurred to me that even though a canvas drawing primitive using a weighted parametric equation is not available to do the actual drawing, it can still be used in place of the conic matrix as the guide for selecting Bézier curves to approximate it.

A Bézier curve with the parametric equation `(1-u)²·A+2·(1-u)·u·B+u²·C` fits any parabolic arc. Adding a weight `w` to that equation gives a rational B-spline curve with the parametric equation `(1-u)²·A+2·(1-u)·u·w·B+u²·C` which fits any conic arc simply by using the correct weight `w`, a value between zero and infinity.

Using a variable substitution of `k/(1-k)` for `w`, where k runs from zero to one, allows for simpler equations in what follows. Note that when `k=1/2`, `w=1`.

Show in the diagram below, `F` lies on the Bézier curve and the point `G` lies on a rational B-spline curve with `k=2/3` (`w=2`). Points `A` and `C` are the starting and ending points of the curve with `B` being the control point. The point `M` is midpoint of `A` and `C`. The parametric equation for the segment `M` to `B` is `u·B+(1-u)·M` so the point `F` lies at `u=0.5` and the point `G` lies at `u=k`. To recurse, new points `D` and `E` are chosen, with the segment `D` to `E` parallel to the segment `A` to `C` with `kr=1/√(2(1-k))+1)`.

Within the recursion, new points for `F` and `G` determined. The points `D` and midpoint `M2` along with the new value for `k`, `kr`, gives the new point `G2`. Similarly, the points `E` and midpoint `M3` along with the new value for `k`, `kr`, gives the new point `G3`. Recursion proceeds until the points `F#` and `G#` are within a half pixel of each other. Note that the line thru `M#` and `G#` passes thru the conic center; in the case of a parabola that line is parallel to its axis.
drawconicark

This code adds adjoint matrix `k##` and conic center values, replaces the refine function and introduces `drawConicArk`. The `discr` and `det` values are now derived from the `k##` values which are not affected by changing sign of all `c##` values:

```JavaScript
    //-var discr = c20 * c02 * 4 - c11 * c11;
    //-var det = discr * c00 + c11 * c01 * c10 -
    //-    c20 * c01 * c01 - c02 * c10 * c10;
    // The adjoint matrix k## values
    var k20 = 4 * c00 * c02 - c01 * c01;
    var k11 = c01 * c10 - 2 * c00 * c11;
    var k10 = c01 * c11 - 2 * c02 * c10;
    var k02 = 4 * c00 * c20 - c10 * c10;
    var k01 = c10 * c11 - 2 * c01 * c20;
    var k00 = 4 * c02 * c20 - c11 * c11; // discr -4(b * b - a * c)
    var discr = k00;
    var det = c02 * k02 + c11 * k11 + c20 * k20 - c00 * k00;
    // conic center [ccx / ccz, ccy / ccz]
    var ccx = k10 / k00;
    var ccy = k01 / k00;
    // The k## values not affected by changing sign of all c## values
    if (det < 0) {
        c20 = -c20;
        c11 = -c11;
        c10 = -c10;
        c02 = -c02;
        c01 = -c01;
        c00 = -c00;
        det = -det;
    }
    
    // Draw a conic arc from A to C, with control point B & a weight of k/(1-k)
    function drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, n) {
        var j = 1 - k;
        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);
        var Fx = 0.5 * (Bx + Mx);
        var Fy = 0.5 * (By + My);
        var kBx = k * Bx;
        var kBy = k * By;
        var Gx = kBx + j * Mx;
        var Gy = kBy + j * My;
        if ((Fx - Gx) * (Fx - Gx) + (Fy - Gy) * (Fy - Gy) < 0.25 || n > 10) {
            csctx.quadraticCurveTo(Bx, By, Cx, Cy);
        } else {
            var kr = 1 / (Math.sqrt(2 * j) + 1);
            ++n;
            drawConicArk(Ax, Ay, kBx + j * Ax, kBy + j * Ay, Gx, Gy, kr, n);
            drawConicArk(Gx, Gy, kBx + j * Cx, kBy + j * Cy, Cx, Cy, kr, n);
        }
    }

    // Find the control points of a quadratic Bézier which at the
    // endpoints agrees with the conic in position and tangent direction.
    // sign is -1 for the lower branch and +1 for the upper.
    function refine(pt1, pt2) {
        var Ax = pt1.px;
        var Ay = pt1.py;
        var Cx = pt2.px;
        var Cy = pt2.py;

        var ux = Ay - Cy;
        var uy = Cx - Ax;
        var uz = Ax * Cy - Cx * Ay;

        //pole = {ux, uy, uz}.{{k20, k11, k10}, {k11, k02, k01}, {k10, k01, k00}};
        var Bz = (k10 * ux + k01 * uy + k00 * uz);
        if (Math.abs(Bz) < CSNumber.eps) {
            return csctx.lineTo(Cx, Cy);
        }
        var Bx = (k20 * ux + k11 * uy + k10 * uz) / Bz;
        var By = (k11 * ux + k02 * uy + k01 * uz) / Bz;

        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);
        // Note: The line thru M and B also passes thru the conic center [ccx/ccz,ccy/ccz]
        var rdx = Bx - Mx;
        var rdy = By - My;

        var k = 0.5;
        if (k00 !== 0) {
            var r = Math.sqrt(-2 * det / (
                My * k00 * (c01 + c11 * Mx + 2 * c02 * My) +
                Mx * k00 * (c10 + 2 * c20 * Mx + c11 * My) -
                (k10 - Mx * k00) * c10 - (k01 - My * k00) * c01));
            if (Math.abs(rdy) < Math.abs(rdx)) {
                k = ((ccx - Mx) * (1 - r) / rdx);
            } else {
                k = ((ccy - My) * (1 - r) / rdy);
            }
        }

        if (!(k >= 0 && k <= 1)) {
            console.log("out-of-range: k=" + k + " r=" + r + " det=" + det + " ccx=" + ccx + " ccy=" + ccy);
            k = k < 0 ? 0 : k > 1 ? 1 : 0.5; // NaN gives 0.5
        }
        console.log("drawConicArk(" + Ax + ", " + Ay + ", " + Bx + ", " + By + ", " + Cx + ", " + Cy + ", " + k + ", 0)");
        drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, 0);
    }
```

elkins0:
Just an additional note. Given any two points which lie on the conic, the control point for either of the parametric equations is the intersection of the tangents to the conic at those two points. That control point is also the polar of the line thru those two points, computed as the product of the homogeneous equation of the line and the adjoint of the conic matrix. *Mathematica* syntax:
```Mathematica
        {Bx, By, Bz} = {ux, uy, uz}.{{k20, k11, k10}, {k11, k02, k01}, {k10, k01, k00}};
```

src/js/libcs/OpDrawing.js
```diff
@@ -568,9 +568,9 @@ eval_helper.drawconic = function(conicMatrix, modifs) {
         // Compute the roots of the x discriminant
         // for points with horizontal tangents
         sol = solveRealQuadratic(
-            c11 * c11 - 4 * c02 * c20,
-            2 * c10 * c11 - 4 * c01 * c20,
-            c10 * c10 - 4 * c00 * c20);
+            discr,
+            4 * c01 * c20 - 2 * c10 * c11,
+            4 * c00 * c20 - c10 * c10);
```
elkins0: In terms of adjoint matrix `k##` values this becomes:
```JavaScript
        sol = solveRealQuadratic(k00, -2 * k01, k02);
```

elkins0:
In terms of adjoint matrix `k##` values these changes are possible:
```JavaScript
         // Compute the roots of the y discriminant
         // for points with vertical tangents
-        sol = solveRealQuadratic(
-            discr,
-            4 * c10 * c02 - 2 * c01 * c11,
-            4 * c00 * c02 - c01 * c01);
+        sol = solveRealQuadratic(k00, -2 * k10, k20);
```
and
```JavaScript
         // Compute the roots of the x discriminant
         // for points with horizontal tangents
-        sol = solveRealQuadratic(
-            discr,
-            4 * c01 * c20 - 2 * c10 * c11,
-            4 * c00 * c20 - c10 * c10);
+        sol = solveRealQuadratic(k00, -2 * k01, k02);
```

elkins0:
@gagern: I came with a much more straight forward way to compute the value of `k` as the result of a thought experiment:
```JavaScript
        // Plug M and B into conic equation, taking square root of the results to get a measure
        var sM = Math.sqrt((c20 * Mx + c10) * Mx + (c02 * My + c01) * My + c11 * Mx * My + c00);
        var sB = Math.sqrt(-((c20 * Bx + c10) * Bx + (c02 * By + c01) * By + c11 * Bx * By + c00));
        var k = sM / (sM + sB);

        if (!(k >= 0 && k <= 1)) {
            console.log("out-of-range: k=" + k + " sM=" + sM + " sB=" + sB);
            k = 1.0;
        }
```

examples/63_Conic.html
```diff
@@ -36,12 +37,55 @@
...
+                ["degen7", [-0.09883720930232559, 1, 0.1453488372093023], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
+                ["degen8", [-0.5111111111111112, 1, 0.0925925925925926], [0.4, 1, 0.2], [1, 0.3333333333333333, 0.16666666666666666], [1, 0.6666666666666666, 0.3333333333333333], [1, 1, 0.5]],
+                ["almost degen", [0,1,0.25], [0.9999999999999999,0.16666666666666666,0.09689922480620154], [0,1,-0.5], [1,0.23931623931623938,0.21367521367521372], [1,0,-0.16666666666666666]],
+                ["almost2", [0,1,0.25], [1,0.16666666666666674,0.09689922480620156], [0.6105263157894737,1,-0.26315789473684215], [1,0.24545454545454548,0.2272727272727273], [1,0,-0.16666666666666666]],
+                ["almost3", [0,1,0.25], [1,0.7410071942446044,0.1798561151079137], [-0.8492063492063492,1,-0.1984126984126984], [1,0,0.2], [1,0,-0.16666666666666666]],
+                ["circle1", [6,1,1], [10,3,1], [10,9,1], [2,9,1], [2,3,1]],
```
gagern: The `opps#` look to me like four collinear points and a fifth one elsewhere, so the conic would be ill defined in that case. The randomness we see is not related to the drawing step, but happens before.

gagern:
@strobelm: The branch now does handle all the test cases we have collected so far. I would say this is ready for merging, unless you can come up with some use case where this leads to results that are considerably worse than those of the original segment-based code.

@elkins0: The idea of basing this on rational B-splines is quite interesting. But after I copied your snippets above into my current code, a number of test cases which used to work will no longer work correctly. So getting this right may require some more time, and seeing how much time we spent on all of this already, I don't dare judge how much longer. Therefore I'd like to get the improvements we have so far into master *now*, and suggest we develop those ideas in a separate pull request. Do you want to open a pull request, using a branch of yours as the source?

One thing I thought of: building the drawing on rational splines relies on the fact that these can represent a conic arc *exactly*. So this makes the solution increasingly specific to conics only. My original choice of names for `c20` and so on was to indicate the powers of `x` and `y` in each of these, in the hope that one day we might have a cubic drawing function using the same approach. This doesn't make too much sense for the adjoint, though, and is at odds with the rest of the codebase. So if you go for the B-spline approach, feel free to rename variables to the more common row-and-column numbers or whatever you consider best.

elkins0:
@gagern: I have a better formulation of `k` now which uses only one square root. Exactly which cases are failing? When I use it with your branch I'm not getting any failures!  Here is exactly what the entirety of my code looks like now:
```JavaScript
    // The adjoint matrix k## values
    var k20 = 4 * c00 * c02 - c01 * c01;
    var k11 = c01 * c10 - 2 * c00 * c11;
    var k10 = c01 * c11 - 2 * c02 * c10;
    var k02 = 4 * c00 * c20 - c10 * c10;
    var k01 = c10 * c11 - 2 * c01 * c20;
    var k00 = 4 * c02 * c20 - c11 * c11;
    
    // Draw a conic arc from A to C, with control point B & a weight of k/(1-k)
    function drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, n) {
        var j = 1 - k;
        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);
        var Fx = 0.5 * (Bx + Mx);
        var Fy = 0.5 * (By + My);
        var kBx = k * Bx;
        var kBy = k * By;
        var Gx = kBx + j * Mx;
        var Gy = kBy + j * My;
        if ((Fx - Gx) * (Fx - Gx) + (Fy - Gy) * (Fy - Gy) < 0.25 || n > 10) {
            csctx.quadraticCurveTo(Bx, By, Cx, Cy);
        } else {
            var kr = 1 / (Math.sqrt(2 * j) + 1);
            ++n;
            drawConicArk(Ax, Ay, kBx + j * Ax, kBy + j * Ay, Gx, Gy, kr, n);
            drawConicArk(Gx, Gy, kBx + j * Cx, kBy + j * Cy, Cx, Cy, kr, n);
        }
    }

    // Find the control points of a quadratic Bézier which at the
    // endpoints agrees with the conic in position and tangent direction.
    function refine(pt1, pt2) {
        var Ax = pt1.px;
        var Ay = pt1.py;
        var Cx = pt2.px;
        var Cy = pt2.py;

        var ux = Ay - Cy;
        var uy = Cx - Ax;
        var uz = Ax * Cy - Cx * Ay;

        // B = {ux, uy, uz}.{{k20, k11, k10}, {k11, k02, k01}, {k10, k01, k00}}
        var Bz = (k10 * ux + k01 * uy + k00 * uz);
        if (Math.abs(Bz) < CSNumber.eps) {
            return csctx.lineTo(Cx, Cy);
        }
        var Bx = (k20 * ux + k11 * uy + k10 * uz) / Bz;
        var By = (k11 * ux + k02 * uy + k01 * uz) / Bz;

        var Mx = 0.5 * (Ax + Cx);
        var My = 0.5 * (Ay + Cy);

        // Plug M and B into conic equation, b = mat.B.B and m = mat.M.M
        var b = (c20 * Bx + c10) * Bx + (c02 * By + c01) * By + c11 * Bx * By + c00;
        var m = (c20 * Mx + c10) * Mx + (c02 * My + c01) * My + c11 * Mx * My + c00;

        // Conceptually k == √m / (√m + √b)
        var k = 1 / (1 + Math.sqrt(-b / m));

        // Handle b == 0 m == 0 and resulting 0/0 == NaN when A == B == C == M
        if (isNaN(k)) k = 1;
        drawConicArk(Ax, Ay, Bx, By, Cx, Cy, k, 0);
    }
```

strobelm:
I'm sorry but we have vanishing conics:
conic

src/js/libcs/OpDrawing.js
```diff
@@ -470,9 +470,21 @@ eval_helper.drawconic = function(conicMatrix, modifs) {
...
+    var k02 = 4 * c00 * c20 - c10 * c10;
+    var k01 = c10 * c11 - 2 * c01 * c20;
+    var k00 = 4 * c02 * c20 - c11 * c11;
+
+    var discr = k00;
+    var det = c02 * k02 + c11 * k11 + c20 * k20 - c00 * k00;
+
+    // conic center
+    var ccx = k10 / k00;
+    var ccy = k01 / k00;
```
elkins0: @gagern: This code is no longer needed:
```JavaScript
    var det = c02 * k02 + c11 * k11 + c20 * k20 - c00 * k00;

    // conic center
    var ccx = k10 / k00;
    var ccy = k01 / k00;

    if (det < 0) {
        c20 = -c20;
        c11 = -c11;
        c10 = -c10;
        c02 = -c02;
        c01 = -c01;
        c00 = -c00;
        det = -det;
    }
```

elkins0: Opps...I take that back. I see you are now using the conic center `ccx` and `ccy` values later in your own code so those values need to stay 😳 However `det` is used nowhere else and `k` is not affected by the sign change to the `c##` values. Also the only use of `discr` in `if (discr <= 0)` could be replaced with `k00` and thus `var discr = k00;` would not be needed.

gagern: We need the determinant non-negative so we know which part of the conic is inside and which is outside, as computed by the `sign` function that is being used in a number of places. For some only a change in sign is relevant, but for others the sign of a single point is relevant, too.

elkins0:
@gagern: Here is something like @strobelm's test case, and yes it vanishes:
```JavaScript
                ["vanish1", [1,0.6666666666666666,0.16666666666666666], [-0.375,1,0.125], [1,0,-0.2], [1,0,0.16666666666666666], [0.9999999999999999,-0.14736842105263156,0.2631578947368421]],
```

elkins0:
@gagern: I created a pull request [#7](https://github.com/gagern/CindyJS/pull/7) to fix-up the [gagern:conicDrawing4](https://github.com/gagern/CindyJS/compare/conicDrawing4) branch using [commit 03169f9 ](https://github.com/gagern/CindyJS/pull/7/commits/03169f9ec126219f8e96a7235e630bba35d3e9f5) so that it works at least as well as the `gagern:conicDrawing3 ` branch.
The pull request is titled:
"Use iterative refinement guided by B-spline for conic drawing, cleanup and fix computation of k."

elkins0:
@gagern: Here are a couple problem cases (`var csctx = new DbgCtx();` is active).
Note that `63_Conic.html` requires the following change in order for these behaviors to be seen:
```JavaScript
-  geometry:gslp});
+  geometry:gslp,
+     ports:[{id:"CSCanvas",width:600,height:600,transform:[{visibleRect:[-10,14,14,-10]}]}]});
```
The results are differ for *Edge* versus *Firefox* and *Chrome* for `parabola2b` in that the blue traces are only drawn by *Edge*; the others only show the points:
```JavaScript
["parabola2b", [0,0,1], [0,1,0.5], [1,0.33333333333333326,0.16666666666666663], [0.33333333333333326,1,0.16666666666666666], [1,0,0.5]],
```
parabola2b
For `hyperbola5` you get the correct result on the  first click and scribbles as a result on subsequent clicks:
```JavaScript
["hyperbola5", [1,1,-0.1], [0.8703703703703702,1,0.154320987654321], [-0.7142857142857142,1,0.07142857142857142], [1,1,0.07142857142857142], [1,-0.7142857142857142,0.07142857142857142]],
```
hyperbola5

elkins0:
@gagern: In `doBoundary` I find the `((1 - axis) * coord + x, axis * coord + y)` pattern of code used to select passing `(coord, y)` versus `(x, coord)` based on `axis` to `mkp` and `sign` to be rather convoluted.

Using `axis ? mkp(other, coord) : mkp(coord, other)` for example is much clearer, replacing the `x` and `y` value with just one `other` value. Here is the full coding for `doBoundary` and use following it:
```JavaScript
    function doBoundary(sign1, sign2, sol, other, axis, sort, extent) {
        var coord;
        if (sign1 !== sign2) { // we need exactly one point of intersection
            if (sol === null)
                return false;
            var center = extent * 0.5;
            var dist0 = Math.abs(center - sol[0]);
            var dist1 = Math.abs(center - sol[1]);
            coord = sol[dist0 < dist1 ? 0 : 1];
            boundary.push(axis ? mkp(other, coord) : mkp(coord, other));
        } else { // we need zero or two points of intersection
            if (sol === null) // have zero intersections
                return true;
            coord = 0.5 * (sol[0] + sol[1]);
            if (!(coord > 0 && coord < extent))
                return true;
            var signMid = axis ? sign(other, coord) : sign(coord, other);
            if (signMid === sign1) // intersections outside segment
                return true;
            if (isNaN(signMid))
                return true;
            // Have two points of solution
            sort = sort * (sol[1] - sol[0]) > 0 ? 0 : 1;
            coord = sol[sort];
            boundary.push(axis ? mkp(other, coord) : mkp(coord, other));
            coord = sol[1 - sort];
            boundary.push(axis ? mkp(other, coord) : mkp(coord, other));
        }
        return true;
    }

    if (!(doBoundary(tl, bl, yLeft, 0, 1, 1, csh) &&
            doBoundary(bl, br, xBottom, csh, 0, 1, csw) &&
            doBoundary(tr, br, yRight, csw, 1, -1, csh) &&
            doBoundary(tl, tr, xTop, 0, 0, -1, csw)))
        return;
```

gagern:
Thanks @elkins0 for these test cases. How did you come up with these? In particular, what made you think of slightly changing the visible rectangle?

Debugging `parabola2b` I found the following: during segmentation, the parabola gets treated like a hyperbola, not like an ellipse, so it doesn't get segmented. Which is fine. Then during the first step of iterative refinement, I had a problem in [this piece of code](https://github.com/gagern/CindyJS/blob/ced9bd3e33ab457f88efe135182d573e2fabaf5c/src/js/libcs/OpDrawing.js#L726-L749). For reasons of symmetry, `c` (the pole of the segment = proposed control point) and `h` (the midpoint of the segment) lie on the axis of symmetry. So one point of intersection will be almost at infinity. But only almost; it is finite and happens to have a positive sign. That confuses the choice of which solution to pick, which is awfully complicated now that I think about it.

`hyperbola5` is a very special situation. I can't reproduce that one in the original coordinate system, as even this slight change alters some rounding to make the problem disappear. The core of the problem is that the corners of the rectangle lie pretty much on the conic. When the signs of two such corners differ, then there has to be exactly one point of intersection in between, but picking one based on distance from the center of the edge was unstable in this case. I've now added an alternative criterion to make that choice based on the sign of the midpoint between the two solutions, which should be more reliable.

How does this look now?

elkins0:
@gagern:
>How did you come up with these?

I wanted to test the situation where the boundary intersections were shared which occurs in the corners of the drawing area. The slightly changed the visible rectangle was happenstance from when I had been testing this a while back and wanted smaller rational numbers to work with in *Mathematica*. Fortunately I also tested this case in the original file and noted that the change was necessary to reproduce the effect I saw.
>How does this look now?

Now the `parabola2b`, `hyperbola5` and `vanish1` cases are giving the expected results.

However, I found another case which I tested as before using the slightly changed the visible rectangle:
```JavaScript
["circle2b", [11,2,1], [-1,14,1], [-1,-10,1], [1,"i",0], [-1,"i",0]],
["circle2c", [-7,2,1], [5,14,1], [5,-10,1], [1,"i",0], [-1,"i",0]],
```
Again you get the correct result on the first click and difficulties on subsequent clicks. What interesting here is that both cases have the same cause; `drawArc` is called to draw a 180° arc on subsequent clicks instead of two separate 90° arcs as is done on the initial click. With `var csctx = new DbgCtx();` active, you can see that the center point of the circle and the point for the 90° have vanished. `refine` simply happens to draw the 180° arc in the correct direction in the `circle2b` case but draws a vertical segment in the `circle2c` case drawing attention to the difficulty.

elkins0:
@gagern: Here's another test case, five regularly spaced points in a line. The matrix for this case is 
`[[0.04,-0.08,-0.2],[-0.08,0.16,0.4],[-0.2,0.4,1]]`. The test case is:
```JavaScript
["sameLines", [1,-2,1], [-7,-6,1], [-3,-4,1], [9,2,1], [5,0,1]],
```

gagern:
I'm not sure we should put any energy into fixing your most recent examples, @elkins0, as I consider them *extremely* rare. For five points in a line, the conic isn't defined, and anything that says otherwise is merely the result of some numeric noise. Personally I don't care whether we draw nothing for this input, or random lines, or whatever. Garbage in, garbage out. If we really want to address such situations, we might want to do so as part of the defining geometric operation. If the input is poor and the resulting matrix can be expected to contain mostly numeric noise, then it might be better to set the matrix to exactly equal the null matrix, suppressing random results due to that noise.

The case of a circle (or ellipse, for that matter) exactly touching the boundary of the visible rectangle is slightly different. There we are dealing with a number of problems at the same time. We have coinciding points of intersection, which should be fine. We have a pole (proposed control point) at infinity which is a problem. To avoid the latter, we need to segment the ellipse , and as part of that we need to sort points by angle while keeping track of which of them we need to connect. But now the point of contact would appear three times in the list, as two intersections and one axis-aligned tangent. Simply sorting these by angle has a *very* high chance of resulting in inconsistencies. I don't have a short and elegant solution here, and I think that the situation is rare enough that we can ignore it for now. But I'd welcome input from other members of the team on this.

elkins0:
@gagern: Commit 5450591f934e0ac0937b06598d596e521dc5668b shows my very much simplified,  shorter and perhaps more elegant solution to boundary and drawing order which also implements closed paths for filling and clipping. `filled:true` can now be used with conic elements.

elkins0:
@gagern: Commit 93379fbb0abe70a45a9abbbfbe42a1e59f6ecdd3 now handles all the test cases brought to light here.

>I think I'd like to see a few more closePath calls.
...
>Makes me wonder whether we should do all our intersections with a rectangle which is the viewport plus half the current line width, or something like that…

The later would be the best way to handle that.

strobelm:
Concerning the borderline case of 5 colinear points i personally would not put any energy into that. I could not reproduce the ellipse case (by hand).

gagern:
There is a know problem if a circle or ellipse barely touches the rectangle in two opposite points, and @kortenkamp said we should try to handle that case correctly. I think I got it reproduced with the alternate visible rectangle.

gagern#7 contains some discussion of yet another approach, which has some benefits. Either of these should be better than what we have so far. I have no objection against you merging this one here now, even though it might not be then end of the story. I'll definitely give an overview about the alternatives in our next video conference. If you want to wait with merging until after that, this would be fine, too. I'd also be interested in your opinion as to which approach you consider easier to maintain in the long run.

I just resolved a conflict, so this should be mergable now.

elkins0:
>There is a know problem if a circle or ellipse barely touches the rectangle in two opposite points, and @kortenkamp said we should try to handle that case correctly.

Commit 93379fbb0abe70a45a9abbbfbe42a1e59f6ecdd3 handles that case where it adds `±eps` in the `0 - eps < x && x < csw + eps` and
`0 - eps < y && y < csh + eps` checks in the `"split"` section of `doBoundary` where each appears twice as shown [here](https://github.com/elkins0/CindyJS/blob/93379fbb0abe70a45a9abbbfbe42a1e59f6ecdd3/src/js/libcs/OpDrawing.js#L612-L626).
>Concerning the borderline case of 5 colinear points i personally would not put any energy into that. I could not reproduce the ellipse case (by hand).

 Commit 93379fbb0abe70a45a9abbbfbe42a1e59f6ecdd3 handles that case with by changing `if (!(d >= 0)) return null;`, originally within `solveRealQuadraticHomog`, now inside `solveRealQuadratic` to:
```JavaScript
    // Treat tiny bit negative as zero else null if more negative or NaN 
    if (!(d >= 0)) 
        if (d >= -CSNumber.eps * (b * b)) d = 0; 
        else return null; 
```
See those [three lines](https://github.com/elkins0/CindyJS/blob/93379fbb0abe70a45a9abbbfbe42a1e59f6ecdd3/src/js/libcs/OpDrawing.js#L385-L387) in context.

elkins0:
Based on @gagern's conicDrawing4 branch, [#9](https://github.com/gagern/CindyJS/pull/9) implements support for the filling of conic elements using eval_helper.drawconic. [63_Conic.html](https://github.com/gagern/CindyJS/blob/50bd48b4c745bd44a0dfa27b8539a2e7db9838d1/examples/63_Conic.html) demonstrates this using an appearance manipulation section.

strobelm:
Hi, i just wanted to review the conic drawing. Could you give me a hint on which branches to look at? I think there were two alternatives.

elkins0:
Several months ago I read in [Meeting 2017 01 17](https://github.com/CindyJS/CindyJS/wiki/Meeting-2017-01-17) that @richter-gebert was going to have a look at the current [gagern:conicDrawing3](https://github.com/gagern/CindyJS/tree/conicDrawing3) and [elkins0:conicDrawing4](https://github.com/elkins0/CindyJS/tree/conicDrawing4) branches, in order to decide which approach would be better in the long run. Has that analysis been completed?

strobelm:
Hi! It took a while but i finally came to a conclusion. After finding an example in which my implementation greatly fails (ouch!) 

screenshot_20170929_172418

i reconsidered the PRs here and after finding that @elkins0 branch has some vanishing configurations (bottom left) 

screenshot_20170929_164926
screenshot_20170929_164942

and the version of @gagern works fine for all configurations i'll merge gagern/conicDrawing3. 

Thanks a lot for all the work and @elkins0 if you want to fix the bugs in your branch i'll consider merging this too (of course).

PR #556, opened 2016-12-29
Fix version of js-beautify

gagern:
Different versions of js-beautify will format our codebase in different ways, even for patch-level version increments.  So I suggest we fix the version the hard way, with the intention of bumping the version every now and then in a conscious update, containing all related code formatting changes in that same commit.

This commit does *not* use the latest version at the time of the commit. That would be 1.6.7 which improves [one line of code](https://github.com/CindyJS/CindyJS/blob/8422099f65445e3f67a12ece5c4e22638ca464d0/plugins/cindygl/src/js/CodeBuilder.js#L296) thanks to a fix for https://github.com/beautify-web/js-beautify/issues/1079, but in [other](https://github.com/CindyJS/CindyJS/blob/v0.8.2/src/js/libcs/OpDrawing.js#L325-L329) [places](https://github.com/CindyJS/CindyJS/blob/v0.8.2/src/js/Events.js#L496) it suffers from https://github.com/beautify-web/js-beautify/issues/1090 and https://github.com/beautify-web/js-beautify/issues/1085#issuecomment-269628467 even though I don't know why earlier versions don't affect us there, too.

I'd like to get this one here in soon, and then hope for the two issues mentioned above to get addressed. And then perhaps @montaga can have a look at bumping to something more recent, as most of the remaining differences are located in the CindyGL codebase.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "concat-with-sourcemaps": "^1.0.4",
     "glob": "^7.0.3",
     "http-proxy": "^1.11.2",
-    "js-beautify": "^1.5.10",
+    "js-beautify": "1.6.2",
     "jshint": "^2.8.0",
     "marked": "^0.3.5",
     "mocha": "^3.1.2",


elkins0:
Having to stay current manually usually leads to procrastination and eventually having to make many changes all at once in order to catch up. Depending on just how much work and thus how error prone such a thing would be, my personal preference would be to make liberal use of curly braces. That would at least handle beautify-web/js-beautify#1079 and beautify-web/js-beautify#1090 in a reasonable way. However there does not appear an option to force the issue by actually requiring curly braces following `if`, `else`, `for`, `do` and `while` to avoid having control statements followed by naked single statements. In the case of blank lines preceding functions I can see where the temptation to force the version temporarily would be stronger. It's really a case of being exposed to the whim of external code changes which have unintended consequences.

gagern:
Having to stay current with some external library or API is a pain. But I don't see much motivation staying current with js-beautify: as long as it provides a consistent and sane coding style for our project, I think we can stick to one version for considerable time, updating only when we find that some idiom we like to use gets handled poorly by the current version but correctly by the next one. And even then, making all required changes is a question of automatically beautifying everything in the code base, and having a look to see that it doesn't add more confusion than it removes. I'd rather have such a commit every half a year or so, than having spurious indentation changes as unintended consequences of an unrelated commit.

I know that the amount of curly braces is somewhat controversial. I've worked with projects that require curly braces for all control constructs that allow them, and although I can see the merit in how they enforce consistent coding, I think several rows of closing braces make the code considerably longer without adding any value. Consider for example this snippet from an older version of the conic drawing branch:

```js
sol = solveRealQuadratic(k00, -2 * k10, k20);
if (sol)
    for (i = 0; i < 2; ++i)
        if (sol[i] >= 0 && sol[i] <= csw)
            boundary.push(
                mkp(sol[i], -0.5 * (c11 * sol[i] + c01) / c02));
```

If I had to write that using braces, that would add three more rows of closing braces and in my opinion add zero value. The same holds for the various sanity checks we use for type checking, overflow detection and the likes. Things like `if (arg0.ctype !== "number") return nada;` or `if (!isFinite(x)) return;` are easily written as a single line and in that fashion don't distract from the main flow of control. Having to use three lines for each of these takes up too much space, in my opinion.

I know that this view is contrary to many style guides, so I'd be willing to abide by a decision to the contrary within the team. But I would want such a decision to at least take these use cases into consideration, and not just blindly follow some style guide. And I'd definitely not want to adapt such a style change just to work around some bugs in our beautifier. I know KaTeX uses [eslint](http://eslint.org/) to style-check their sources, including required braces. I think it offers some more options than js-beautify. I have no experience whether its `--fix` mode is equivalent to the automated beautification js-beautify performs, but I guess it might. So if we are unhappy with js-beautify in some way, we might want to have a closer look at that.

The other changes I see between version 1.6.4 and 1.6.7 affects indentation of various CindyGL functions. I haven't even started to work out why these changed, but as the new indentation looks better to me, I guess it might be intentional. Nevertheless, if we don't fix the version but allow both 1.6.4 and 1.6.7, then different developers beautifying the same code in turns will switch between these versions, swamping the git history with noise. Plus I have no idea what kind of intentional or unintentional style changes future versions of js-beautify may bring, so I'd not assume that adding braces and fixing some indentation would guard us against those.

bitwiseman:
@gagern - The beautifier has extensive tests to check for changes in formatting, but obviously never enough.  If we had seen the changes you've reported, we wouldn't have released the changes as a dot release, but at least as a minor version update.   

We'd appreciate your help to add tests and make fixes.  That will do a better job addressing the problem than simply locking to a specific version.    You don't have to "hope for the two issues mentioned above to get addressed" - you've already done bisects spotted where the change occurred, you're quite a ways toward implementing the tests and fixes yourself.

We'd be happy to discuss taking code samples from this codebase as baselines to check for changes, too.

gagern:
@bitwiseman wrote:
> We'd appreciate your help to add tests and make fixes. That will do a better job addressing the problem than simply locking to a specific version.

While I'll be happy to write test cases and file issues whenever js-beautify doesn't do as good a job as I'd expect it to, I have to do so at my own time. When I decide “now I want to bump the js-beautify version”, then I've got the time to carefully look at all the changes and handle them appropriately. But when I'm busy developing something else, then a change in js-beautify behavior, be it for better or worse, is simply a major distraction. As is the case now, where [Travis checks are failing](https://travis-ci.org/CindyJS/CindyJS/jobs/187511599) simply because the latest version of js-beautify doesn't do things exactly the way previous versions did.

A locked version communicates that “this codebase follows the style of js-beautify version foo”, and that relationship is enforced by our Travis checks. If a more recent version of js-hint formats our code the same way, there is no reason to use the more recent version. If a more recent version formats the code differently, that change in style should be made explicit, and the newer style should be made mandatory for everyone dealing with the codebase from that point onwards. So I still don't see a reason not to lock the version, for our specific use case.

> You don't have to "hope for the two issues mentioned above to get addressed" - you've already done bisects spotted where the change occurred, you're quite a ways toward implementing the tests and fixes yourself.

I haven't even started looking at how the code actually works. Seeing how people who know the codebase accidentially break something by fixing something else, I'm not too keen on just plunging ahead with proposed fixes that might have even more severe unintended consequences. If js-beautify doesn't get fixed by someone else, I'll have to make a decision whether to get more involved with its codebase or whether to switch this project here to some other tool.

> We'd be happy to discuss taking code samples from this codebase as baselines to check for changes, too.

The examples I posted in my issue reports were adapted from code I found in our codebase.

Feel free to add a test which checks out our codebase and runs the latest version of js-beautify on it. If that results in any difference, decide whether the change is accidential or intentional. In the former case, you can fix it before your next release. In the latter case, we'd be happy to receive a pull request asking to adjust our code and bump the js-beautify version number. But I guess you shouldn't make that a part of some automated test suite to be performed on each commit. Merely a suggested step prior to a release. The steps to do such a test would be

```sh
git clone https://github.com/CindyJS/CindyJS.git
cd CindyJS
npm install
npm install js-beautify@‹latest commit›
node make beautified
```

Of course you can come up with a way to re-use an existing clone instead of getting a new one for every build. In that case you'll have to do a `git checkout -- .` to clean up possible changes introduced by js-beautify. You might also want to symlink the `js-beautify` directory instead of doing an `npm install` for it, so that you can test a version before pushing it to GitHub. I think you get the idea.

gagern:
I've merged this since the current state of affairs blocks development. Feel free to continue discussion and/or suggest alternatives. I don't want to enforce my solution as the final one, but it's a quick and minimally invasive step forward.

bitwiseman:
@gagern - Well, now that you've merged this, you can update to 1.6.8 as it fixes the issues you filed.

gagern:
@bitwiseman see #558 for that.

PR #559, opened 2017-01-04
How to release

gagern:
Backporting features and bug fixes, and tagging the resulting commits in the right order, is an important part of the release process. I've now documented that aspect, and followed the updated procedure for the 0.8.3 and 0.7.7 releases I just made. The part for backporting bug fixes by cherry-picking them is still untested, as we won't do patch-level releases before we reach 1.0.0.

Please review whether the wording is clear, but feel free to also suggest changes to the procedure or additional detail in one of the points.


strobelm:
Thanks a lot!

PR #563, opened 2017-01-09
Restore renderBody function for manual

### LINKED ISSUE #561: md2html.js - page.then is not a function
https://github.com/CindyJS/CindyJS/blob/v0.8.3/ref/js/md2html.js#L455 cannot be executed since the function `then` has not been defined on `Page`. We need this for the current built process of the website.

gagern:
Fixes #561. As a drive-by fix, also addresses issues where errors loading node-sass don't cent reported properly.


montaga:
Thank you.
With this building the website works again :-)

PR #565, opened 2017-01-11
Update KaTeX to v0.7.0 plus canvas rendering

gagern:
KaTeX has made a new release recently, and I think this is a good point to update our dependency as well. Many of the things we had to customize before are now supported officially. Two things remain missing, though: render to canvas and support for the degree symbol entered as °. So I merged in the former and rewrote the latter inside the katex plugin.

The files used here originate from [a merge between `v0.7.0^` and canvas](https://github.com/gagern/KaTeX/commits/d59fb1607ea47a70221fa238adf3e725665a).


strobelm:
I checked the TeX examples by hand, looks fine. Thanks Martin!

PR #569, opened 2017-01-19
Add fillrule to polygondrawing

strobelm:
This adds a modifier for the canvas fill rule. The second commit changes the default behaviour from evenodd to nonzero which would restore compatibility to Cinderella, but this breaks compatibility to previous releases (at least we had a mozfillrule set to evenodd). Of course one could only cherry pick the first. 

I won't be able to make address changes for the next few days, so feel free to modify my code as you wish.

### DIFF
--- a/examples/PolyEvenOdd.html
+++ b/examples/PolyEvenOdd.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Construction.cdy</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+
+        #CSConsole {
+            background-color: #FAFAFA;
+            border-top: 1px solid #333333;
+            bottom: 0px;
+            height: 200px;
+            overflow-y: scroll;
+            position: fixed;
+            width: 100%;
+        }
+    </style>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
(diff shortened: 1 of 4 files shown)


gagern:
For the record: I just had a look at the remaining calls to `csctx.fill`, checking that the choice of fill rule makes no difference for these:

* [`eval_helper.drawarc`](https://github.com/strobelm/CindyJS/blob/78f0304c33373bb0ece1f4e711579dcec5f5f03a/src/js/libcs/OpDrawing.js#L218)
* [`eval_helper.drawcircle`](https://github.com/strobelm/CindyJS/blob/78f0304c33373bb0ece1f4e711579dcec5f5f03a/src/js/libcs/OpDrawing.js#L287)
* [`draw_arrowhead`](https://github.com/strobelm/CindyJS/blob/78f0304c33373bb0ece1f4e711579dcec5f5f03a/src/js/libcs/Render2D.js#L544)
* [`drawpoint`](https://github.com/strobelm/CindyJS/blob/78f0304c33373bb0ece1f4e711579dcec5f5f03a/src/js/libcs/Render2D.js#L563)

> The second commit changes the default behaviour from evenodd to nonzero which would restore compatibility to Cinderella, but this breaks compatibility to previous releases (at least we had a mozfillrule set to evenodd).

We don't have the old behavior documented. We do have Cinderella compatibility as our declared goal. The old support was specific to Mozilla anyway. I very much doubt anything relies on it. So I'd say we can include this in 0.8.4 without worrying too much. But thanks for pointing this out nonetheless.

PR #571, opened 2017-01-25
Textures for Cindy3D surfaces

gagern:
At the moment this is a very limited implementation of surface textures for Cindy3D. There can only be one texture in the whole scene, and that texture will get applied to all meshes, so it can't even be combined with meshes that don't have a texture. Nevertheless, this is better than nothing, and as long as people using this fix the CindyJS version they use (i.e. all the parts of the version number), they should be fine.

[Triceratops.zip](https://github.com/CindyJS/CindyJS/files/731139/Triceratops.zip) is a demo of this. It will likely need its own web server and won't run from the local file system due to picture access. I guess once this gets merged, I'll put that or something like it on the web page.


examples/cindy3d/18_TexTorus.html
```diff
@@ -0,0 +1,44 @@
...
+end3d();
+</script>
+<script type="text/javascript">
+createCindy({canvasname:"CSCanvas",scripts:"cs*",images:{tex1:"tex1.png"}});
+</script>
+</head>
+
+<body>
+  <canvas id="Cindy3D" style="border: none;" width="632" height="452"></canvas>
+  <canvas id="CSCanvas" width="50" height="50" style="border:none"></canvas>
```
montaga: this should be `<div id="CSCanvas" width="50" height="50" style="border:none"></div>` to satisfy travis.

plugins/cindy3d/src/js/Ops3D.js
```diff
@@ -362,17 +362,40 @@ CindyJS.registerPlugin(1, "Cindy3D", function(api) {
...
                        coerce.toString(a, colortype).toLowerCase()),
         "topology": (a => topology =
-                       coerce.toString(a, normaltype).toLowerCase()),
-        "colors": (a => colors = coerce.toList(a).map(elt => coerce.toColor(elt))),
+                       coerce.toString(a, topology).toLowerCase()),
+        "colors": (a => colors = coerce.toList(a).map(
+          elt => coerce.toColor(elt))),
+        "uv": (a => uv = coerce.toList(a).map(
+          elt => coerce.toHomog(elt, [0, 0, 0], 2))),
+        "texture": (a => texture = coerce.toString(a)),
```
montaga: With this it is only possible to have textures that are refered as string. But not `cameravideo()`, for instance.
But that could be considered as an additional feature.

gagern:
I completely forgot the Travis complaints on this one. Supporting non-string texture arguments would definitely be nice to have, but I'll need some rescaling work to handle image dimensions which are not a power of two in these cases. Better do that in a follow-up.

PR #572, opened 2017-01-30
Touch support for Cindy3D

### LINKED ISSUE #39: Cindy3D should support touch events
The Cindy3D plugin [doesn't handle touch events](https://github.com/CindyJS/CindyJS/blob/8bcfbab5945f20f4b18e3b8f0c03a37411baeacf/src/js/cindy3d/Viewer.js#L221-L272) yet. A look at [`Event.js`](https://github.com/CindyJS/CindyJS/blob/8cfaaedbb611a910e1baab75b9c0f7b838ebe617/src/js/Events.js#L178-L208) should help get this up and running.

gagern:
Fixes #39.

This adds touch event support for Cindy3D. It also increases floating-point precision, which improves rendering of many elements on some mobile devices.

@richter-gebert can you have a look how this works for you? The iPads I've been testing this on still don't support EXT_frag_depth, according to http://webglreport.com/, which breaks some examples, in particular all those with the torus knot. But apart from that, most examples look reasonably decent now, and can be interacted with in what I consider an intuitive way.

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -283,6 +283,7 @@ module.exports = function build(settings, task) {
         "Camera",
         "Appearance",
         "Viewer",
+        "Controls",
         "Lighting",
         "PrimitiveRenderer",
         "Spheres",
--- a/make/commands.js
+++ b/make/commands.js
@@ -150,6 +150,7 @@ exports.closureCompiler = function(jar, opts) {
         }
         if ([
             "output_wrapper_file",
+            "externs",
         ].indexOf(key) !== -1)
             this.input(val);
         if ([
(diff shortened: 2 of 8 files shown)


gagern:
@kranich my experience shows that pretty much all examples look seriously broken on mobile devices with mediump. So I wonder what's better: showing this brokenness or not showing anything. I believe it's easier for users to associate the latter with lacking device support, while the former will be seen as a problem with the library. Not sure.

kranich:
@gagern OK, if it's deliberate, please proceed.

PR #573, opened 2017-01-30
Adjust specular reflectiveness based on shininess

gagern:
This prevents dull objects from looking extremely overexposed. Look at the example to see the effect of these changes here:

Before

After

@richter-gebert do you think we need more flexible manual control, or can we use this automatic coupling between shininess and specular reflectiveness and avoid burdening users with yet another parameter to tweak?


strobelm:
Looks way better than before. If we need more flexibility for tuning parameters we can add them later. Thanks @gagern!

PR #577, opened 2017-02-06
labeled getter and setter

### LINKED ISSUE #576: Labels of geometric points
Firstly, labels of geometric points cannot be turned on or off within scripts; neither for points created as part of the geometry in createCindy() nor of points created by createpoint(). I.e. the commands
`A.labelled = boolean`
and
`A.labeled = boolean`
don't work.

It is, however, possible to (de-)activate the label of a point within the definition in createCindy(). But, secondly, one has to use the command
`labeled: boolean`
whereas the CindyJS manual suggests to use
`A.labelled = boolean`
in the scripts.

So, first, it should be possible to change whether the label is shown. Second, the commands should be consistent throughout CindyJS. As "labeled" is the American English version, it should be the "correct" one.

strobelm:
This should resolve #576. I'm admitting both AE and BE diction.

### DIFF
--- a/examples/labeled.html
+++ b/examples/labeled.html
@@ -6,38 +6,27 @@
 <meta charset="UTF-8">
 <link rel="stylesheet" href="../build/js/CindyJS.css">
 <script type="text/javascript" src="../build/js/Cindy.js"></script>
-<script id="csinit" type="text/x-cindyscript">
 
-// Initialization code, executed once up front.
-
-</script>
 <script id="csdraw" type="text/x-cindyscript">
 fillcircle((0,0),5, color->(1,1,0), alpha->0.2);
 drawtext((0,1.5), "labeled");
 drawtext((0,-6), "unlabeled");
-drawtext((-8,-8), "labeled:" + A.labelled);
-if(|A.xy| < 5,  A.labelled = true, A.labelled = false);
-// Drawing code, executed whenever the canvas gets redrawn.
-
+drawtext((-8,-8), "labeled:" + A.labeled);
+if(|A.xy| < 5,  A.labeled = true, A.labelled = false);
 </script>
+
 <script type="text/javascript">
 
-var cdy = CindyJS({ // See ref/createCindy documentation for details.
+var cdy = CindyJS({   
   ports: [{id: "CSCanvas", width: 500, height: 500}],
   scripts: "cs*",
   language: "en",
   defaultAppearance: {
-    // See GeoBasics.js for possible attributes.
   },
   geometry: [
     {name:"A", type:"Free", pos:[0,0], labeled:true}
-    // For allowed types, see GeoOps.js.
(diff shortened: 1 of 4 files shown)


src/js/libcs/Accessors.js
```diff
@@ -219,6 +221,11 @@ Accessor.setField = function(geo, field, value) {
             geo.pinned = value.value;
         }
     }
+    if (field === "labeled" || field === "labelled") {
```
gagern: If you have the name mapping in `generalFields`, using a single spelling variant here should suffice, right?

gagern:
To fully address #576, this PR should change the documentation to use AE spelling for the sake of consistency.

src/js/libcs/Accessors.js
strobelm: If i'm reading the code correctly this only applies to the getter code. We never defined something like `if (Accessor.generalFields[field])` for the setter function which could use the mapping.

strobelm:
I added an example and updated the documentation.

examples/labeled.html
```diff
@@ -0,0 +1,48 @@
...
+    // See GeoBasics.js for possible attributes.
+  },
+  geometry: [
+    {name:"A", type:"Free", pos:[0,0], labeled:true}
+    // For allowed types, see GeoOps.js.
+    // For allowed properties, see csinit in GeoBasics.js.
+  ] // End of geometry array.
+});
+
+// Remove all comments after adjusting this template for your use case.
```
gagern: Please clean comments.

examples/labeled.html
```diff
@@ -0,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
+<link rel="stylesheet" href="../build/js/CindyJS.css">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
```
gagern: I'd drop this unused script.

examples/labeled.html
```diff
@@ -0,0 +1,48 @@
...
+
+// Initialization code, executed once up front.
+
+</script>
+<script id="csdraw" type="text/x-cindyscript">
+fillcircle((0,0),5, color->(1,1,0), alpha->0.2);
+drawtext((0,1.5), "labeled");
+drawtext((0,-6), "unlabeled");
+drawtext((-8,-8), "labeled:" + A.labelled);
+if(|A.xy| < 5,  A.labelled = true, A.labelled = false);
```
gagern: Perhaps we should mix spellings, to ensure both work.

src/js/libcs/Accessors.js
gagern: Perhaps we should reconsider that approach one day, but that would be a separate PR.

strobelm:
This should address all your remaining points.

PR #578, opened 2017-02-06
Cindy3D: Allow supplying textures by object not name, allow NPOT textures

gagern:
Now a texture can be given using any CindyJS-supported image format, not only using a string by name.  There is also support for textures which are not a power of two in size, rescaling them to make them power of two.  These two combine nicely to allow e.g. a live camera video as texture.

See https://github.com/CindyJS/CindyJS/pull/571#pullrequestreview-19546447 for previous discussion about this.

At the moment there is no fast-track approach to use CindyGL output as a texture without going via a HTMLCanvasElement. Such a shortcut would certainly be feasible, but probably should be achieved by making the corresponding property part of the externs specification, to allow for better interoperability with future extensions. Probably best handled in a subsequent PR.

### DIFF
--- a/plugins/cindy3d/src/js/Triangles.js
+++ b/plugins/cindy3d/src/js/Triangles.js
@@ -156,14 +156,26 @@ Triangles.prototype.texture = null;
 /** @type {?WebGLTexture} */
 Triangles.prototype.textureObj = null;
 
+/** @type {?HTMLCanvasElement} */
+Triangles.prototype.textureScaler = null;
+
 /**
  * @param {Viewer} viewer
  */
 Triangles.prototype.render = function(viewer) {
   const gl = viewer.gl;
   if (this.texture) {
+    if (!this.texture.ready)
+      return;
+    if (this.texture.cdyUpdate)
+      this.texture.cdyUpdate();
     if (this.textureObj === null) {
       this.textureObj = gl.createTexture();
+    }
+    let w = this.texture.width, h = this.texture.height;
+    if (true) {
+      w = nextPowerOfTwo(w);
+      h = nextPowerOfTwo(h);
     }
     gl.activeTexture(gl.TEXTURE0);
     gl.bindTexture(gl.TEXTURE_2D, this.textureObj);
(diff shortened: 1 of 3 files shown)


plugins/cindy3d/src/js/Triangles.js
```diff
@@ -173,8 +185,22 @@ Triangles.prototype.render = function(viewer) {
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER,
                      gl.LINEAR_MIPMAP_LINEAR);
     gl.hint(gl.GENERATE_MIPMAP_HINT, gl.NICEST);
-    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE,
-                  this.texture.img);
+    let /** @type {HTMLImageElement|HTMLCanvasElement|HTMLVideoElement} */ img
+        = this.texture.img;
+    if (w !== this.texture.width || h !== this.texture.height) {
```
montaga: Probably it makes sense to have a
`if (this.texture.live || (this.img.generation < this.texture.generation)) {
this.img.generation = this.texture.generation;
...
`  here to avoid multiple rescalings of the same image.

plugins/cindy3d/src/js/Triangles.js
```diff
@@ -173,8 +185,22 @@ Triangles.prototype.render = function(viewer) {
...
+      if (!this.textureScaler) {
+        this.textureScaler = /** @type {HTMLCanvasElement} */
+          (document.createElement("canvas"));
+        this.textureScaler.width = w;
+        this.textureScaler.height = h;
+      }
+      let ctx = this.textureScaler.getContext("2d");
+      ctx.clearRect(0, 0, w, h);
+      ctx.drawImage(this.texture.img, 0, 0, w, h);
```
montaga: Scaling artifacts might occur with this approach. Probably one gets better results if we fill only one corner of img and modify the texture accesses (This has been done in CindyGL). But as long as we do not have an example, where these artifacts become visible, I guess it is okay.

gagern: Using only a partial image will likely lead to artifacts with wrap-around, particularly in the presence of mipmaps. I've read that it's actually possible to use non-power-of-two textures in WebGL if one doesn't use wraparound and doesn't use mipmaps, so I guess that internally this does essentially what you just suggedted.

PR #580, opened 2017-02-07
Compose transformations

gagern:
This allows for the composition of transformations.

Instead of a visual example there is a sizable test suite to ensure that transforming a point by the transformations one after the other has the same effect as applying the combined transformation.  The test framework was also extended to verify some basic properties of Möbius transformations.

### DIFF
--- a/src/js/libgeo/GeoOps.js
+++ b/src/js/libgeo/GeoOps.js
@@ -2589,6 +2589,107 @@ geoOps.TransformPolygon.updatePosition = function(el) {
     }));
 };
 
+geoOps.TrComposeTrTr = {};
+geoOps.TrComposeTrTr.kind = "Tr";
+geoOps.TrComposeTrTr.signature = ["Tr", "Tr"];
+geoOps.TrComposeTrTr.updatePosition = function(el) {
+    var a = csgeo.csnames[el.args[0]];
+    var b = csgeo.csnames[el.args[1]];
+    el.matrix = List.normalizeMax(List.productMM(b.matrix, a.matrix));
+    el.dualMatrix = List.normalizeMax(List.productMM(b.dualMatrix, a.dualMatrix));
+};
+
+geoOps._helper.composeMtMt = function(el, m, n) {
+    var add = CSNumber.add;
+    var sub = CSNumber.sub;
+    var mult = CSNumber.mult;
+
+    function f1(a, b, c, d) { // a*b + c*d
+        return add(mult(a, b), mult(c, d));
+    }
+
+    function f2(a, b, c, d, e, f, g, h) {
+        return add(f1(a, b, c, d), f1(e, f, g, h));
+    }
+
+    function f3(a, b, c, d, e, f, g, h) {
+        return sub(f1(a, b, c, d), f1(e, f, g, h));
+    }
+
+    var addsub = n.anti ? f3 : f2;
+    var subadd = n.anti ? f2 : f3;
+    var v = List.normalizeMax(List.turnIntoCSList([
+        subadd(m.ar, n.ar, m.cr, n.br, m.ai, n.ai, m.ci, n.bi),
+        addsub(m.ar, n.ai, m.cr, n.bi, m.ai, n.ar, m.ci, n.br),
+        subadd(m.br, n.ar, m.dr, n.br, m.bi, n.ai, m.di, n.bi),
+        addsub(m.br, n.ai, m.dr, n.bi, m.bi, n.ar, m.di, n.br),
+        subadd(m.ar, n.cr, m.cr, n.dr, m.ai, n.ci, m.ci, n.di),
+        addsub(m.ar, n.ci, m.cr, n.di, m.ai, n.cr, m.ci, n.dr),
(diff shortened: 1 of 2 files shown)


src/js/libgeo/GeoOps.js
```diff
@@ -2353,6 +2359,9 @@ geoOps.TrAffine.updatePosition = function(el) {
 geoOps.TrSimilarity = {};
 geoOps.TrSimilarity.kind = "Tr";
 geoOps.TrSimilarity.signature = ["P", "P", "P", "P"];
+geoOps.TrSimilarity.initialize = function(el) {
+    el.isEuclidean = 1;
```
strobelm: Please correct me if i'm wrong but similarity transformations should be non-euclidean?

src/js/libgeo/GeoOps.js
```diff
@@ -2426,6 +2441,9 @@ geoOps.TrReflectionP.updatePosition = function(el) {
 geoOps.TrReflectionL = {};
 geoOps.TrReflectionL.kind = "Tr";
 geoOps.TrReflectionL.signature = ["L"];
+geoOps.TrReflectionL.initialize = function(el) {
+    el.isEuclidean = -1;
```
strobelm: A property which is called "is*" i would have expected something of type bool. Not a show-stopper.

src/js/libgeo/GeoOps.js
```diff
@@ -2589,6 +2589,107 @@ geoOps.TransformPolygon.updatePosition = function(el) {
...
+geoOps.TrComposeTrTr.kind = "Tr";
+geoOps.TrComposeTrTr.signature = ["Tr", "Tr"];
+geoOps.TrComposeTrTr.updatePosition = function(el) {
+    var a = csgeo.csnames[el.args[0]];
+    var b = csgeo.csnames[el.args[1]];
+    el.matrix = List.normalizeMax(List.productMM(b.matrix, a.matrix));
+    el.dualMatrix = List.normalizeMax(List.productMM(b.dualMatrix, a.dualMatrix));
+};
+
+geoOps._helper.composeMtMt = function(el, m, n) {
```
strobelm: I don't really get what's going on here.

src/js/libgeo/GeoOps.js
gagern: The term `Euclidean` is probably inappropriate here. Essentially this is referring to a transformation which fixes the ideal circle points I and J. That's essentially the intersection between generalized Möbius transformations and projective transformations. The term comes from Cinderella terminology. Would `isSimilarity` be easier to understand?

src/js/libgeo/GeoOps.js
gagern: True. The problem this abuse of the field is trying to achieve is that we have both orientation-preserving and orientation-reversing operations. And while it's easy enough to distinguish these on real coordinates, that distinction is far harder along complex detours. On the other hand, the distinction is evident in the construction sequence, just as the classification as a similarity is evident there. So that's why I'm abusing the flag in this way. `similarityDirection` or `similaritySign` or something like this might be alternatives, although they would be more confusing in contexts where the sign does not matter and the boolean coercion is really all that matters. Do you or someone else have any better names to suggest?

src/js/libgeo/GeoOps.js
gagern: Let's see. The best way to think about this is probably in terms of matrices `[[x, -y], [y, x]]` representing our doubly complex numbers. In this setup, a Möbius transformation is a 4×4 matrix. I computed the product of these, and tried to capture the formulas for the individual components in a short formula. There is also the matter of conjugation: if the second operation (left factor, `n`) is an anti-Möbius transformation, then the first transformation (right factor, `m`) has to be conjugated (replacing all `y` by `-y` in the doubly complex representation) before the matrices are multiplied. That's what the `n.anti ? … : …` constructs are all about. `addsub` does add in the direct and subtract in the reversing situation, for `subadd` it's the other way round. Then I collect all those coordinates, normalize them together before constructing a Möbius transformation from them. Does this help? How can this be made clearer, short of pasting the whole output from Sage as a comment to explain the matrix multiplication?

src/js/libgeo/GeoOps.js
strobelm: Ah okay! isSimilarity would be great, a comment about the property would be also fine.

src/js/libgeo/GeoOps.js
strobelm: This helps indeed! We should find a way to document the code generated by computer algebra systems. Let's assume github vanishes (unlikely i know, but far from impossible) at some point and we only have the source code left. I don't know if we ever could reconstruct what's the idea behind these lines. Probably the best place would be a comment in the code itself (imo).

gagern: I don't know. It depends on the level of detail. A comment like “extracted coefficients from CAS computation of matrix product” might be all right, but adds pretty little detail. Otherwise I think that the full computation will require too much space to include in a comment. I'd rather save the Sage Jupyter notebook somewhere, and/or the HTML export from that. We could use the code repository for this, or the website, or a dedicated computations repository.

strobelm:
The code is fine. We should discuss documenting CAS generated code in the next meeting.

PR #584, opened 2017-02-24
Convex hull 

strobelm:
Hi this is the new code for convexhull3D.

The code itself works fine (tanks to @nasay), but i could not persuade the tests to work. Any comment on that would be very appreciated.  

This supersedes #360.


strobelm:
From our side the branch should be ready for a merge. It would be great if @gagern could have a quick look if we terribly broke something.

PR #586, opened 2017-02-27
Add Plugin for basic symbolic computations

montaga:
This adds a plugin which implements basic symbolic operations that are needed for the raytracer.

It provides a function `symdiff`. This operator computes the symbolic differentiation of a function and inserts its symbolic expression as # in the third argument, which then will be evaluated. Example usages are
`symdiff(f(x), x, df(x):=#);` or `symdiff(x^5+y+z, x, dxf(x,y,z):=#);`
It is needed for the Raytracer because here we want to compile the derivatives of a function to GLSL Code.

Furthermore, the plugin adds a `degree` function, which computes the degree of a polynomial. The variables that should be considered are given as (up to 3) consecutive arguments. An example usage is
`degree(x*x, x) == 2;`, `degree(x^5*y^2+x*a^10, x, y) == 7;`, `degree(sin(x), x) == -1;`. `-1` is returned for non-polynomial functions.

So far, both operators only support basic arithmetic operations, hence all algebraic expressions, and `sin` and `cos`. I think, as it is not part of the core-CindyJS, it is fine to extend this plugin to more operators only when needed.


strobelm:
A great first step! Thanks Aaron.

PR #595, opened 2017-04-03
Add `readpixels` functions. (+RK4-molecule simulation in CindyGL)

montaga:
For more advanced GPGPU programming one needs to be able to read back big amounts of data from the GPU to CPU. So far, single pixels could only be read with `imagergb[a]`, but this is to slow if one wants to transfer a lot of data. So I have introduced a `readpixels` function which returns a long array containing all the pixel data of a image.
This function has been approved by @richter-gebert . Maybe the name of the so far undocumented function will change at a later point. 

As an running GPGPU example that also uses this function a N-body (molecule) simulation has been added that runs its computations on the GPU.
Also 2 minor bugfixes for CindyGL are included within this PR.

### DIFF
--- a/examples/cindygl/48_molecules.html
+++ b/examples/cindygl/48_molecules.html
@@ -1,0 +1,333 @@
+
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Runge-Kutta</title>
+<meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


strobelm:
Looks good, thanks @montaga

PR #597, opened 2017-04-27
compatibility with Cinderella handling of drawtext modifier

kortenkamp:
In Cinderella it is possible to align text using the align->“center“ modifier. CindyJS so far only handles align->“mid“. This commit adds „center“.


strobelm:
Looks fine, thanks @kortenkamp!

PR #602, opened 2017-05-04
Kortenkamp/new example football

kortenkamp:
This adds an example that does mostly text drawing and calculations. I used it to verify the fix for #597 .


montaga:
To me this looks fine :+1:

PR #604, opened 2017-06-06
snap to grid for point on line and point on segment

strobelm:
This PR adds snap to grid for point on line and point on segment.


src/js/libgeo/GeoOps.js
```diff
@@ -3060,6 +3068,25 @@ geoOps.Poly.updatePosition = function(el) {
...
+    var projPos = geoOps._helper.projectPointToLine(pos, line);
+    projPos = List.normalizeZ(projPos);
+
+    var sx = projPos.value[0].value.real;
+    var sy = projPos.value[1].value.real;
+    var rx = Math.round(sx / csgridsize) * csgridsize;
+    var ry = Math.round(sy / csgridsize) * csgridsize;
+    var newpos = List.realVector([rx, ry, 1]);
+    if (Math.abs(rx - sx) < 0.2 && Math.abs(ry - sy) < 0.2 &&
+        CSNumber.abs(List.scalproduct(line, newpos)).value.real < CSNumber.eps) {
```
montaga: Are there any cases when `List.scalproduct(line, newpos)` becomes complex (for example if line vanishes in the complex)? Maybe it is better to use
`CSNumber.isAlmostZero ` instead.

src/js/libgeo/GeoOps.js
```diff
@@ -3060,6 +3068,25 @@ geoOps.Poly.updatePosition = function(el) {
...
+        pos = newpos;
```
montaga: A feeling tells me that `pos = geoOps._helper.projectPointToLine(newpos, line);` might be better here for the stability of consecutive operations.

kortenkamp:
Did anybody test whether this breaks tracing?

strobelm:
Thanks for the remarks @montaga! 
@kortenkamp most likely this should not mess with tracing. `pos` is handed over to tracing and indicates that we moved the mouse to `pos` and then tracing takes over moving the point according to tracing logic of point on line. The only difference is now that we manipulate the mouse position in order to snap to grid points. Especially we are never setting the homogeneous coordinates of the element directly.

PR #605, opened 2017-06-13
fix travis env for c++11

strobelm:
Hi, according to https://github.com/travis-ci/travis-ci/issues/4771#issuecomment-241439984 this should fix our travis problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,15 @@ node_js:
   - "0.12"
 script: node make build=release alltests
 sudo: false
+env:
+  - CXX=g++-4.8
 addons:
   apt:
+    sources:
+     - ubuntu-toolchain-r-test
     packages:
      - default-jdk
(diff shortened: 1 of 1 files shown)


kranich:
libsass compilation takes ages! Maybe we should use [less](https://www.npmjs.com/package/less) instead?

strobelm:
superseded by #607

gagern:
As for sass vs. less: I prefer the sass language slightly, but the feeling is not strong enough to rule out less.

In what situations would you encounter rebuilds? I see them for Travis builds against latest version of Node, but that's mainly because we are using an outdated version of libsass. Instead of movinbg to less, I'd just bump the sass module so we can use a binary release there.

kranich:
Yes, for Travis builds with latest Node. If a newer sass version speeds up the build process significantly, I would appreciate that very much.

PR #606, opened 2017-06-17
Use latest version of Complex Curves (a64babd)

kranich:
Equation must be bivariate and at least quadratic


kranich:
Can be merged after #605.

PR #608, opened 2017-06-18
Bump node-sass version to 4.5.3

kranich:
As suggested by @gagern in #605.

It is simply not very motivating when the build fails due to a compilation error during the sass build, after I have waited for over 10 minutes. :-(


kranich:
Build time down to 1:36 min, much better! :-)

PR #612, opened 2017-07-06
Automatic use webgl 2.0 in Cindy3D if EXT_frag_depth not available

montaga:
With this, WebGL 2.0 is used in Cindy3D if EXT_frag_depth not available. This hopefully should solve the rendering issues on iPads, because it seems that iOS 11 will support WebGL 2.0, which automatically supports frag_depth. However, we could not test it on iPads yet.

At this time, switching to WebGL 2.0 whenever available, seems not good as rendering problems become visible in "examples/cindy3d/01_ThreeSpheres.html" in Firefox and WebGL 2.0.

Thank you, @strobelm for helping with the first prototype :+1:

### DIFF
--- a/plugins/cindy3d/src/js/ShaderProgram.js
+++ b/plugins/cindy3d/src/js/ShaderProgram.js
@@ -25,6 +25,17 @@ GlError.prototype.toString = function() {
  */
 function ShaderProgram(gl, vertexShaderCode, fragmentShaderCode) {
   this.handle = gl.createProgram();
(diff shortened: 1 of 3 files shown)


montaga:
We just have tested it and it also enhances the results on current iPads with iOS 10 also significantly :) However, still some little rendering artifacts, maybe because of a lower precision, are visible

strobelm:
Works well on multiple devices and is definitely a major improvement!

PR #613, opened 2017-07-14
Separate build of Complex Curves library from build of Complex Curves plugin

kranich:
I have been experimenting with [gulp](http://gulpjs.com/) for building Complex Curves. In order to facilitate a transition, I would like to build the Complex Curves library in plugins/ComplexCurves/lib/ComplexCurves and the plugin itself independently.

@gagern The build system you have written for CindyJS seems to be very similar to gulp. In your opinion, would it make sense to use gulp for building CindyJS instead of our custom build system? I think this might be more friendly towards contributors or plugin authors, who may already know gulp. (Besides, we already use gulp for the website.)


kranich:
@gagern Please have a look if it is OK to build library and plugin separately.

Do you have any objections against using gulp for building the library in the near future?

kranich:
@strobelm Can you please give your assent for merging this? As far as I remember, you had no objections when we talked about it lately.

strobelm:
Works fine for me. Merged.

kranich:
@strobelm Thanks!

Issue #476, opened 2016-09-16
Automatically support colorplot via CindyGL


gagern:
As far as I understand it, [the `colorplot$3` function implemented by CindyGL](https://github.com/CindyJS/CindyJS/blob/v0.7.3/plugins/cindygl/src/js/Plugin.js#L68-L71) should serve as a reasonably compatible replacement for the corresponding function built into Cinderella. So I suggest we [auto-load the plugin](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/Setup.js#L506-L509) if we detect any use of this function. That would make widgets exported from Cinderella more likely to work out of the box.

Also make sure to remove [the _not available yet_ line](https://github.com/CindyJS/CindyJS/blob/v0.7.3/ref/Function_Plotting.md#creating-a-colorplot-colorplotexprvecvec) in the reference manual. Speaking of reference manual, it would be nice to also see a description of the other functions added by CindyGL, `colorplot$1` in particular. At least I think that the fact that the support comes from a plugin is no reason against documenting it as long as we ship the plugin.

If you are worried about constructs which are not supported by CindyGL, or perhaps about browsers with insufficient WebGL capabilities, we can still open another ticket requesting a fallback to CPU-based colorplot for those situations.

kortenkamp:
I agree. Can we implement auto-load by just putting in a stub function that will load the plugin and replace itself?

gagern:
@kortenkamp: We might, but the loading would have to happen asynchronously, which means the first few function invocations might get lost. I think we have a superior machinery in place, one which ensures that the plugin gets loaded before the widget launces. Note that “[auto-load the plugin](https://github.com/CindyJS/CindyJS/blob/v0.7.3/src/js/Setup.js#L506-L509)” was a link in my post, referencing the auto-load machinery we already have in place for quickhull3d. The only situation where that would fail is if the function invocation were part of some button script code or dynamically evaluated expression string.

kortenkamp:
Ok, I understand. And I forgot that this is working already, so, yes, go ahead, I'd say.

kranich:
Fixed by #616.

PR #616, opened 2017-08-18
autoload CindyGL if colorplot is called

montaga:
The CindyGL-plugin will be automatically loaded whenever the function colorplot is called. That makes is possible to export colorplots from Cinderealla out of the box, (as far as CindyGL can transcompile the CindyScript expressions within colorplot to GLSL.)


montaga:
With this also #476 should be fixed :-)

strobelm:
Works great! Thanks!

PR #620, opened 2017-08-31
ignore clearimage on undefined image

montaga:
Without this `examples/88_Vectorfield.html` fails because here clearimage is called on an undefined image. One should not do this intentionally, however, if one does it, it should be ignored instead of crashing CindyJS.


strobelm:
Great, thanks @montaga. Could you beautify the code?

PR #622, opened 2017-08-31
added webcam support for Safari 11

montaga:
Safari 11 did not support
`var url = window.URL.createObjectURL(stream);`. The alternative `video.srcObject = stream;` also works on Firefox and Chromium. Furthermore some lines were added to make Safari accepting to play the stream.

### DIFF
--- a/src/js/libcs/OpImageDrawing.js
+++ b/src/js/libcs/OpImageDrawing.js
@@ -466,8 +466,15 @@ evaluator.cameravideo$0 = function(args, modifs) {
     var img = loadImage(video, true);
     console.log("Opening stream.");
     openVideoStream(function success(stream) {
+        /* does not work in Safari 11.0 (beta)
         var url = window.URL.createObjectURL(stream);
         video.src = url;
+        */
+        video.srcObject = stream;
+        video.setAttribute('autoplay', '');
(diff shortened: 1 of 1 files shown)


strobelm:
Hi if i'm reading this correctly this breaks backwards compatibility (https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices), the old way using createObjectFromURL should be supported by more browsers (https://developer.mozilla.org/de/docs/Web/API/URL/createObjectURL). Could we implement a switch?

montaga:
The entirely new way via `navigator.MediaDevices.getUserMedia()` is not chosen here; we still use the old way via `navigator.getUserMedia()` (and `webkitGetUserMedia`, `mozGetUserMedia`, `msGetUserMedia` as alternative)

Honestly, I do not know how to check whether `video.srcObject` is supported or not supported. In both cases, it will be `null` after initialization. On the other hand, `window.URL.createObjectURL` also exists in Safari, but it gives an error when calling. Trying to produce and catch an error or to determine the specific browser by investigating the string in `navigator.userAgent` seems to me bad programming practice...

strobelm:
Alright then. Code looks good, i found a webcam somewhere and tested the code sucessfully ond FF and Chromium.

PR #625, opened 2017-08-31
Deprecate node 0.12 Travis CI run

### LINKED ISSUE #615: Deprecate support of node 0.12
Is there any reason why we still use node 0.12 in our Travis CI builds? This causes some problems with using gulp for building the Complex Curves plugin. (Some dependencies of gulp require node >= 4.0.0.)

See #614.

kranich:
Run Travis CI with at least node 4.2.6, install node 8.4.0 unless node >=4.2.6 is installed.

This should fix #615.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - node
-  - "0.12"
+  - 4
 script: node make build=release alltests
 dist: trusty
 sudo: false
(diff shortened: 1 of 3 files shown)


kranich:
This could be improved by downloading the latest version of node 8.x from https://nodejs.org/dist/latest-v8.x/. To that end, the filename of the file to be downloaded could be determined from https://nodejs.org/dist/latest-v8.x/SHASUMS256.txt. What do you think?

montaga:
For me auto-downloading the latest node 8.x version does not justify the implication of the extraction of the filename from https://nodejs.org/dist/latest-v8.x/SHASUMS256.txt which is a potential error source. But if anyone feels so, feel free to do so.

PR #630, opened 2017-09-22
Remove e.preventDefault() for keyboard-events

montaga:
Right now, when a CindyJS-applet that has a `keydown`/`keyup`/`keydown`-script is focused, then all keyboard events that are addressed to the browser (for example `CTRL+S`, `⌘+]`) are ignored as long TAB is not involved.
Currently, @standupmaths, who uses CindyJS+a clicker for a show, had some problems because he could not switch different applets with `⌘+]` via a clicker.

Do we need this behavior? Why was it introduced? Or should we filter all requests that involve `⌘`?
This pull request forwards all keyboard events to the browser.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -165,19 +165,13 @@ function setuplisteners(canvas, data) {
         addAutoCleaningEventListener(canvas, "keydown", function(e) {
             if (e.keyCode === 9 /* tab */ ) return;
             cs_keydown(e);
-            if (!cscompiled.keytyped) {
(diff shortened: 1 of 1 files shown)


kortenkamp:
Please check back with @richter-gebert – also please check whether this affects text input fields. I think that a script should receive these events and should also be able to 'consume()' them. Just think also about using ⌘ as a modifier.

So, please be careful. I am currently on the road and don't have the time to check for showstoppers.

strobelm:

PR #634, opened 2017-10-23
Proposal: universal assignment (<<-)

montaga:
DO NOT MERGE until the discussion is finished, because this would fundamentally change some paradigms in CindyJS.

This PR adds a "universal assignment"-operator `<<-` (syntactically alternative suggestions: `@=` or `:-`) which changes the value behind a reference. This operator can be used to enable call-by-reference or to accelerate list operations. Furthermore, such an operator could be used for fast JSON/dict-Objects.

### DIFF
--- a/ref/Variables_and_Functions.md
+++ b/ref/Variables_and_Functions.md
@@ -257,7 +257,49 @@ when called.
     > x*10 + x
     < 45
 
+### Universal assignment of values
+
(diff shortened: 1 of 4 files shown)


montaga:
Another syntactic proposal: `=>`, indicating a reference pointing to some value:
For instance, a swap-function would look like this:
```
swap(a, b) := (
  tmp => b;
  b => a:
  a => tmp;
);

kortenkamp:
IMHO we should avoid `=>`, as it looks like the boolean implication symbol, and also can be confused with `>=`.

strobelm:

PR #639, opened 2018-03-21
: Operator

strobelm:
This adds the ":" to geo ops and lists.

Please have a close look at the change at  ```namespace.getvar```, is there a better way to catch the case where a variable is undefined?

### DIFF
--- a/examples/doublepointop.html
+++ b/examples/doublepointop.html
@@ -12,13 +12,28 @@
 
 </script>
 <script id="csdraw" type="text/x-cindyscript">
+println("=== lists ===");
 lst = [1,2,3];
+lst:"age"=17;
+lst:"list"=lst;
+lst:"pt" = B;
+println(lst:"age");
+println(lst:"list");
+println(lst:"pt");
+println(lst:"age");
+
+println(lst:"undef");
+
+
+println("=== geo ===");
 A:"age"=17;
 A:"list"=lst;
 A:"pt" = B;
 println(A:"age");
 println(A:"list");
 println(A:"pt");
+
+println(A:"undef");
 // Drawing code, executed whenever the canvas gets redrawn.
 
 </script>
--- a/examples/doublepointop.html
+++ b/examples/doublepointop.html
@@ -23,6 +23,7 @@ println(lst:"pt");
 println(lst:"age");
 
 println(lst:"undef");
+println(lst:pii);
 
 
 println("=== geo ===");
--- a/examples/doublepointop.html
+++ b/examples/doublepointop.html
@@ -12,8 +12,13 @@
 
 </script>
 <script id="csdraw" type="text/x-cindyscript">
+lst = [1,2,3];
 A:"age"=17;
+A:"list"=lst;
+A:"pt" = B;
 println(A:"age");
+println(A:"list");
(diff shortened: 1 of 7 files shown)


montaga:
Thank you for implementing this.
One day, this operator needs a documentation! As I tested it, I was surprised that it seems to work also call by value :-) I have to understand this.

There is one problem with using non-real numbers as keys. For example:
```
a = [];
a:(1) = 1;
a:(1+i) = 2;
errc(a:1); //gives 2, but I would expect 1 (and so does Cinderella)
```
Furthermore String in Javascript does not work for objects as we would like it here. For instance, `String({a:32}) == '[object Object]'`, which causes the current implementation to misbehave:

```
a = [];
a:[1,2,3] = 10;
a:[3,4,5] = 34;
errc(a:[1,2,3]); //gives 34, Cinderella gives 10
```

One option to deal with this would be to utilize `Dict.key(x)`, which injectively converts every CindyScript value to a string. However, then 
```
lst = [];
lst:12.3 = 4.56;
println(lst:"12.3");
```
would not work anymore, because `12.3` is number, where `"12.3"` is a string. Honestly I would prefer this behaviour, but Cinderella apperantly considers real-numbers as string. So, in oder to emulate the Cinderella behaviour, one has to convert real-numers (without imaginary part) to Strings containing the number and then use a more sophisticated function like `Dict.key(x)`.

montaga:
`var stack = this.vars[name] || [];` with the consecutive check `stack.length === 0` seems reasonable to me.

strobelm:
Thanks @montaga!
I'm currently moving the evaluation code and i'll address the mentioned points :-).

kortenkamp:
Hey guys,

first of all thank you @strobelm for implementing this operator. I am a bit confused, as I *never* *ever* used the colon operator with an array, but I thought that the selector has to be a string. That is, something like `a:[1,2,3]` is not CindyScript for me, but `a:"1,2,3"` or similar would be.

Can you ask @richter-gebert whether any object should be available as a selector or whether this may be restricted to strings?

kortenkamp:
ah, and I forgot: `:` is definitely not `doubledot`, but `colon`

richter-gebert:
The logic is that *everything* that enters as a value is first converted to a string
via Cinderellas niceprint function. If one does not drastically abuse it,
This is very convenient:

If you do 

a=[];
a.”7”=123;

You get

err(a:(3+4)). —> 123

kortenkamp:
Very good, that clarifies it pretty much!

montaga:
```
a = [1,2,3];
a:"bla" = 23;

errc(a:"bla");

a_2 = 34;
errc(a:"bla");
```
gives
```
 ===> 23
 ===> ___
```

montaga:
Okay, this fixes the issue above :+1: I think, we also need a copy on write for userdata:
```
thing1=[-54,12];
thing1:"color"="red";

thing2=thing1;
thing2:"color"="blue";

errc(thing1:"color"); //should be red, but it gives blue
```

strobelm:
Well it took me quite a while (implicit object copy by reference can be quite annoying) but i think i figured it out. Furthermore there are tests now!

PR #644, opened 2018-03-30
Determine last working marked version

kranich:
Recent build problems could be fixed by reverting to marked 0.3.12. Let us determine which is actually the last working version of marked, so that maybe the problems can be fixed in marked upstream.

Please do not merge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2997,9 +2997,9 @@
       "dev": true
     },
     "marked": {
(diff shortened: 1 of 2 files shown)


kranich:
Local tests have shown that 0.3.12 apparently _is_ the last working version. 0.3.13-0.3.19 do not work.

PR #646, opened 2018-05-18
show labels for segments

montaga:
With this PR, CindyJS will show labels for finite segments that have `labeled: true`.
However, the exact meassures are not synchronized with Cinderella yet.

### DIFF
--- a/src/js/libgeo/GeoRender.js
+++ b/src/js/libgeo/GeoRender.js
@@ -106,11 +106,17 @@ function drawgeoline(el) {
                 [el.startpos, el.endpos], modifs);
             if (el.labeled && !el.tmp) {
                 var lbl = el.printname || el.name || "S";
-                var l = List.cross(el.startpos, el.endpos);
+                var orientedline = List.scalmult(
+                  CSNumber.real(Math.sign(el.startpos.value[2].value.real)*Math.sign(el.endpos.value[2].value.real)),
+                  List.cross(el.startpos, el.endpos)
+                );
+
                 var npos = {
-                    'x': l.value[0].value.real,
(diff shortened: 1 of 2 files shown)


src/js/libgeo/GeoRender.js
```diff
@@ -99,6 +104,27 @@ function drawgeoline(el) {
         if (zz.value.real >= 0) { // finite segment
             evaluator.draw$2(
                 [el.startpos, el.endpos], modifs);
+            if (el.labeled && !el.tmp) {
+                var lbl = el.printname || el.name || "S";
+                var l = List.cross(el.startpos, el.endpos);
```
strobelm: I think the information is already given in ``el.homog``.

src/js/libgeo/GeoRender.js
```diff
@@ -99,6 +104,27 @@ function drawgeoline(el) {
...
+                var npos = {
+                    'x': l.value[0].value.real,
```
strobelm: Mhmmm actually I don't get why this works at all? And shouldn't there be a check for complex values?

strobelm: Forget about the comment concerning the complex numbers, the check is done via ``!List._helper.isAlmostReal(el.homog)``

src/js/libgeo/GeoRender.js
montaga: We need the oriented line here (because the label is supposed to be on the same always side). Due to the normalization, the orientation sometimes got lost using `el.homog` causing jumps of the label. But there was a small bug, because also the points were not oriented as well.

strobelm:
Looks fine to me! Thanks!

PR #649, opened 2018-05-24
Enhance drag of circle by radius

montaga:
With this PR

- the option `narrow` for CircleByRadius is implemented
- jumps are avoided if the size of CircleByRadius is changed by dragging (as it is the case for points)

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -27,23 +27,30 @@ function getmover(mouse) {
                     el.narrow : 20) / sc)
                 continue;
         } else if (el.kind === "C") { //Must be CircleMr
-            var mid = csgeo.csnames[el.args[0]];
+            var normalizedmid = List.normalizeZ(csgeo.csnames[el.args[0]].homog);
             var rad = el.radius;
-            var xx = CSNumber.div(mid.homog.value[0], mid.homog.value[2]).value.real;
-            var yy = CSNumber.div(mid.homog.value[1], mid.homog.value[2]).value.real;
-            dx = xx - mouse.x;
-            dy = yy - mouse.y;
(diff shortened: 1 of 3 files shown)


src/js/Events.js
```diff
@@ -35,13 +35,21 @@ function getmover(mouse) {
             dy = yy - mouse.y;
             var ref = Math.sqrt(dx * dx + dy * dy);
             dist = ref - rad.value.real;
-            dx = 0;
-            dy = 0;
+
+            if (ref > 0) {
```
strobelm: I don't get what's going on here?

kortenkamp: It checks whether the distance of the mouse coordinates to the reference point on the circle is larger than 0, I guess. But it looks a bit strange to me, too.

montaga: `dx` and `dy`  should store the vector of the mouse to the actual thing that is moved (compare, for instance, the branch for `if (el.kind === "P")`).

In the `if (el.kind === "C")`-branch, `dx` and `dy` is first "wrongly" initialized as the vector of the mouse to the center of the circle (We use these values to compute `dist`). However, we do not move the center of circle, the reference point is supposed to a point on the circle (which has the coordinates `(xx-dx/ref*rad.value.real,xx-dx/ref*rad.value.real)`, the projection of the mouse to the circle ). In the `if (ref > 0) {`-branch I update `dx`/`dy` accordingly. `ref>0` is just checked to avoid division by zero.
If you find this code bad to read (I am also not really happy, but I tried to be minimally invasive ), I can rewrite this branch.

strobelm:
Great!

PR #650, opened 2018-05-24
autoload katex

strobelm:
This pull request loads katex if needed. skipInit is necessary if the plugin es loaded at runtime, e.g. with EditableText.


kortenkamp:
This is a very welcome addition!!

PR #652, opened 2018-05-25
playtone and playmelody - rebased

montaga:
This fixes the timing bug of https://github.com/CindyJS/CindyJS/pull/426 and is rebased with the current master.
The timing seems to be correct on current versions of Firefox, Chrome and Safari :-)


montaga:
If travis plays along with this, then the soundfonts will be downloaded via `make deploy` or `make soundfonts` and hence we do not need to upload the soundfonts to cindyjs.org manually.
However, `make deploy` needs some more time.

PR #653, opened 2018-05-28
update to closure compiler 20180506

strobelm:
We haven't updated the closure compiler for quite a while and it also contains improvements for object literals (which we use heavily). See https://github.com/google/closure-compiler/wiki/Releases#september-10-2017-v20170910

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - node
-  - 4
+  - 8
 script: node make build=release alltests
 dist: trusty
 sudo: false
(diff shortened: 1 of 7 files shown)


kranich:
The decision back then was to update Closure Compiler just once in awhile, when it has changes that we benefit from.

kranich:
Apparently, unlike with a lower node version, in `tools/prepare-deploy.js` the dependencies in `lib` and `node_modules` were prefixed with `build/js`. I have not found out what caused this, so I am not sure if 
d6fed5b is the proper fix.

@strobelm What do you think?

kranich:
There are some (new?) Closure Compiler warnings for CindyJS. Maybe we can address these as well while we are at it?

strobelm:
Oh I completely forgot about this PR. The fix looks fine to me. Thanks a lot @kranich!

PR #654, opened 2018-05-28
build ComplexCurves only for `make ComplexCurves` or `make deploy`

montaga:
With this, the ComplexCurves-Plugin is only built when `make ComplexCurves` or `make deploy` is executed, but it will be avoided for a `make`, `make live`, etc.

This reduces the build-complexity for most users.


kranich:
`make alltests` still causes ComplexCurves to be built:

```
[ComplexCurves.plugin] Forcing run of ComplexCurves.plugin since setting 'cc_closure_warnings' changed value
```

Issue #500, opened 2016-09-28
Distinguish text and currenttext for EditableText


gagern:
As @kortenkamp wrote in https://github.com/CindyJS/CindyJS/pull/499#issuecomment-250219642:

> In Cinderella, `currenttex` has different semantics than `text`. With `currenttext` you can retrieve the current text in the `EditableText`, even if it has not been confirmed with the enter key.

We should probably do something similar on the CindyJS side. Use the `text` property of the geometric element to reflect `text` and the `value` property of the HTML element to reflect `currenttext`. Use the enter key handling from #493 to copy from the latter to the former. Perhaps also perform that copying in response to [the `change` event](https://developer.mozilla.org/en-US/docs/Web/Events/change), since users might commit changes by switching focus without pressing enter.

Timing could be tricky, since we also have event listeners to run scripts in response to key events on the text box. If we want to ensure that we commit the text first and run the script afterwards, then we probably have to consolidate the keydown-for-enter and the keydown-for-CindyScript handlers into a single event handler which does things in the correct order.

montaga:
has been fixed through #656

PR #656, opened 2018-05-28
fix accessors

strobelm:


montaga:
my last commit should respect #500

PR #657, opened 2018-05-29
also update .text if change event is triggered

montaga:
This is a little fix to #656, which was fixing #536


strobelm:
Look fine, could you just restore ``GeoRender.js``?

PR #658, opened 2018-05-29
run make beautify

montaga:


strobelm:
Interesting that travis didn't catch this...

PR #661, opened 2018-06-03
CindyGL: enhance random()

montaga:
This pseudo-random function does yield fewer artifacts on a set of selected examples and furthermore, requires fewer arithmetic operations than the previous function.


kortenkamp:
image

I'm not sure that this is really solved… the artifacts are clearly visible…

montaga:
Okay, we are not ready yet with this and we need a more expensive way to generate the pseudo-random numbers. Thank you for this further testcase.

PR #671, opened 2018-06-19
skip I, J, i, j during create$2, use P0, S0, etc...

montaga:
The original lookup in Cinderella skips I an J as point names (for good reason). That should happen for create$2 as well. Also i and j are skipped for line names.

Furthermore, the names P0, S0, C0 will be used (as in Cinderella).


kortenkamp:
I am not sure about i and j for lines, but actually it’s good to skip them as well. I’ll check in Cinderella.

kortenkamp:
Ah, great. And not a surprise.

PR #672, opened 2018-06-25
Ecmascript 6

strobelm:
*edit* While we are at it: I added shims for es6 and 7 for compatibility.

Hi,

this PR enables ES6 for CindyJS and updates the closure compiler to a more recent version. 

@kortenkamp pointed out that we should still support Internet Explorer 11. Therefore i changed the compilation process to accept ES6 as input and the closure compiler will transpile the code to ES5. 
To check if transpilation is possible I added closure compiler to the tests.

Code which can't be transpiled e.g.
```
function *f(x) {
  x.someMethod(yield 1);
}
```
will throw errors like
```
closure             ] src/js/Head.js:4: ERROR - This code cannot be converted from ES6. Undecomposable expression: Please rewrite the yield or await as a separate statement.
[closure             ] See https://github.com/google/closure-compiler/wiki/FAQ#i-get-an-undecomposable-expression-error-for-my-yield-or-await-expression-what-do-i-do
[closure             ]   x.someMethod(yield 1);
[closure             ]                ^^^^^^^
[closure             ] 
[closure             ] 1 error(s), 6 warning(s)
```.

### DIFF
--- a/Administration/jshint.conf
+++ b/Administration/jshint.conf
@@ -7,5 +7,6 @@
   "undef": true,
   "unused": false,
   "maxerr": 5000,
-  "predef": ["console", "numeric", "ClipperLib", "convexhull", "module", "enableInlineVideo", "ImageData"]
+  "predef": ["console", "numeric", "ClipperLib", "convexhull", "module", "enableInlineVideo", "ImageData"],
+  "esversion" : 6
 }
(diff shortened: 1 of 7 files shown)


montaga:
With the recent commit, `make` will behave the same as a `make build=release` per default. The only disadvantage I can see is that debugging is harder and the build time is a bit higher because the closure compiler is run for the core Cindy.js (before that it was used for several plugins anyway).
If one likes to use the old setting, still `make build=debug` can be used. I think it is reasonable for developers to use this setting.

PR #673, opened 2018-06-25
add interpolate modifier for drawimage

montaga:


kortenkamp:
This looks fine to me - thanks a lot for the implementation!

Only thing that could be discussed is whether a way to globally change the default would be nice, but that would be another ER.

strobelm:
Hi, since ctx.imageSmoothingEnabled is not a web standard we should probably check if the property exists.

See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled#Browser_compatibility

kortenkamp:
What happens if we set it although it does not exist?

PR #675, opened 2018-07-02
Remove duplicates

strobelm:
This PR uses randomized proving to find duplicate elements.


src/js/libgeo/GeoBasics.js
```diff
@@ -296,6 +312,19 @@ function addElementNoProof(el) {
         csgeo.ifs.push(el);
     }
 
+    // collect sets
+    var setsRe = new RegExp("^[P|L|S|C]s$"); // Ps, Ls, ...
```
montaga: better use `var setsRe = /^[P|L|S|C]s$/;` for the constant RegExp here, because using the RegExp constructor function leads to possible slower runtime compilation of the regular expression. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions )

PR #676, opened 2018-07-03
Implement 'keys' operator

strobelm:
This implements the 'keys' operator for colon dictionaries.

### DIFF
--- a/examples/keys.html
+++ b/examples/keys.html
@@ -1,0 +1,56 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Cindy JS Example</title>
+<meta charset="UTF-8">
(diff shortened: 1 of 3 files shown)


montaga:
Thank you!
Unfortunately, the code
`
forall(keys(A), k,
  errc(k + ": " + A:k)
);
`
does not work. However, if one uses `return General.string(k);` instead of `return General.string('"' + k + '"');`, then it works. Should we change it like that?

montaga:
I checked it with the current release of Cinderella, and yes, there are also no `"`s. So I will change this...

strobelm:
Ah nice catch! I deliberately changed it since the documentation states that there should be double quotes.

strobelm:
Should be fixed now.

montaga:
thanks :)

PR #677, opened 2018-07-06
Lastpointertype

strobelm:
This PR implements ```lastpointertype()``` which will return "mouse", "touch" or "pen" to CindyScript. If no pointing device ever entered then we will return ```nada```.

### DIFF
--- a/src/js/Events.js
+++ b/src/js/Events.js
@@ -465,9 +465,10 @@ function setuplisteners(canvas, data) {
     addAutoCleaningEventListener(canvas, "touchmove", touchMove, true);
     addAutoCleaningEventListener(canvas, "touchend", touchUp, false);
 
-    addAutoCleaningEventListener(window, "pointerover", function(e) {
+    addAutoCleaningEventListener(canvas, "pointerover", function(e) {
         mouse.pointerType = e.pointerType;
     });
+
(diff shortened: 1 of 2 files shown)


kortenkamp:
May I suggest a different name for that function?

First – I don't think that "last" is necessary, as `mouse()` is not called `lastmouse()` either. If you really want to use such a preposition, I suggest to use "current".

Second – I suggest to use `inputdevice` instead of `pointerdevice`.

Third – great to have this! Please let us also discuss whether we can implement Cinderella's `mtlocal()` also, as touch devices suggest multi-touch.

montaga:
I agree that the function should be renamed. `inputdevice` would appear logical to me.

We should also take in consideration that the implementation would not work on current versions of safari because the `pointerover` event is[ not supported yet](https://caniuse.com/#search=pointerover). Apple luckily decided to implement PointerEvents [recently in Safari 13](https://developer.apple.com/documentation/safari_release_notes/safari_13_release_notes), though still many functions are missing. But I think it is just a question of time until these features become available.


For the time beeing, the [mulitiid()>0](https://github.com/CindyJS/CindyJS/blob/master/ref/User_Input.md#getting-the-current-multi-id-multiid) can be used to distinguish touches from the mouse.

PR #680, opened 2018-07-10
Recursively delete elements

strobelm:
This PR implements the deletion of dependent elements. We have to build a dependency tree since we only track a directed graph in the "wrong" direction with geoel.args.  

This PR also enables to call ```removeelement``` from CindyScript.

### DIFF
--- a/src/js/libgeo/GeoBasics.js
+++ b/src/js/libgeo/GeoBasics.js
@@ -362,71 +362,30 @@ function addElementNoProof(el) {
 
 // TODO Remove dependencies also
 function removeElement(name) {
-    var i, el, debug = false;
+    var i, el, debug = true;
     if (debug) console.log("Remove element " + name);
 
     // TODO Check if name exists
     delete csgeo.csnames[name];
 
-    for (i = 0; i < csgeo.gslp.length; i++) {
-        el = csgeo.gslp[i];
-
-        if (el.name === name) {
-            if (debug) console.log("Removed element from gslp " + name);
-            csgeo.gslp.splice(i, 1);
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


kortenkamp:
I don’t have the time to review the code right now, but I remember that I had implemented this 20 years ago for Cindy and at first the runtime was quadratic in the number of elements of the construction, though linear runtime is possible. This was quite a problem, so I just ask someone to check this ;)

strobelm:
Oh i forgot to update the incidences, i'll close the PR and reopen it asap.

strobelm:
Updating the incidences and furthermore we remove all elements at once which should improve runtime complexity.

strobelm:
Thank @kortenkamp for pointing this out!

src/js/libgeo/GeoBasics.js
```diff
@@ -360,75 +360,91 @@ function addElementNoProof(el) {
...
-                if (debug) console.log("Removed element " + name + " from set of " + sname);
+            if (cmpMap[setel.name]) {
+                if (debug) console.log("Removed element " + setel.name + " from " + arrn);
                 return false;
             } else return true;
         };
     };
 
+    // update incidences
+    keys.forEach(function(name) {
```
montaga: Honestly, it is not clear to me, why this might not cause a quadratic running time in pathological cases.

But heuristically, I am willing to believe that this runs smooth enough in most cases because there are only a few incidences.

kortenkamp: The best way to avoid quadratic runtime is to mark all elements that are to be deleted and not recurse for the marked ones. 

Again, sorry for just dropping comments and not implementing it properly, but that's the only thing I can do within the 2 minutes of attention span…

PR #681, opened 2018-07-11
Editor for CindyJS-files

montaga:
This PR includes a current dump from [https://montaga.github.io/cindyeditor/](https://montaga.github.io/cindyeditor/), which is a prototype of how a proper future editor for CindyJS-applets could look. The work started with @richter-gebert s `UIExperiments.html`. I have added

* a windows-based UI,
* For the intermediate state, a system to create plugins for a applet based on CindyScript code.
* a rudimentary version of an import and export (which should work fine with the exports from Cinderella)
* a very rudimentary Inspector (actually we need an `inspect`-command to do this properly),
* some basic configuration to set the visibleRect and
* a script editor, which is based on CodeMirror.

I am aware that the editor is far from being finished and that the code at some places is not tidy and several side effects are not handled correctly.  Furthermore, the editor right now is based on ES6, which is not supported by some browsers. (This could be solved by a transpilation one day)

But I think that this place is more appropriate for the further development of the editor, and I invite you to contribute and enhance this editor.  I would suggest for the beginning a rather lax review policy on commits on this folder because the editor at this state is an experiment to find out what is possible.

This PR does not change anything of the core of CindyJS. And I think that the editor should remain as a separate module. However, there are some features of CindyJS that the editor would also need such as namespaces, specific CindyScript commands or infrastructure of CindyScript based plugins. These changes can be committed as PRs that modify both the CindyJS kernel (which should still be rigorously reviewed) and the editor.


kortenkamp:
I highly approve merging this, for the same reasons @montaga proposes it.

strobelm:
Hi nice to have this! We should eventually include this in the deployment process.

PR #682, opened 2018-07-12
Only run mouseclick if the courser has not moved too far

montaga:
With this PR, mouseclick in CindyJS be handeled as it is in Cinderella: If the mouse is moved some distance while having a button pressed, then the mouseclick-script will not be evoked.

Before that, the mouseclick-event coincided with the mouseup-event in almost every case.

### DIFF
--- a/examples/mouseevents.html
+++ b/examples/mouseevents.html
@@ -1,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Mouse events</title>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+m = [(0,0), (0,0), (0,0)];
+ma = [0,0,0];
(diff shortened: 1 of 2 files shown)


kortenkamp:
Ah, this reminds me of a problem I noticed – is it possible that mouse click is never invoked on tablets? At least I had to move my code from click to up in order to make the same construction work on the iPad.

strobelm:
Shouldn't be ```hasmoved``` somewhere be reset to false?

montaga:
@strobelm : `hasmoved = false;` is executed with the `mousedown`-event.

@kortenkamp : Thank you for this remark :-) With the last commit, "mouseclick" should also work for iPad (as it is the case with the other mouse-events, there is no support for multi-touch yet)

kortenkamp:
Great - thank you for adding that. As for multitouch: I would like to discuss this, as I put some thought into the implementation in Cinderella (the Java version) of multitouch and its integration with CindyScript. Maybe we should try to setup a video conference?

strobelm:
Thanks!

PR #724, opened 2018-12-17
Add delimiter modifier to format

### LINKED ISSUE #715: How to output number with fixed decimals and with comma instead of period?
I am outputing some numbers that needs to be rounded to the first decimal. I am using:

     round(num*10)/10;

which gives me `1, 1.1, 1.2 ...`

I need to print the `1` also as `1.0`.

Is there an elegant way to do this with CindyJS? In JS it is `tofixed()`.

<br>
Furthermore, is there a way to output `4,1` instead of `4.1`?

strobelm:
This PR adds a modifier to `format` to set the delimiter. This should resolve #715.

For example:
```
    > format(1.23456, 2, delimiter->",")
    < "1,23"
```

### DIFF
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -4327,7 +4327,11 @@ evaluator.format$2 = function(args, modifs) { //TODO Angles
             erg1 = erg;
             erg = erg.substring(0, erg.length - 1);
         } while (erg !== "" && erg !== "-" && +erg === +erg1);
-        return "" + erg1;
+        var tmp = "" + erg1;
+        if (modifs.delimiter && modifs.delimiter.ctype === "string") {
+            tmp = tmp.replace(".", modifs.delimiter.value);
+        }
+        return tmp;
     }
 
     function fmt(v) {
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -4328,6 +4328,7 @@ evaluator.format$2 = function(args, modifs) { //TODO Angles
             truncate = modif.value;
     }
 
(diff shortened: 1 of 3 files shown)


strobelm:
Oh I see. Should be fixed now.

montaga:
Thank you :+1: 
Are we supposed to be happy with
```
> format(exp(2*pi*i), 2, delimiter->",", truncate->false)
< 1,00 - i*0,00
```
or should it rather be
```
> format(exp(2*pi*i), 2, delimiter->",", truncate->false)
< 1,00
```
?

src/js/libcs/Operators.js
```diff
@@ -4320,22 +4320,37 @@ evaluator.format$2 = function(args, modifs) { //TODO Angles
...
 
     function fmt(v) {
         var r, i, erg;
         if (v.ctype === 'number') {
-            r = fmtNumber(v.value.real);
-            i = fmtNumber(v.value.imag);
-            if (i === "0")
+            r = fmtNumber(v.value.real, truncate);
+            i = fmtNumber(v.value.imag, truncate);
+            if (v.value.imag === 0.0)
```
montaga: What about
```suggestion
            if (i === fmtNumber(0, truncate))
```
maybe also the case ` i===-0` should be checked as well.

strobelm:
Nice catch @montaga. I added some unittests to verify the function.

montaga:
great. I noticed that this bug appeared even before adding this feature in CindyJS. But, now it behaves as Cinderella does and we have these nice modifiers :+1: .

Issue #700, opened 2018-09-30
PointOnConic


bellezaodiada:
It seems this is still not implemented correctly, as discussed in issue 166.

bellezaodiada:
And re-raised in issue 635

PR #725, opened 2018-12-17
Point on conic

montaga:
Due high demand (#700 and some visualizations for mathe-vital.de), we should have a working version of PointOnConic. However, tracing this geometric operation is difficult (also compare with the long mail from @gagern at 04.11.16 in our mailing list).

So I am giving a working implementation as a mixture of 
* project to nearest PointOnConic by @gagern. This implementation will only used if the point itself is moved. His comments on his code:
> This works reasonably well for moving the point with the mouse.  The combination of orthogonal projection and Euclidean distance, discarding complex points of intersection, will lead to a very intuitive placement of the point at the nearest position on the conic.  This completely fails to work if the conic moves from under the point, though, as the point will become complex in this situation.
* and the approach that is used in Cinderella, where the center of the conic is connected with the point on the conic and the farpoint of the direction is traced. This implementation will be used if the conic under the point is moved.

As this is my first implementation of a geometric operation, please review it carefully!

### DIFF
--- a/examples/157_PointOnConic.html
+++ b/examples/157_PointOnConic.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<title>Point on Conic</title>
+<meta charset="UTF-8">
+<link rel="stylesheet" href="../build/js/CindyJS.css">
+<script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script type="text/javascript">
+
+var cdy = CindyJS({
+  ports: [{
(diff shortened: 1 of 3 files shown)


strobelm:
Great! The remaining problem looks like a degenerate situation. Any idea how Cinderella behaves in this case?

montaga:
Do not merge this, since the intuitive approach that projects to the nearest point on conic seems not continuous and might break all the tracing!

strobelm:
I'll close the PR until we sorted things out.

PR #727, opened 2018-12-18
Format locale

strobelm:
This is the draft for number localisation. We should have a debate on how the modifier should be called (@kortenkamp , @richter-gebert ?). Currently it's called `locale`.

Some examples:
```
    > number = 123456.789;
    > format(123456.789, 3, locale->"de");
    < 123456,789
    > format(123456.789, 3, locale->"zh-Hans-CN-u-nu-hanidec");
    < '一二三四五六.七八九'
    > format(123456.789, 3, locale->"ar-EG");
    > '١٢٣٤٥٦٫٧٨٩'
```

### DIFF
--- a/ref/String_Operators.md
+++ b/ref/String_Operators.md
@@ -374,6 +374,12 @@ The modifier `delimiter` can be used to set decimal seperator.
     > format(1.23456, 2, delimiter->",")
     < "1,23"
 
+The modifier `truncate` can be used to prevent truncation of the output.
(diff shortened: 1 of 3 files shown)


montaga:
This is cool :+1: 
I see that it can also be globally affected through the option `language`.

It would be also cool to have an (global) `grouping` modifier, which is per default `false` (separator for thousand/lakh/crore...).

PR #729, opened 2018-12-23
extends drawtext to draw text at any angle

DivaharJayaraman:
Hi,

CindyJS has been great for some of my PhD tasks but wanted to draw text at an angle, which is missing.

Implemented a primitive version of drawing text at any angle. Three of the tests fail saying "`ctx.translate is not a function`"; they seem to run fine when I manually check in a browser.

Please have a look.

Thanks,
Divahar

### DIFF
--- a/examples/163_textAtAnyAngle.html
+++ b/examples/163_textAtAnyAngle.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>Cindy JS Example</title>
+  <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


montaga:
Its great to hear that. Thank you, @DivaharJayaraman for sharing back your contribution!
We wont merge this before all problems regarding the continuous integration are solved. `make alltests` should not return any errors.
It might be worth to try to add the functions `translate` and `rotate` to `FakeCanvas` in ` ref/js/runtests.js` (appr. line 410).

PR #731, opened 2019-01-07
Fix segments in randomized proving and make the whole process more robust

strobelm:
The prover checked the wrong property of a segment element for the start and end position, this is fixed now. Furthermore the incidence and duplicate guessing is now done in a try/catch block since both processes are not mission critical.


montaga:
looks good :)

PR #732, opened 2019-01-08
Physics: map rubber to appropriate spring.

### LINKED ISSUE #703: Behavior Rubber not implemented yet
I'm trying to convert this to CindyJS, however the physics does not work.
[GoldenGate.zip](https://github.com/CindyJS/CindyJS/files/2483172/GoldenGate.zip)

strobelm:
This is based on PR #731 and should fix #703.


strobelm:
Well it _should work_ on Internet Explorer since we merged the es6 shims. Of course we could use ```json.stringify/parse``` to achieve the same result.

I agree on the inconsistency. We could talk to @richter-gebert after the seminar tomorrow.

strobelm:
@montaga: perhaps we could set up a Windows test machine.

kortenkamp:
> I just noticed that CindyJS per default sets `stype` to 1 (rubber), which means that any spring exported from Cinderella will behave like a rubber. In Cinderella, a spring behaves more like as it has `stype` 0.
> Thus CindyJS currently is inconsistent with the export from Cinderella.

Is there a need to change the export in Cinderella? That would be easy.

strobelm:
I think this can be merged: slight incompatibility is better than non-working widgets. Perhaps @kortenkamp can at some time change the export behavior of Cinderella.

PR #750, opened 2019-01-17
WIP: fix halfplane

### LINKED ISSUE #741: drawing of degenerate shape computation crashes CindyJS
When I draw the (degenerate) shape, `halfplane(join(A,B),A)~~circle(A,1)` for two points `A` and `B` with `fill` or `draw`, CindyJS crashes.
There should be some better error handling.

strobelm:
Hi this is proposal for halfplane, please test if this fixes #741.

Furthermore this also handles the case if the defining point is incident to the defining line and returns the line (more precisely a polygon with two points on the line). I'm not sure if this is the desired behavior (although it's certainly better than numerical flickering). 

Would be great to have some input here.

### DIFF
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -3775,37 +3775,69 @@ evaluator.halfplane$2 = function(args, modifs) {
             p = w1;
             l = w0;
         }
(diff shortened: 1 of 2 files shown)


montaga:
I tested it, it will fix #741.

Issue #524, opened 2016-10-25
dash type, colour and perhaps other attributes


JHM27:
Do these work yet in CindyJS?

They don't seem to work for me!

gagern:
http://cindyjs.org/examples/108_DashType.html demonstrates some dashed lines, and many of the examples use colored lines. Even export from Cinderella works, with the caveat of #109 still being open. Please make sure you're using [the latest Cinderella beta version](http://beta.cinderella.de/public/), and attach the exported HTML file as well as the CDY file (probably renamed to `*.zip` to keep GitHub happy) so we can have a look at them.

PR #754, opened 2019-01-21
allow lineModifs for connect, drawcircle and drawarc

### LINKED ISSUE #753: modifier `dashtype` not supported for `connect`
@s-takato pointed out that `connect([[1,0],[3,4]],dashtype->3);` does not export well to CindyJS.
Using this modifier for `draw` works fine, so we already have an implementation for this modifier.
It should be an easy fix to enable this modifier also for `connect`.

montaga:
According to [the reference](https://cindyjs.org/ref/Elementary_Drawing_Functions.html), the functions `connect`, `drawcircle` and `drawarc` should handle the same modifiers as the `draw` operator. With this PR this should be the case now.

Furthermore, this should fix #753 and #524 .


montaga:
And probably this also fixes #524

strobelm:
Hi!
Thanks for looking into this. Handling non-degenerate arcs sometimes as lines and sometimes as conics feel a bit dissonant. If I read this correctly this would also rule out filling if dashing is enabled.
Actually is there a reason not to enable dashing for conics? If not then I would rather enable conic dashing  for example like this:

https://github.com/strobelm/CindyJS/tree/drawmodifs

montaga:
Thank you @strobelm . 
Your code is more readable and it makes conic dashing possible, for instance with `drawconic([[1,0,0],[0,2,0],[0,0,-3]], dashtype->4);`
So I merged it into this branch.
The only reason I see why one might not like to have dashing for conics is because the dashing pattern might be inconsistent with the ``circumference'' of the conic, but this rather a tiny problem on the users side.

PR #755, opened 2019-01-22
WIP: Div by 0

strobelm:
Here is my proposal for division by 0.

It implements the arithmetic of CP^1, so all signs are ignored for infinity. Internally ``-Infinity`` still can be used although ```niceprint``` will ignore the sign. On the low level ```∞``` is defined as (∞,∞) so both, real and imaginary part are infinite. 

Comments are very welcome.


kortenkamp:
Hm. I don’t understand why you deliberately remove the distinction between -Infty and Infty. Yes, CP^1 is nice, but not always what you want (in particular if you stay in the reals). 

Another question for me would be whether CindyGL uses the same arithmetic – I think we should stick to whatever is there for high performance.

PR #756, opened 2019-01-30
WIP: OpSound

pdywilson:
This is a first working implementation of OpSound.


src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,403 @@
...
+      var masterGain = audioCtx.createGain();
+      masterGain.gain.value = 0;
+      lines[line] = {oscNodes: [], masterGain: masterGain};
+      for(var i=0; i < harmonics.length; i++){
+        lines[line].oscNodes.push(playOscillator(line, partials[i].value.real*(i+1)*freq, harmonics[i].value.real));
+      }
+      masterGain.gain.linearRampToValueAtTime(amp, audioCtx.currentTime+attack);
+      masterGain.connect(audioCtx.destination);
+      if(damp > 0){ //not properly mathematically but does the job
+        masterGain.gain.setTargetAtTime(0.0, audioCtx.currentTime+attack,(1/damp));
```
montaga: maybe it is worth to try to stop here all the oscillating nodes after 6*(1/damp) seconds, because then the sound will be unhearable anyway (e^-6<1/256).

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,457 @@
...
+            else if (erg.ctype === 'string'){
+              var val = erg.value;
+            }
+            else {
+                var val = 0;
+            }
+            if (val === 0) {
+              line = 0;
             }
+            else if (!linesDict[val]){
```
montaga: better use `!linesDict.hasOwnProperty(val)` here, because `linesDict[0]=0`, and therefore always `!linesDict[0]` holds

montaga:
Could you please run `make beautify` and commit?

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,457 @@
...
-    if (v0.ctype === 'number') {
-        handleModifs();
-        var lines = sound.lines;
-        var f = v0.value.real;
-        if (lines[linenumber] === 0) {
-            // Was bitte sollte die Funktion T an dieser Stelle sein?
-            // lines[linenumber]=T("sin", {freq:f,mul:0.6}).play();
+    handleModifs();
+    if (harmonics.length>partials.length){
+      partials = Array(harmonics.length).fill(CSNumber.real(1))
```
montaga: missing semicolon. 
Some output "Ignoring modifier partials, because the length of partials does not match the length of harmonics" would be nice

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,457 @@
...
+              restart = erg.value;
+          }
         }
     }
 
+    var freq = evaluate(args[0]).value.real;
+    var amp = 0.5;
+    var damp = 0;
+    var duration = 1;
+    var harmonics = List.asList(CSNumber.real(1)).value;
```
montaga: `var harmonics = [CSNumber.real(1)];` would be shorter

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,457 @@
...
+    var line = 0;
+    var partials = Array(harmonics.length).fill(CSNumber.real(1));
+    var attack = 0.01;
```
montaga: It would be good if this constant could also be changed via a modifier.

montaga:
A general question, since I am not an export on the webaudio API: Can you recycle the old oscNodes, after they stopped? Because right now the elements of `lines[line].oscNodes` are always replaced by new ones generated via `playOscillator`.

src/js/libcs/OpSound.js
pdywilson: tried this, but throws random errors, probably because, if it is stopped elsewhere, it can't be stopped again. I dont think this makes a big difference anyways.

pdywilson:
> A general question, since I am not an export on the webaudio API: Can you recycle the old oscNodes, after they stopped? Because right now the elements of `lines[line].oscNodes` are always replaced by new ones generated via `playOscillator`.

Nope, this is one of the key concepts of the web audio API that you "fire and forget", so after one oscnode is stopped you cant start it again. what we could do is never stop oscillatornodes and only turn off the gain and turn it on again when the note should be played again. This would mean once started my oscillators always keep oscillating in the background quietly even if i dont need the sound. I don't know which approach is better, do we still have issues with "glucksen" or other issues? because i have never experienced this on my machine.

I implemented the other suggestions and ran make beautify.

montaga:
Currently there is a bug if I just run `playsin(440)` without a `line` modifier: `lines[0]` apparently is not initialized then.

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,450 @@
...
+                lines[line].masterGain.gain.linearRampToValueAtTime(0.0, audioCtx.currentTime + release);
+                for (var i = 0; i < lines[line].oscNodes.length; i++) {
+                    lines[line].oscNodes[i].oscNode.stop(audioCtx.currentTime + release); //helps
                 }
-                if (linenumber > 10) {
-                    linenumber = 10;
+                lines[line] = 0;
+            } else {
+                for (var i = 0; i < harmonics.length; i++) {
+                    lines[line].oscNodes[i].oscNode.frequency.setValueAtTime(partials[i]*(i + 1) * freq, audioCtx.currentTime);
```
montaga: partials[i].value.real ?

montaga:
`playwave` also does not work due to a missing conversation of the CindyJS-list to a Float32Array.

src/js/libcs/OpSound.js
pdywilson: true!

montaga:
These images for the examples are not contained in the PR. We should check back with @richter-gebert whether we want to include these examples.

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
+    var ac = null;
+    if (!window.AudioContext && !window.webkitAudioContext) {
+        console.warn('Web Audio API not supported in this browser');
+    } else {
+        ac = new(window.AudioContext || window.webkitAudioContext)();
+    }
+    return function() {
+        return ac;
+    };
+}();
```
montaga: Remove `()` here; do not call this function here! We do not need the to initalize the audioContext whenever CindyJS is loaded. Then also `make alltests` does not throw errors (however, you should consider all the warnings from `make jshint`)

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
+};
 
 evaluator.playsin$1 = function(args, modifs) {
+    if (!audioCtx) {
+        audioCtx = getAudioContext();
+    }
 
     function handleModifs() {
         var erg;
+        if (modifs.amp !== undefined) {
```
montaga: maybe you can avoid some redundancy by using a helper function here. (`damp = realmodif("damp", 0 [default value])` etc.)

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
+        }
+    }
+
+    var freq = evaluate(args[0]).value.real;
+    var amp = 0.5;
+    var damp = 0;
+    var duration = 1;
+    var harmonics = [CSNumber.real(1)];
+    var line = 0;
+    var partials = Array(harmonics.length).fill(CSNumber.real(1));
```
montaga: maybe the code would be more readable and error-prone if you avoid the CSNumer data structure internally for `harmonic` and `partials`.

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
+        }
+        if (duration >= 0) {
+            gainNode.gain.setValueAtTime(0, audioCtx.currentTime + duration + attack);
+        }
+        return {
+            oscNode: oscNode,
+            gainNode: gainNode
+        };
+    }
+
```
montaga: In general you could add some more comments for the following code.

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
-        handleModifs();
-        var lines = sound.lines;
-        var f = v0.value.real;
-        if (lines[linenumber] === 0) {
-            // Was bitte sollte die Funktion T an dieser Stelle sein?
-            // lines[linenumber]=T("sin", {freq:f,mul:0.6}).play();
+evaluator.playsin$0 = function(args, modifs) {
+    var erg;
+    if (modifs.line !== undefined) {
+        erg = evaluate(modifs.line);
```
montaga: we have seen the same code in line 64. Avoid redundancy!

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
-            lines[linenumber].set({
-                freq: f
-            });
+
+evaluator.playfunction$1 = function(args, modifs) {
+    if (!audioCtx) {
+        audioCtx = getAudioContext();
+    }
+
+    function handleModifs() {
```
montaga: maybe you could make `handleModifs` one level more global and reuse the old one from above (avoid redundancy)

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
...
+        }
+        if (modifs.release !== undefined) {
+            erg = evaluate(modifs.release);
+            if (erg.ctype === 'number') {
+                release = erg.value.real;
+            }
+        }
+    }
+
+    function getBufferNode(arr) {
```
montaga: (almost) redundant

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,533 @@
 //*******************************************************
 // and here are the definitions of the sound operators
 //*******************************************************
+var lines = [0];
+var linesDict = {0: 0};
```
montaga: maybe you could make one data-structure out of this (only `lines = {}`).

src/js/libcs/OpSound.js
```diff
@@ -1,48 +1,312 @@
...
+        let bufferNode = ctx.createBufferSource();
+        bufferNode.buffer = buffer;
+        return bufferNode;
+    }
+};
 
-var sound = {};
-sound.lines = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0];
+evaluator.stopsound$0 = function() {
+    OpSound.audioCtx.close();
```
montaga: we need some `if(OpSound.audioCtx)` before this line

montaga:
It would be good if you could add some mini-examples for each of the implemented functions and their modifiers.

montaga:
great :-) If everything works fine, it can be merged, but I want to take a look into that

Two more feature requests for modifiers by @richter-gebert: (no need to include them yet into this PR)
* `pan`, a value between `-1` and `+1` indicating which speakers are used. (low priority)
* `phaseshift` for `playsin`, a value between `0` and `2*pi` (otherwise interpreted mod(2 pi)) indicating a phase offset. I,e, if `phaseshift->a`, the waves `sin(omega t + a)=sin(omega t) cos(a)+cos(omega t) sin(a)` for frequency `omega/(2 pi)`  are played instead of `sin(omega t)`. (higher priority)

montaga:
the examples sounds great :+1: (only the `random()` noise was a bit quiet)

src/js/libcs/OpSound.js
```diff
@@ -138,23 +138,25 @@ evaluator.playsin$1 = function(args, modifs) {
...
     function setPhaseShift(oscNode) {
         //set coefficients of the fourier transform
+        let real, imag;
         if (noPartialsGiven) {
-            var real = new Float32Array(harmonics.length);
-            var imag = new Float32Array(harmonics.length);
-            imag[1] = harmonics[0]*Math.cos(phaseshift);
-            real[1] = harmonics[0]*Math.sin(phaseshift);
+            real = new Float32Array(harmonics.length);
+            imag = new Float32Array(harmonics.length);
```
montaga: `harmonics.length + 1`

src/js/libcs/OpSound.js
```diff
@@ -76,24 +96,35 @@ evaluator.playsin$1 = function(args, modifs) {
...
+    let partials = OpSound.handleModif(modifs.partials, 'list', [1]);
+    let attack = OpSound.handleModif(modifs.attack, 'number', 0.01);
+    let release = OpSound.handleModif(modifs.release, 'number', 0.01);
+    let restart = OpSound.handleModif(modifs.restart, 'boolean', true);
+    let phaseshift = OpSound.handlePhaseshiftModif(modifs.phaseshift, 'list', harmonics.length);
+    let pan = OpSound.handleModif(modifs.pan, 'number', 0);
     let noPartialsGiven = harmonics.length > partials.length;
 
+
+    if (!noPartialsGiven) {
```
montaga: you should do this after
`
if (harmonics.length > partials.length) {
        partials = Array(harmonics.length).fill(1);
}
`

src/js/libcs/OpSound.js
```diff
@@ -76,24 +96,35 @@ evaluator.playsin$1 = function(args, modifs) {
...
+        for (let i = 0, j = 0; i < harmonics.length; i++) {
+            if (Math.abs(partials[i] - 1) < 1e-10) j++;
```
montaga: `noPartialsGiven = noPartialsGiven & (Math.abs(partials[i] - 1) < 1e-10)`
would be more simple and `j` could be avoided.

(and partials[i] might not exist before `if (harmonics.length > partials.length) { partials = Array(harmonics.length).fill(1); }`)

src/js/libcs/OpSound.js
```diff
@@ -136,22 +169,23 @@ evaluator.playsin$1 = function(args, modifs) {
         }
     }
 
-    function setPhaseShift(oscNode) {
+    function setPhaseShift(oscNode, curphaseshift) {
```
montaga: for simplicity, why not passing `phaseshift` instead of `phaseshift[0]==curphaseshift`? (or even using the variable, which is in the scope)

pdywilson: for `precalculated == false` setPhaseShift is called for each harmonic individually to call createWave for each Node, so then we pass phaseshift[i] as curphaseshift into the function. otherwise we don't use curphaseshift but the phaseshift from the variablescope. It all gets a little twisted because we use multiple nodes for harmonics in the one case but a single node for the other case.

montaga:
closed since it is contained in #776

PR #757, opened 2019-01-31
JSON support

strobelm:
This PR adds JSON support to CindyScript. 
For example:
```
circ = circle(A,1);
json1 =  
 {
    "test" : circ,
    "undef" : undef,
    "name":"John",
    "age":30,
    "bool": false,
    "cars": [
        { "name":"Ford", "models":[ "Fiesta", "Focus", "Mustang" ] },
        { "name":"BMW", "models":[ "320", "X3", "X5" ] },
        { "name":"Fiat", "models":[ "500", "Panda" ] }
    ]
};
```
yields the following results:
```
json1.name --> "Joe", json.test --> "Circle"
``` 

The setters work analogously.

Geometric objects (as values) are also supported. Keys are of type string only. I could add geo objects (using niceprint) if desired.

### DIFF
--- a/ref/Dictionaries.md
+++ b/ref/Dictionaries.md
@@ -25,10 +25,10 @@ The expression `{"key1" : val, ...}` created a basic dictionary.
 ### Printing a dictionary
 
 **Description:**
-As usual the function `print` will output the object. This will generally be not a valid JSON object: therefore one can use the modifier `JSON`. The output then can be used by a JSON parser.
+As usual the function `print` will output the object. This will generally be not a valid JSON object, since CindyJS uses different data types than JavaScript.
 
-    > print({"a" : 1, "b" : false, "c": "mystring"}, JSON->true);
-    * {"a":1,"b":false,"c":"mystring"}
+    > print({"a" : 1, "b" : false, "c": "mystring"});
+    * {a:1, b:false, c:mystring}
 
 **Circular Dictionaries**
 Circular objects are allowed, but a printout will be shortened. The default recursion depth is 1000 but can be adjusted with the modifier `maxDepth`:
--- a/ref/Dictionaries.md
+++ b/ref/Dictionaries.md
@@ -205,6 +205,21 @@ The variable is local to the expression.
 
 ------
 
+
+### Deleting keys from a dictionary:
+
+A key-value pair can be deleted from a dictionary by setting its value to an undefined value.
+
+    > json = {"a":1, "b": 2};
+    > json.a = nada;
+    * Warning: Accessing undefined variable: nada
+    > forall(json, k, println(k));
+    * 2
+    > keys(json)
+    < ["b"]
+
+------
+
 # Deprecated `dict`
 
 This should not be used anymore.
--- a/ref/Dictionaries.md
+++ b/ref/Dictionaries.md
@@ -1,24 +1,213 @@
 # Dictionaries
 
-Dictrionaries are maps from keys to values.
-Like all other data structures in CindyJS, dictionaries are immutable.
-Modifying a dictionary does in fact create a new dictionary.
-Since dictionary keys have to be compared in full for dictionary lookups,
-users are advised against using large data structures as dictionary keys.
-The order of key-value pairs in a dictionary is implementation-defined.
+Dictrionaries are maps from keys to values. We designed the syntax closely to JSON. **Note that entries are are given by reference** (like geometric objects).
 
-**This is an experimental feature in its early stages of development.**
+###  Elementary Dictionary Operations
 
-All of the features described here are subject to changes without notice,
-i.e. may change even without a major version bump.
-If you want to avoid accidents, please inform the CindyJS development team
-of how you are using dictionaries, so that we can try to maintain compatibility
-for your use cases or inform you if things are likely to break soon.
+**Description:**
+Without arguments, this creates an empty dictionary.
 
-At this early stage, a lot of the functionality is only available through
-named functions but should become available through operator symbols, too.
-The named functions will probably remain as an alternative
-for the sake of compatibility and clarity.
+    > {}
+    < {}
+
+------
+
+### Creating a simple dictionary: `{"key" : val, ...}`
+
+**Description:**
+The expression `{"key1" : val, ...}` created a basic dictionary. 
+
+    > {"a" : 1, "b" : 2};
+    < {a:1, b:2}
+
+------
+
+### Printing a dictionary
+
+**Description:**
+As usual the function `print` will output the object. This will generally be not a valid JSON object: therefore one can use the modifier `JSON`. The output then can be used by a JSON parser.
+
+    > print({"a" : 1, "b" : false, "c": "mystring"}, JSON->true);
+    * {"a":1,"b":false,"c":"mystring"}
+
+**Circular Dictionaries**
+Circular objects are allowed, but a printout will be shortened. The default recursion depth is 1000 but can be adjusted with the modifier `maxDepth`:
+
+    > dic={"a" : 1}; dic.b = dic; print(dic, maxDepth->3);
+    * Warning: We visited a key-value pair very often or encountered a very deeply nested dictionary. Dictionary is probably cyclic. Output will be probably incomplete.
+    * {a:1, b:{a:1, b:{a:1, b:{a:..., b:...}}}}
+
+
+------
+
+### Accessing Elements of Lists: `‹dict›.name`
+
+**Description:**
+Basic (static) access is possible via the dot operator. The variable after the dot is **not** evaluated.
+
+    > json = {"a" : 1, "b" : 2};
+    > json.a
+    < 1
+
+------
+### Retrieving all keys and values of a dictionary: `keys(‹dict›)` and `values(‹dict›)`
+
+**Description:**
+To retrieve all keys or values of a dictionary as a list one can use `keys(‹dict›)` and `values(‹dict›)`
+
+
+    > json = {"a" : 1, "b" : 2};
+    > keys(json);
+    < ["a", "b"]
+    > values(json);
+    < [1, 2]
+
+------
+
+### Dynamically accessing Elements of Lists: `‹dict›_‹var›` and `take(‹dict›,‹var›)`
+
+**Description:**
+One can dynamically access the individual elements of a dict either with the infix operator `‹list›_‹var›` or the functional operator `take(‹list›,‹var›)`. The keys (`var`) can be of any printable type.
+
+
+    > json = {"a" : 1, "b" : 2};
+    > json_"a"
+    < 1
+    > take({"a" : 1, "b" : 2}, "b")
+    < 2
+
+------
+
+### Applying an expression: `apply(‹dict›,‹expr›)`
+
+**Description:**
+This operator generates a new list by applying the operation `‹expr›` to all elements of a dict (but not to the keys) and collecting the results.
+As usual, `#` is the run variable, which successively takes the value of each element in the list.
+
+    > apply({"a" : 1, "b" : 2}, #^2);
+    < {a:1, b:4}
+
+------
+
+### Applying an expression: `apply(‹dict›,‹var›,‹expr›)`
+
+**Description:**
+Similar to `apply(‹list›,‹expr›)`, but the run variable is now named `‹var›`.
+The variable is local to the expression.
+
+    > v = 123;
+    > apply({"a" : 1, "b" : 2}, v, v^2);
+    < {a:1, b:4}
+    > v
+    < 123
+
+------
+
+
+### Applying an expression: `apply(‹dict›,‹varVal›, ‹varKey›,‹expr›)`
+
+**Description:**
+Similar to `apply(‹list›,‹var›, ‹expr›)`, but with two running variables: `‹varVal›` for the value and `‹varVal›` for the key.
+The variables are local to the expression.
+
+    > apply({"a" : 1, "b" : 2}, v, k, k+v);
+    < {a:a1, b:b2}
+
+------
+
+### Selecting elements of a list: `select(‹list›,‹boolexpr›)`
+
+**Description:**
+This operator selects all elements of a dict for which a certain condition is satisfied.
+The condition is supposed to be encoded by `‹boolexpr›`.
+This expression is assumed to return a `‹bool›` value.
+As usual, `#` is the run variable, which successively take the value of all elements in the list.
+
+    > select({"a" : 5, "b" : 2}, isodd(#));
+    < {a:5}
+
+------
+
+### Selecting elements of a list: `select(‹list›,‹var›,‹boolexpr›)`
+
+**Description:**
+Similar to `select(‹dict›,‹boolexpr›)`, but the run variable is now named ‹var›.
+The variable is local to the expression.
+
+    > v = 123;
+    > select({"a" : 5, "b" : 2}, v, isodd(v))
+    < {a:5}
+    > v
+    < 123
+
+------
+
+### The forall loop: `forall(‹dict›,‹expr›)`
+
+**Description:**
+This operator is useful for applying an operation to all elements of a dict.
+It takes a `‹dict›` as first argument.
+It produces a loop in which `‹expr›` is evaluated for each **value** of the list.
+For each run, the run variable `#` takes the value of the corresponding list value.
+
+**Example:**
+
+    > a={"a" : 5, "b" : 2};
+    > forall(a,println(#))
+
+This code fragment produces the output
+
+    * 5
+    * 2
+
+**Modifiers** 
+
+forall supports the modifier `iterator` that changes the behavior of the running variable: "value" iterates of the dictionary values (the default), "key" iterates over the dictionary keys and "pair" returns an elementary dictionary element that key and value can be accessed using the ".key" and ".value property".
+
+**Example iterator key:**
+
+    > a={"a" : 5, "b" : 2};
+    > forall(a,println(#), iterator->"key")
+
+This code fragment produces the output
+
+    * a
+    * b
+
+
+**Example iterator pair:**
+
+    > a={"a" : 5, "b" : 2};
+    > forall(a,println([#.key, #.value]), iterator->"pair")
+
+This code fragment produces the output
+
+    * [a, 5]
+    * [b, 2]
+
+------
+
+### The forall loop: `forall(‹dict›,‹var›,‹expr›)`
+
+**Description:**
+Similar to `forall(‹dict›,‹expr›)`, but the run variable is now named `‹var›`.
+The variable is local to the expression.
+
+    > v=994;
+    > k=123;
+    > forall({"a" : 5, "b" : 2},v,println(v))
+    * 5
+    * 2
+    > v
+    < 994
+    > k
+    < 123
+
+------
+
+# Deprecated `dict`
+
+This should not be used anymore.
 
 ## Creating a new: `dict()`
 
--- a/src/js/libcs/Json.js
+++ b/src/js/libcs/Json.js
@@ -28,7 +28,11 @@ Json.getField = function(obj, key) {
 };
 
 Json.setField = function(where, field, what) {
-    where[field] = what;
+    if (what.ctype === 'undefined' && where[field]) {
+        delete where[field];
+    } else {
+        where[field] = what;
+    }
 };
 
 Json.GenFromUserDataEl = function(el) {
--- a/src/js/libcs/Json.js
+++ b/src/js/libcs/Json.js
@@ -65,9 +65,8 @@ Json.GenFromUserDataEl = function(el) {
 Json._helper.GenJSONAtom = function(key, val) {
     return {
         "ctype": "JSON",
-        "isJSONAtom": true,
         "value": {
-            'key': key,
+            'key': General.string(key),
             'value': val
         }
     };
--- a/src/js/libcs/Json.js
+++ b/src/js/libcs/Json.js
@@ -100,35 +100,12 @@ Json._helper.niceprint = function(a, modifs, options) {
         return Json.niceprint(a, modifs, options);
     }
 
-    if (options.printValidJSON) {
-        if (a.ctype === "number" && a.value.imag === 0) {
-            return a.value.real;
-        }
-        if ("boolean" === a.ctype) {
-            return a.value;
-        }
-        if (a.ctype === 'list') {
-            var erg = "[";
-            for (var i = 0; i < a.value.length; i++) {
-                erg = erg + Json._helper.niceprint(evaluate(a.value[i]), modifs, options);
-                if (i !== a.value.length - 1) {
-                    erg = erg + ', ';
-                }
-
-            }
-            return erg + "]";
-        }
-
-        return "\"" + String(niceprint(a)) + "\"";
-    }
-
     return niceprint(a);
 };
 
 Json.niceprint = function(el, modifs, options) {
     if (!options) {
         options = {};
-        options.printValidJSON = false;
         options.printedWarning = false;
         // track depth
         options.visitedMap = {};
@@ -138,10 +115,6 @@ Json.niceprint = function(el, modifs, options) {
         options.visitedMap.maxElVisit = 5000;
 
         if (modifs) {
-            if (modifs.JSON) {
-                let jmodif = evaluate(modifs.JSON);
-                if (jmodif.ctype === "boolean") options.printValidJSON = jmodif.value;
-            }
             if (modifs.maxDepth) {
                 let depth = evaluate(modifs.maxDepth);
                 if (depth.ctype === "number") options.visitedMap.maxLevel = depth.value.real;
@@ -167,11 +140,8 @@ Json.niceprint = function(el, modifs, options) {
                     console.log("Warning: We visited a key-value pair very often or encountered a very deeply nested dictionary. Dictionary is probably cyclic. Output will be probably incomplete.");
                     options.printedWarning = true;
                 }
-                if (options.printValidJSON) {
-                    return "\"" + key + "\"" + ":" + '"..."';
-                } else {
-                    return key + ":" + '...';
-                }
+
+                return key + ":" + '...';
             }
             // update only once a recursive call
             if (visitedMap.newLevel) {
@@ -180,22 +150,9 @@ Json.niceprint = function(el, modifs, options) {
                 visitedMap.newLevel = false;
             }
         }
-        if (options.printValidJSON) {
-            return "\"" + key + "\"" + ":" + Json._helper.niceprint(elValKey, modifs, options);
-        } else {
-            return key + ":" + Json._helper.niceprint(elValKey, modifs, options);
-        }
+        return key + ":" + Json._helper.niceprint(elValKey, modifs, options);
     }).join(", ") + "}";
 
-
-    // pretty print 
-    if (options.printValidJSON) {
-        try {
-            jsonString = JSON.stringify(JSON.parse(jsonString), null, 0);
-        } catch (e) {
-            console.log("Could not parse JSON string.");
-        }
-    }
 
     return jsonString;
 };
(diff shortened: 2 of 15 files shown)


strobelm:
What stills needs to be done:

- loops over JSON objects
- alternative access using the brackets operator (currently only dot is supported)

montaga:
Thank you for the implementation. :+1: 

Just one question before reviewing: Is the call-by-reference behavior by purpose? (i.e. `json1 = {"a":1}; json2=json1; json2.a=2;` causes `json1.a` to be 2?)

If yes, I could happily live with it and would be happy, because then this JSON structure would be very efficient and for (semi-)run-time critical purposes I would even use this JSON structure instead of lists because the lists in CindyScript destroy the asymptotic running time. Furthermore, it would be a functional difference to the already existing dict-data structure.

However, this would be different from the usual call-by-value semantics in CindyScript and should be mentioned in the documentation.

strobelm:
Oh thanks for catching this, actually this is not intended. Obviously i'm doing a shallow copy instead of a deep copy somewhere.

But still we might want to discuss this. @kortenkamp what's your opinion? I will talk to @richter-gebert tomorrow.

kortenkamp:
I would love having references for JSON (like it is the case for geometry). 

Ulli

strobelm:
I think this PR is now in a stage that can be merged. I added traverses through the dictionary (`forall, select, apply`). Forall especially iterates over the keys (like JavaScript), this might be up for discussion.  

Please test any corner cases I might not have thought of ;-).

kortenkamp:
I would love to have modifiers for forall that can change whether keys (or indices), values or key-value pairs are iterated over. I would be happy to implement that in Classic Cindy if we agree on it.

strobelm:
We discussed the topic today and thought of this:

1.  `forall(‹list›,‹expr›)` and `forall(‹list›,‹var›,‹expr›)` should iterate over **values** (no matter if list or map).
2.  We introduce a 4-ary forall: `forall(‹list›,‹value›,<key>,‹expr›)` that also sets a variable to the running index.

This preserves consistency and also introduces an easy way to iterate over the keys.

kortenkamp:
May I still vote for iterating over valued per default and then introducing a modifier that can change this behavior to iterate over keys (indices) or key,value pairs? This is also backwards compatible and removes the problem that accidentally introduced commas are harder to find and debug. 

I don’t think that overloading forall is a better solution than using a modifier. 

Please discuss again!!

richter-gebert:
My intention was

That we (or at least  Bernhard and me) wanted a version of the list-apply list-forall like:

Apply(list, v,i,prog)

Where v iterates over the content of list and I iterates through the indices.


If we introduces this then

 Apply(jasonstuff, v,i,prog)

Would be consistent with that?

Opinions???

cu

Jürgen

--------------------------------------
Prof. Jürgen Richter-Gebert
Geometrie und Visualisierung
Zentrum Mathematik --- TU München
Boltzmannstr. 3
85747 Garching bei München
Tel: 089/28918354, Sekr: 089/28918359
e-mail: richter@ma.tum.de

kortenkamp:
Again - I would appreciate pairs and modifiers!!

I think that having a 4ary version will introduce difficult to find bugs in your own scripts. And if we return (i,v) for # then it’s fine!

> Am 06.02.2019 um 15:32 schrieb Jürgen Richter-Gebert <notifications@github.com>:
> 
> My intention was
> 
> That we (or at least Bernhard and me) wanted a version of the list-apply list-forall like:
> 
> Apply(list, v,i,prog)
> 
> Where v iterates over the content of list and I iterates through the indices.
> 
> 
> If we introduces this then
> 
> Apply(jasonstuff, v,i,prog)
> 
> Would be consistent with that?
> 
> Opinions???
> 
> cu
> 
> Jürgen
> 
> --------------------------------------
> Prof. Jürgen Richter-Gebert
> Geometrie und Visualisierung
> Zentrum Mathematik --- TU München
> Boltzmannstr. 3
> 85747 Garching bei München
> Tel: 089/28918354, Sekr: 089/28918359
> e-mail: richter@ma.tum.de
> 
> 
>

strobelm:
Correct me if I'm wrong:  with the modifier we would have difficulties with unnamed lists, e.g.:
 ```   
forall([4,5,6],k,println(???_k), iterator->"key")
    * [1,4]
    * [2,5]
    * [3,6] 
```

**Comment by kortenkamp** [comment]
No, if you want to do that you just cannot access the list elements. So what... you just asked for the index. I don’t think that’s a problem. 

And I know use cases for that - for example if you just want to do something that involves the length of some list.

**Comment by richter-gebert** [comment]
Hi all,
 Let me briefly discus various some semantic/syntactic issues 


1. Independent from son I always wanted 4-ary operators for iterating through lists.

Like 

apply(list, val, ind, dosomething)


In my CindyJS Code I very often use constructs like

apply(1..length(list),ind, val=list_ind; dosomething)

Really often!

In Javascript there is the possibility to do something like that when iterating through lists
And it is very often needed.


2. I consider hash tables a la Json as a kind of particularly indexed list where the indices are not numbers but strings. 

For that reason it would be very natural to have the

apply(json, val, ind, dosomething)

In analogy to the list case

This is wehre my preference comes from (and I consider that a strong point).

3.Yes, I agree sometimes one might be interested to iterate thorough keys.
But if one wants to do so one will most probably also have the values.
So we decide for iterating through keys code will very often look like

apply(json, key, val=json_key;dosomething)

I propose (and in Classic Cinderella we had that)
A function like keys(json) that gives a list of all keys.

If one wants to intentionally iterate through alle keys 
One could use

apply(keys(json),key,dosomething)


What do you think?


cu

Jürgen

--------------------------------------
Prof. Jürgen Richter-Gebert
Geometrie und Visualisierung
Zentrum Mathematik --- TU München
Boltzmannstr. 3
85747 Garching bei München
Tel: 089/28918354, Sekr: 089/28918359
e-mail: richter@ma.tum.de

**Comment by kortenkamp** [comment]
It seems to me that my contra arguments are not heard... what is the reason for not doing it via modifier?

> Am 06.02.2019 um 22:18 schrieb Jürgen Richter-Gebert <notifications@github.com>:
> 
> Hi all,
> Let me briefly discus various some semantic/syntactic issues 
> 
> 
> 1. Independent from son I always wanted 4-ary operators for iterating through lists.
> 
> Like 
> 
> apply(list, val, ind, dosomething)
> 
> 
> In my CindyJS Code I very often use constructs like
> 
> apply(1..length(list),ind, val=list_ind; dosomething)
> 
> Really often!
> 
> In Javascript there is the possibility to do something like that when iterating through lists
> And it is very often needed.
> 
> 
> 2. I consider hash tables a la Json as a kind of particularly indexed list where the indices are not numbers but strings. 
> 
> For that reason it would be very natural to have the
> 
> apply(json, val, ind, dosomething)
> 
> In analogy to the list case
> 
> This is wehre my preference comes from (and I consider that a strong point).
> 
> 3.Yes, I agree sometimes one might be interested to iterate thorough keys.
> But if one wants to do so one will most probably also have the values.
> So we decide for iterating through keys code will very often look like
> 
> apply(json, key, val=json_key;dosomething)
> 
> I propose (and in Classic Cinderella we had that)
> A function like keys(json) that gives a list of all keys.
> 
> If one wants to intentionally iterate through alle keys 
> One could use
> 
> apply(keys(json),key,dosomething)
> 
> 
> What do you think?
> 
> 
> cu
> 
> Jürgen
> 
> --------------------------------------
> Prof. Jürgen Richter-Gebert
> Geometrie und Visualisierung
> Zentrum Mathematik --- TU München
> Boltzmannstr. 3
> 85747 Garching bei München
> Tel: 089/28918354, Sekr: 089/28918359
> e-mail: richter@ma.tum.de
> 
> 
>

**Comment by strobelm** [comment]
Actually the modifier version does not feel "Cindyonic" (or "Cindyesk"?) to me. Especially if you want both, key and value:
```
forall(map, vk, iterator->"vkpair", println([vk_1, vk_2]));
```
I have no intuition on what is key and what is value in `vk_1` and `vk_2` and it's quite lengthy.

Perhaps we could have a middle ground with an tuple key-value pair:
```
forall(map, [v,k], println([k,v]));
```
This is still 3-ary and should also be less error prone.

**Comment by kortenkamp** [comment]
Hm. The lengthiness comes through splitting and recombining. You could just write
```
forall(map, println(#), iterator->"pairs")
```

Observe that this makes it possible to use the generic `#` in the 2-ary version, which is not the case for your approach. And if we introduce `.value` and `.key` as synonyms for `.y` and `.x` respectively we can even write:

```
forall(map, println(#.key+" is the key for value"+#.value), iterator->"pairs"); 
```

Personally, I like the `[v,k]` construct, but it looks more like ruby to me than CindyScript. And it is much more difficult to implement.

**Comment by montaga** [comment]
This is likely a question of taste. I would prefer the 4-ary versions, because it is minimal in the length of required code (For me, CindyScript is more about writing fast hacks to get things done, than the first choice for well readable code). Writing `iterator->"values"` would be longer than `val=json_key`, so most of the time I would not use the modifier.

```
forall(list, [x,y], println(x));
```
could be misunderstood as lazy pattern: If `list` was a list of pairs/points, then I would read the code as some pattern matching where `x` and `y` are matched to the first/second element of the pairs.

**Comment by kortenkamp** [comment]
Well, `iterator->values` wouldn’t be necessary if iterating the values is the default...

But please consider our user target group: if we want to enable people other than @montag or @richter to do any scripting, then readable code should be possible, and avoidable bugs should be avoided. Teachers and students should be able to write code as well. 

I think of CindyScript as a rapid prototyping tool as well, but it is also the extension language for a geometry system to be used in education. I think it is really important not to close that door - so far CindyScript is accessible.

**Comment by montaga** [comment]
I do not have a strong opinion on that (only a personal preference) and may the chieftains come to a decision ;-).
______________
Back to the PR:
```
Jason = {"age":13};
Jasonica = {"age":15};

Jason.sister = Jasonica;
Jasonica.brother = Jason;

errc(Jason);
```
gives a stack-overflow.

It seems that there is still some internal call by reference behavior:
```
json = {"key":1};
json.self = json;
json.self.key = 2;
errc(json.key); //gives currently 2
```
should be, if everything was call-by-value `1` and not `2`.

Or shall we switch back to the call-by-reference scheme? I think that there are many good reasons, for instance in the example above or in graphs, it would be nice to transverse via references and the running time would be better, because not everything would have to be copied again.

It might worth to have a lock in the code of the dict, where a working call-by-value behavior has already been implemented.

In several applications it would be desirable to check weather a JSON-object has a given key and to be able to delete a key-value pair.

**Comment by strobelm** [comment]
Thanks! I'll add this to the tests. @kortenkamp and @richter-gebert kindly made a BDFL decision. Here the poorly translated version:

```
- Functions keys(….) and values(….) that return a list.
- Default value iteration in forall
- Modifier: iterate -> {key, value,pair}
- Field access: .key and .value
- 4-ary apply
```
I hope that I can come up with an implementation soon.

**Comment by strobelm** [comment]
And I agree with @montaga and @kortenkamp that we should have a reference behavior (iirc @richter-gebert did not state preferences directly).

**Comment by kortenkamp** [comment]
I am confident @richter-gebert  agrees – he introduced a workaround for adding properties to empty list objects to get this behavior in CC.

**Comment by richter-gebert** [comment]
Fully Agreed.
--------------------------------------
Prof. Jürgen Richter-Gebert
Geometrie und Visualisierung
Zentrum Mathematik --- TU München
Boltzmannstr. 3
85747 Garching bei München
Tel: 089/28918354, Sekr: 089/28918359
e-mail: richter@ma.tum.de

**Comment by strobelm** [comment]
@montaga good point with the cyclic example. Do we allow cyclic references internally and then refuse to print them or do we prevent cycles in the first place?

**Comment by montaga** [comment]
I would not disallow cycles in general because checking weather a cycle is build has non-constant running time for a assignment. And it is desirable to represent graphs via this data-structure (which in this sense probably should not be called JSON anymore), so cycles can appear in nature.

For the recursion in `niceprint` I would use some temporary `isvisited` flag attached to each object. If a element is to be printed that has already been visited, then I would output something like `...` and stop the recursion. Basically `niceprint` of a JSON would perform a DFS, which does not give us any asymptotic costs.

**Comment by kortenkamp** [comment]
Or we could do bfs with a maximum depth? Changeable through a modif?

**Comment by strobelm** [comment]
And here is quite an update. I think the code should be fine now. Please test it thoroughly (especially @richter-gebert ). Please let me know if something strikes you odd.

**Comment by montaga** [comment]
Cool! Many cyclic bugs do not appear anymore. :+1: 
_______________
`json = {"a":[1,{"b":2},3]}; json.a_2.b=34;` apparently does not parse. However, `json = {"a":[1,{"b":2},3]};(json.a_2).b=34;` does. Should it be like that? In JS `json = {"a":[1,{"b":2},3]}; json.a[1].b` would be valid.
_______________
An other observation: The output of `errc` of a (non-cyclic)-JSON, which does not contain any quotes, can not be parsed anymore, because for the declaring a dictionary we demand quotes for the key. This breaks some paradigm. I do not know, whether we could also parse dictionaries without quotes, as it would be valid in JS (but not in plain JSON). Otherwise we should IMO change the output to valid JSON.

**Comment by strobelm** [comment]
> Cool! Many cyclic bugs do not appear anymore. +1

:-)

> 
> `json = {"a":[1,{"b":2},3]}; json.a_2.b=34;` apparently does not parse. However, `json = {"a":[1,{"b":2},3]};(json.a_2).b=34;` does. Should it be like that? In JS `json = {"a":[1,{"b":2},3]}; json.a[1].b` would be valid and also in CindyScript `[[[[0]]]]_1_1_1` would be valid.

I was aware of the fact and I didn't like it either. To fix this, if I'm not mistaken, we have to change the operator's precedence, which I was a bit afraid of. I can if @richter-gebert and @kortenkamp tell me to do so ;-).

> 
> An other observation: The output of `errc` of a (non-cyclic)-JSON, which does not contain any quotes, can not be parsed anymore, because for the declaring a dictionary we demand quotes for the key. This breaks some paradigm. I do not know, whether we could also parse dictionaries without quotes, as it would be valid in JS (but not in plain JSON). Otherwise we should IMO change the output to valid JSON.

Of course we can change the behaviour to JSON by default. It's implemented that way since Cinderella/CindyJS behaves the same for arrays:
```
err(["this", "is"]);
 ["this", "is"] ==> [this,is] 
```

**Comment by montaga** [comment]
I noticed, that the `json = {"a":[1,{"b":2},3]}; json.a_2.b=34`-"bug" is not really json related. It also occurred before due to the CindyScript specification.

The expression `allpoints()_1.color = [0,1,0];` compile neither in CindyJS nor in Cinderella, because it is parsed as `allpoints()_(1.)color`. The user has to type in `(allpoints()_1).color` instead. I guess this issue will become more prevalent with JSON.

However, I doubt that we should change the JSON syntax because of that. (As I mentioned somewhere else, I would love to see `[]`-postfix accessors one day for both arrays and JSON objects in a C/Javascript-like manner, because it would lower the hurdles for many authors with experience in this world. And to start indexing with zero simplifies many scenarios. If `[]` is used together with `.` everything should be well parsable)

**Comment by kortenkamp** [comment]
I think this behavior was an incompatible introduction when we switched to the new parser for CindyJS 3. 

I would be happy to roll back to the earlier version... but probably there is no way to return.

**Comment by montaga** [comment]
The parsing problem should be solved with PR #815. If you agree with this fix, feel free to merge the branch `montaga:parser/exclude-some-dots-from-numbers` into this branch.

I encountered another bug: Dynamically accessing Elements of Lists works only with strings and not numbers:
```
G = {};
G_1 = 3; //works
G_2 = 2; //works
errc(G); //G ===> {1:3, 2:2}
errc(G_1); //gives __
```
This syntax would be really nice (I would love to use json objects as a replacement for the lists in some scenarios). If I replace the key with a string, the example works nicely, but the code would not look that nice:
```
G = {};
G_1 = 3;
G_2 = 2;
errc(G);
errc(G_"1"); // ===> 3
```

File: src/js/libcs/Json.js [hunk 1]
```diff
@@ -0,0 +1,223 @@
...
+            res = evaluate(fct);
+        }
+    } else if (iteratorType === "key") {
+        for (let k in li) {
+            namespace.setvar(runVar, General.string(k));
+            res = evaluate(fct);
+        }
+    } else { // pair
+        for (let k in li) {
+            namespace.setvar(runVar, Json._helper.GenJSONAtom(k, li[k]));
```
montaga: Do we really need JSONAtoms? Because having them, a lot of things would be needed to changed as well. For instance, the code

```
a = {"a":2, "b":3};
forall(a, x, iterator->"pair",
      x.value=x.value^2;
      errc(x.value)
);
``` does give currently `2` and `3`. However, I would expect `4` and `9`.

Would not it be easier if these pair-iterators are just plain JSON-objects them-self with having the keys `key` and `value`? (the only disadvantage I see is that niceprint of "just-JSON-iterators" would look like `{key:a, value:2}`, which is okay IMO because it would help developers to debug.)

**Comment by strobelm** [comment]
Good point, I'll have a look. Thanks!

**Comment by montaga** [comment]
```
forall({"a":2, "b":3}, x, iterator->"pair",
       x.value = x;
      print(x)
);
```
gives a `Maximum call stack size exceeded`.

Why do we need an extra treatment for JSONAtoms at all? (as far as I understand, only the output differs). Aside the bad style, one can write code such as
```
forall({"a":2, "b":3}, x, iterator->"pair",
      x.valuesquared = x.value^2;
      print(x); //does show anything about valuesquared
      print(x.valuesquared) //works fine
);
```

File: ref/Dictionaries.md [hunk 2]
```diff
@@ -1,24 +1,213 @@
...
+    > v = 123;
+    > apply({"a" : 1, "b" : 2}, v, v^2);
+    < {a:1, b:4}
+    > v
+    < 123
+
+------
+
+
+### Applying an expression: `apply(‹dict›,‹varVal›, ‹varKey›,‹expr›)`
```
montaga: could we have apply$3 with ` iterator->"pair"` such as it is done with forall for this purpose?
(anyway, the order first ‹varVal› and then ‹varKey› and then misleading)

strobelm:
Well, since this is a BDFL ruling, I don't want to shake the decision.

I don't see why the ordering is misleading. Actually it should align with

```
apply(‹list›,‹var›,‹expr›)
```
?

montaga:
Ok, now I recaptured the BDLF-decision from February. So let us stick with it. If it is the case that at a later point, the decision is made to have the `iterator` modifier for `apply` and `select`, it can still be easily added without inducing incompatibility.

montaga:
```forall({"a":2, "b":3}, x, iterator->"pair",
       x.key = x.key;
      print(x)
);```
prints some `{[object Object]:2}`.

**Comment by strobelm** [comment]
Well I guess you were right @montaga: the code benefits from removing the json atoms. Thanks again!

**Comment by montaga** [comment]
Thank you, @strobelm, for implementing this huge piece of code. The implementation works fine, and I think we are close to merging this. I also converted an old dictionary example to JSON, and it worked fine as well.

I only had the following observations.

- The following code gives a `Warning: Dictionary printing failed!` and nothing is output:
 ```
  json = {"a":23, "b":true, "c":[1,2,3]};
  print(json, JSON->true);
 ```
- It is not clear to me how I can get an internal JSON-compatible-string within CindyScript? It is a realistic scenario that somebody wants to send a Cindy-JSON object as string though the `javascript` function. Using the CindyScript `json={"a":"bla"}; javascript("json = " + json);` would mean to evoke invalid JavaScript code `json={a:bla}`. What about enabling `json={"a":"bla"}; javascript("json = " + text(json, JSON->true));`?
- JSON objects will induce some conflicts if they are used as keys for now obsolete Dictionary. The `Dict.key` function would also need a check for JSON. I would ignore this problem because I cannot think of a scenario where these two data-structures are used together, and Dict was always marked as a feature that might vanish.
- There is no possibility to delete an existing key from a JSON-object. This could be implemented at a later point in time if needed.

File: src/js/libcs/Json.js [hunk 3]
```diff
@@ -0,0 +1,223 @@
...
+    // track a new recursive call
+    visitedMap.newLevel = true;
+    visitedMap.level += 1;
+
+    var keys = Object.keys(el.value).sort();
+    var jsonString = "{" + keys.map(function(key) {
+        // update visitedMap
+        let elValKey = el.value[key];
+        if (!visitedMap[elValKey]) {
+            visitedMap[elValKey] = 1;
```
montaga: This line is equivalent to
`visitedMap["[object Object]"]=1;`.
Different key-value pair are not distinguished by this approach.
because CindyScript-values are objects and casted to a string when used as a key.

Maybe you just simplify this code and bound the DFS on the total number of visited nodes?

strobelm:
> Thank you, @strobelm, for implementing this huge piece of code. The implementation works fine, and I think we are close to merging this. I also converted an old dictionary example to JSON, and it worked fine as well.
> 
> I only had the following observations.
> 
>     * The following code gives a `Warning: Dictionary printing failed!` and nothing is output:
> 
> 
> ```
>  json = {"a":23, "b":true, "c":[1,2,3]};
>  print(json, JSON->true);
> ```

Because you found a bug, I forgot to hand over the modifs and options to the recursive call.

> 
>     * It is not clear to me how I can get an internal JSON-compatible-string within CindyScript? It is a realistic scenario that somebody wants to send a Cindy-JSON object as string though the `javascript` function. Using the CindyScript `json={"a":"bla"}; javascript("json = " + json);` would mean to evoke invalid JavaScript code `json={a:bla}`. What about enabling `json={"a":"bla"}; javascript("json = " + text(json, JSON->true));`?
> 
This is intended for external usage of the JSON Object: e.g. someone wants to use Cindy generated content within JavaScript.
>     * JSON objects will induce some conflicts if they are used as keys for now obsolete Dictionary. The `Dict.key` function would also need a check for JSON. I would ignore this problem because I cannot think of a scenario where these two data-structures are used together, and Dict was always marked as a feature that might vanish.
> 
I would ignore this, too.
>     * There is no possibility to delete an existing key from a JSON-object. This could be implemented at a later point in time if needed.

Yes, indeed! I already thought about it, but there is no obvious way for me to achieve this. As far as I know the ```:``` operator of Cinderella did not have this either. Or it was an undocumented feature. And there is not user accessible type like ```None``` or ```nana```.

Could @kortenkamp and @richter-gebert comment on how to delete JSON entries?

src/js/libcs/Json.js
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+            return key + ":" + Json._helper.niceprint(elValKey, modifs, options);
+        }
+    }).join(", ") + "}";
+
+
+    // pretty print 
+    // to be valid JSON we need to replace single with double quotes
+    if (options.printValidJSON) {
+        jsonString = jsonString.replace(/'/g, '"');
```
montaga: I am not sure on that. `{"a'":1}` should be valid JSON. However, we should escape double quotes and newlines within the keys/values. They are hard to enter in CindyJS:
```
evilstr1 = unicode("0022");
evilstr2 = "
"
json = {"a":evilstr1, "b":evilstr2};
print(json); 
```
This gives
```
{a:", b:
}
```
valid JSON could be something like `{"a":"\"", "b":"\n"}`. Should we use the `encodeURI()` function for key in line 197 and all printed strings?

montaga: update: better use the `JSON.stringify`-method. It does exactly what we need.
We can apply it either to the plain strings.

montaga:
`WeakMap` seems to be a solution for the one issue :+1: 

---

printing the following JSON-objects still does not work, but should be doable with not too much work:
```
print({"a":[1,2,3]}, JSON->true);
print({"a":"b'c"}, JSON->true);
print({"a'b":"c"}, JSON->true);
print({"a
  b":"c
  d"}, JSON->true);
print({"a":[{"c":"d",2,3]}, JSON->true);
```

In order to fix some escaping problems, I would try to replace `"\"" + key + "\""` with `JSON.stringify(key)` and `"\"" + String(niceprint(a)) + "\""` with `JSON.stringify(niceprint(a))`.

---
> > ```
> > How to use a JSON-compatible-string within CindyScript? It is a realistic scenario that somebody wants to send a Cindy-JSON object as string though the `javascript` function. Using the CindyScript `json={"a":"bla"}; javascript("json = " + json);` would mean to evoke invalid JavaScript code `json={a:bla}`. What about enabling `json={"a":"bla"}; javascript("json = " + text(json, JSON->true));`?
> > ```
> 
> 
> This is intended for external usage of the JSON Object: e.g. someone wants to use Cindy generated content within JavaScript.
> 
For me it is still not clear, how I could use the Cindy generated JSON-objects nicely in JavaScript. With the current implementation, JSON-strings can only be output by using `print[ln](..., JSON->true)`, which uses `csconsole.out`/`console.log`. Certainly, developers should not be asked to overwrite `console.log` in a hackish manner in order to obtain the string in javascript.

At least I would pass `modifs` to `niceprint` within the function `evaluator.text$1`. Then at least my suggested method `javascript("json = " + text(json, JSON->true));` would work. 

A probably nicer way could be to add a `getvalue(<varname>)` to the CindyJS instance, which then returns a (classical) javascript-object with the value of the variable with name <varname>. However, this would involve some work. And how should complex numbers be returned?

---
A proposal for the syntax to deleta a key-value-pairs from a dictionary: What about using the syntax
```
json = {"a":"b"};
json.a = _;
```
This would resemble the syntax for [Undefining functions](https://cindyjs.org/ref/Variables_and_Functions.html#_$28u_$29u$3au$3du$5fu).

src/js/libcs/Json.js
```diff
@@ -0,0 +1,226 @@
...
+
+Json._helper.DeepClone = function(o) {
+    var out, v, key;
+    out = Array.isArray(o) ? [] : {};
+    for (key in o) {
+        v = o[key];
+        out[key] = (typeof v === "object" && v !== null) ? Json._helper.DeepClone(v) : v;
+    }
+    return out;
+};
```
montaga: we do not need this function anymore.

src/js/libcs/Json.js
```diff
@@ -0,0 +1,226 @@
...
+
+        return "\"" + String(niceprint(a)) + "\"";
+    }
+
+    return niceprint(a);
+};
+
+Json.Atomniceprint = function(el) {
+    return "\{" + el.value.key + ":" + niceprint(el.value.value) + "\}";
+};
```
montaga: We do not need this function anymore

kortenkamp:
> Yes, indeed! I already thought about it, but there is no obvious way for me to achieve this. As far as I know the `:` operator of Cinderella did not have this either. Or it was an undocumented feature. And there is not user accessible type like `None` or `nada`.
> 
> Could @kortenkamp and @richter-gebert comment on how to delete JSON entries?
> 

I think that setting the value of a key to `nada` would be a good way to delete it.  And what you say is not true - there is a user accessible type like `None` or `nada`. Try
```
isundefined(nada)
```
which evaluates to `true`.

Of course, if you write `nada="nothing"` then `nada` will be `"nothing"`. But if you don't touch it, `nada` will be the undefined value. 

And also of course: This is fully supported by i18n. Try `nix`, `rien` and `nüscht`.  I can set and unset a dictionary entry like this:
```
A:"key" = "Hallo";
err(A:"key"); 
A:"key" = nüscht;
err(A:"key");
```

Which gives the following output in classic Cindy:
```
A:"key" ==> Hallo
A:"key" ==> ___
```

montaga:
Okay @strobelm , so you removed the valid-JSON-printing and we do not have to debug this feature.

I think aside deleting, we are ready to merge this feature.

Regarding deleting with setting to `nada` in the current implementation:
```
json = {"a":1, "b": 2}
json.a = nada;
forall(json, k, errc(k)); //gives ___ and "b"
errc(keys(json));//gives [a,b]
```
IMO this behaviour should be different.

montaga:
With the last commit I have added an way to delete key-value pairs from a json by setting them to undefined values. The example above now gives

```
json = {"a":1, "b": 2};
json.a = nada;
forall(json, k, errc(k));  //gives 2
errc(keys(json)); //gives [b]
```

As there likely is nobody employed to review this change and this feature is urgently needed for another project, I will merge it into the master. Possible bugfixes on this issue now have to be done by seperate PRs.

strobelm:
🥳

Issue #719, opened 2018-12-16
Font-Size of drawtext("...", size->17) should be according to stage grid (e.g. 0.7) and not to HTML pt


q2apro:
To my understanding the font size is set with pt.

However, this leads to problems when scaling the app / canvas. See here: https://github.com/CindyJS/website/issues/55#issuecomment-447637254

![image](https://user-images.githubusercontent.com/6764548/50053687-08e1dd00-0142-11e9-9a79-3efdc41fa20d.png)

The font size does not scale if you reduce the canvas size (does not matter if CSS or Javascript).

I strongly suggest to change the font size to match the canvas size / grid size to avoid scaling problems.

Or to be able to setup the font-size as percentage (in reference to the defaultAppearance textsize). Maybe that is already possible?!

montaga:
I agree with that. I also would like to have some nice way to specify the font according to the size of the canvas.

q2apro:
My **current workaround** is to specify the font size **only globally** and do not (cannot) touch it in the cindy script itself later on. Which however requires to not change any font-size later on.

See down the `font_size` declaration and the parameter `textsize: font_size,`.

```

	<script>
		var canvas_size = 500; 
		var font_size = 20;
		
		var screen_width = window.innerWidth || window.clientWidth || window.clientWidth;
		
		if(screen_width < 480)
		{
			canvas_size = screen_width;
			font_size = screen_width/27; // 15; // 450 ~ 15, 300 ~ 12
		}
		
		CindyJS({
			scripts: "cs*",
			//autoplay: true,
			ports: [{
				id: "CSCanvas",
				//background: "#EEE",
				axes: false,
				width: canvas_size,
				height: canvas_size,
				transform: [{
					visibleRect: [-1, 7, 14, -7]
				}],
				grid: 1,
			}],
			
			defaultAppearance: {
				/*clip: "none",
				pointColor: [1,0,0],
				lineColor: [0,0,1],
				pointSize: 5,
				lineSize: 1,
				alpha: 1,
				overhangLine: 1,
				overhangSeg: 1,
				dimDependent: 0.7,
				fontFamily: "sans-serif",
				textsize: 20,
				lineHeight: 1.45*/
				textsize: font_size,  // HERE
			},
			
			geometry: [
				{name:"A", kind:"P", type:"Free", pos:[rect.w, 0]},
[...]

PR #759, opened 2019-02-04
Virtual resolution

montaga:
This PR adds virtual resolutions to CindyJS. They should enable an SVG-like upscale of the drawing canvas without adjusting the line-thickness or font sizes. Two optimal parameters `virtualwidth` and `virtualheight` for each port regulate a scaling factor. The idea is that the size of the canvas on the screen can be specified via CSS, Javascript, or a CindyJS parameter. However, CindyJS renderes it as it would have the given virtual resolution.

image

Demand for such a feature exists:
 - for several full-screen applications such as for exhibits of IMAGINARY ( for instance in https://imaginary.github.io/applauncher2/ ) or in the TUM interactive (https://interactive.app.tum.de/web/overview). The devices and screens should be exchangeable without causing additional development effort on the widgets.
- and it could be a more satisfying solution to #719

So far, my implementation is more a proof of concept. The following is still missing: 
- [x] Rescale mouse inputs accordingly 
- [x] the scaling should work together with browser induced scaling
~~- [ ] consider different ratios for scaling in `x` and `y`  directions. Or are there no useful applications at all?~~

However, I do not want to spend more efforts as long as we did not discuss whether we do want this semantic for this feature. I feel that @kortenkamp and @richter-gebert have a strong opinion on that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ js_make: $(NPM_DEP)
 ######################################################################
 
 # Specify build=release on the command line to run closure compiler
-build=release
+build=debug
 
 build/js/Cindy.plain.js: js_make
 	$(JS_MAKE) plain
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ js_make: $(NPM_DEP)
 ######################################################################
 
 # Specify build=release on the command line to run closure compiler
-build=debug
+build=release
 
 build/js/Cindy.plain.js: js_make
 	$(JS_MAKE) plain
--- a/examples/163_virtualresolution.html
+++ b/examples/163_virtualresolution.html
@@ -1,0 +1,77 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+        <title>Cindy JS</title>
+        <script type="text/javascript" src="../build/js/Cindy.js"></script>
+        <link rel="stylesheet" href="../css/cindy.css">
+    </head>
+
+	<body style="font-family:Arial;">
+
+        <h1>CindyJS: Virtual resolutions</h1>
+
+        <script id='csdraw' type='text/x-cindyscript'>
+						fillcircle((0,0),5,color->[1,1,.7]);
+						fillcircle((0,0),1,color->[1,.7,.7]);
+						drawtext((0,.9),"test1");
+            drawtext((0,0),"test2");
+						draw((1,-100),(1,100));
+						draw((0,0.12),(2,0.12));
+						
+						draw(A,B);
+						N = 100;
+						zigzag = apply(1..N, k, (2-(k/N)^2,.5+.4*(-1)^k));
+						connect(zigzag, size->1);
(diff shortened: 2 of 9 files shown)


montaga:
Note before merging:
if https://github.com/CindyJS/CindyJS/pull/760 has been merged before, then also the touch inputs have to be rescaled with `vscale` accordingly.

kortenkamp:
I still wonder whether all this is the right approach. We do have an independent coordinate system (the homogeneous coordinates) which is mapped to screen coordinates through a transformation. Why do we need another transformation behind that?

Sorry to be over critical here, but this is just another level of complexity that will make it more complex to continue development.

montaga:
Thank you for this comment. I am also not totaly convinced with this, but I do not know how to make it better since we have calculations both in the space of homogeneous coordinates as well as in the coordinate space, where both have to scale.
With commit b15f8eb, I shifted this rescaling to `csport.to`, so there wont be any additional  development process should be more simple for additional input devices. This are less lines of codes, but the logic is becoming even more inconsistent.

In total we are mapping between 4 different spaces:

A: The homogeneous coordinates.
B: The space of virtual pixels.
C: CSS pixels.
D: Physical pixels.

For drawing, we are going from top to bottom. For input events, we go from bottom to top.

In an ideal and simple world, we would have the identity between B, C, and D.

For modern devices, we need something between C and D. This is already done via a `csctx.scale`. Luckily, the browser does the reverse map from D to C for us; so mouse coordinates do not have to be rescaled.

Before this PR there was no space B and `csport.from` and `csport.to` mapped between A and C.

What makes everything difficult is that there are a lot of computations in space C, for instance `size`, `thickness`, which are specified in pixels.  If we would try to scale all these operations, the changes would be more invasive and occur at many places.
So I introduced space B, where these operations are done. And after that, these virtual pixels are scaled to CSS pixels via some canvas-operation. In the implementation before your comment, `csport.from` and `csport.to` mapped between A and B. With b15f8eb I modified `csport.from` and `csport.to` such that they map between `A` and `C`, whereas `csport.matrix` still maps between `A` and `B`. I am only semi-happy with this. The advantage is that the implementation of new input devices is simpler because only `csport.to` can be used as before, but I do not like the different meaning "dissonance" between `csport.to`  and `csport.matrix`. But we cannot forget about space B because a lot of computations are done there.

montaga:
We still need to agree about whether some stretching changing the ratio should be allowed through virtual resolution. Implementing this is no problem.
However, I feel like that this ``feature'' would only be used by accident and never on purpose.

kortenkamp:
Please do not allow stretching in either direction through virtual resolution. This should not be necessary to adapt to strange displays nowadays (we can assume square pixels!), and it would help people to misuse the coordinate system for other effects. With other software not named here it leads to confusion all the time didactically.

strobelm:
Code looks good and works well for me. I really lake the  `csctx.setTransform`, wasn't aware of that.
Merged.

PR #760, opened 2019-02-05
Add ability to script multitouch events

montaga:
Due to an exhibition, we need to be able to handle multitouch input events. This PR adds the scripts `multidown`, `multiup` and `multidrag`. Furthermore with

- `multiid()` the id of the current touch event is obtained
- `multiidlist()` a list of all active touch event ids is obtained
- `mouse(id->number)` the position of the touch with the given id can be obtained.

image

This PR does not introduce `mtregional` or `mtlocal`. Furthermore, no multitouch interaction with geometry is enabled, because this might be complicated due to tracing.

It would be good that we could agree on the names of the functions/scripts soon so that it can be merged without losing the possibility to implement a broader multi-touch support later. Closer to the javascript like convention would be to have `multi` replaced with `touch` and `drag` replaced with `move`.

### DIFF
--- a/examples/multitouch.html
+++ b/examples/multitouch.html
@@ -22,7 +22,7 @@ drawtext((0,0), multiidlist());
 
 al = 1;
 forall(events, e,
-  dim = if(e_3=="down" % e_3=="up", .7, 1);
+  dim = if(e_3=="down" % e_3=="up", drawtext(e_1, e_3 + " of " + e_2, alpha->al);.7, 1);
   draw(e_1, color->hue(e_2/sqrt(20))*dim, alpha->al);
   al = al*.95;
 );
--- a/examples/multitouch.html
+++ b/examples/multitouch.html
@@ -1,0 +1,70 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Multitouch scripting</title>
+    <script type="text/javascript" src="../build/js/Cindy.js"></script>
+<script id="csinit" type="text/x-cindyscript">
+events = [];
+
+addevent(pos, id, kind) := (
+  events = [pos, id, kind] <: events;
+  if(length(events)>=200,
+    events = events_(1..200);
+  );
+);
+</script>
+
+
+<script id="csdraw" type="text/x-cindyscript">
+drawtext((0,0), multiidlist());
+
+al = 1;
+forall(events, e,
+  dim = if(e_3=="down" % e_3=="up", .7, 1);
+  draw(e_1, color->hue(e_2/sqrt(20))*dim, alpha->al);
+  al = al*.95;
+);
+
+</script>
+<script id="csmultidown" type="text/x-cindyscript">
+addevent(mouse(id->multiid()), multiid(), "down");
+</script>
+
+<script id="csmultidrag" type="text/x-cindyscript">
+addevent(mouse(id->multiid()), multiid(), "draw");
+</script>
+
+<script id="csmultiup" type="text/x-cindyscript">
+addevent(mouse(id->multiid()), multiid(), "up");
+</script>
+
+    <script type="text/javascript">
+var cdy = CindyJS({
+  scripts: "cs*",
+  defaultAppearance: {
+    dimDependent: 0.7,
+    fontFamily: "sans-serif",
+    lineSize: 1,
+    pointSize: 5.0,
+    textsize: 12.0
+  },
+  angleUnit: "°",
+  ports: [{
+    id: "CSCanvas",
+    width: 680,
+    height: 350,
+    transform: [{visibleRect: [-9.06, 9.34, 18.14, -4.66]}],
+    background: "rgb(168,176,192)"
+  }],
+  csconsole: false,
+  autoplay: true,
+  cinderella: {build: 1901, version: [2, 9, 1901]}
+});
+    </script>
+</head>
+<body>
+    <div id="CSCanvas"></div>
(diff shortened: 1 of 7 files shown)


kortenkamp:
Should that be backported to Classic?

montaga:
I see the current primary application for multi-touch are exhibitions and touch-devices, on which right now only CindyJS runs.
If these functions are available within Cinderella, this will make the export easier.

strobelm:
I don't see any grave problems with this PR. Perhaps you could replace the `let` declarations with `var` (IE 11 will be grateful).

montaga:
In `build/js/Cindy.js` the string "let " does not appear anymore due to the closure trans-compilation :-)

I am just wondering whether it would be nicer to force the ids to start with 1 instead with 0, to be more coherent with CindyJS lists.  Unfortunately, I cannot find in the specification how the identifiers should behave, aside from that they are long and unique ( https://w3c.github.io/touch-events/#dom-touch-identifier )

kortenkamp:
As the ids are arbitrary I think it makes no sense to force a certain start value. Actually there should be no assumptions whatsoever for them on the users side.

montaga:
After some discussion today, we changed the behavior in two ways:

- The mouse itself will trigger `multidown`, `multiup` and `multimove` events. The mouse will always have identifier zero. This makes the development process to enable multi-touch easier: All `mouse`-scripts can be replaced by corresponding `multi`-events. But still, they can be used on a device with a mouse. For each event, only one script is evaluated (when combining `mouse` and `multi` scripts, then some touch-events will trigger both scripts, which is a restriction due to backward compatibility)

- CindyJS now manages the multiids. The mouse will always have `multiid()==0`. Each touch event is attached to a unique ID that is the smallest positive integer that is not used for another touch event. Having small positive integer makes it easy to combine the input with CindyJS lists.

kortenkamp:
So, just to be clear here: the distinction between the multi and single things is that we guarantee the down-move-up sequence for the old mouse events, while this is not guaranteed for the multi version. 

But it is still guaranteed for each id. 

Question: is anybody willing to implement mtlocal? It is really convenient to just have local variables for each touch.

montaga:
`mtlocal`/`mtregional` would require some non-trivial work in `Namespace.js`.

Right now I have other priorities than implementing this.

I have only a vague idea about how they should work. Could you please explain it? I only saw the example at https://www.researchgate.net/publication/261823421_Interaktives_Whiteboard_iPad_Co_--_das_Klassenzimmer_der_Zukunft


I assume that `mtlocal` is supposed to be specified at the beginning of the `multidown` script. This declares that the variable has some local meaning in the following `multidrag` and `multiup` scripts for the touch with the same identifier?
Would it be more consistent to CindyJS to call it `mtregional`?
Would be there some semantic to access these local-vatiables within the draw-script (and a given identifier)?

kortenkamp:
For the classic implementation I just did the same that local() was doing (might be called regional by now), it was quite easy. And it just makes sure that you have a variable that can be used in multitouch events that behaves like a variable in single touch, taking care of switching the variable namespace automagically. 

Should be documented, btw.

montaga:
This PR is now ready to be reviewed and merged.

montaga:
Update: Apparently, Edge and thus surface with pointer events is not supported.

strobelm:
Any chance to get this working? Would be great to have, especially if we
think about the large Microsoft touch systems in Garching.

montaga:
Safari on iOS currently does not support PointerEvents, so we would need to maintain an implementation for both TouchEvents and PointerEvents.

As I do not really have other priorities right now, I do not want ti implement a dual pipelines for PointerEvent, unless there is actual demand for some exhibitions i near future. Edge is going to switch from edgeHTML to blink, so this problem will likely solve by itself after some time.

kortenkamp:
Is there any chance to distinguish between Pencil and finger on iOS?

montaga:
Yes, it apparently it should be (https://stackoverflow.com/questions/34986373/javascript-touch-event-distinguishing-finger-vs-apple-pencil). Once needed, we could introduce a function `touchtype()` and `tochetype(id->...)`. Also the force is accessible.

strobelm:
Hi! I had some issues with the PR:
Firstly, sometimes i get a JS error:

```Events.js:477 [Intervention] Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.```

And secondly, sometimes the touch up seems to hang: although i removed all hands from the screen there is still an element shown in the array :-(. It's not easy to reproduce, though. It can probably provoked using 10 fingers.

montaga:
I think the first problem occurs because we do not consider the recommendation of using passive listeners https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md (and it is likely that the problem also existed before this PR)
So far we also cannot do, because we are blocking scrolling through `touchmove`-events via `preventDefault()`. Honestly, I am also not really happy with this behavior on mobile devices, because this makes several pages with CindyJS applets such as cindyjs.org inconvenient to navigate on a mobile phone. We could try to change this behavior, but then we would lose the ability to use touch-events to interact with the widgets if they are put on a scroll-able page. In some other PR, we could add a global option that could make the touch-event listeners passive.

I cannot reproduce the second issue :( Maybe the order of the coming events sometimes is wrong. The last commit should make the organization of `multipos` more robust against swaps in the order. Could you please try and report whether the second issue still occurs.

strobelm:
I could not reproduce the issues :+1:

PR #781, opened 2019-04-08
enhance compatibility OpSound

montaga:
With this, OpSound works again on Safari on iOS.


strobelm:
Now that's a weird bug.

PR #788, opened 2019-04-26
proper multi-touch for more than 10 fingers

montaga:
This might have been the reason why some applets showed strange behavior  when the entire arm has been used for multi-touch.


strobelm:
I will never understand why the default sorting strategy is lexicographical 🙄

Am 26. April 2019 16:04:10 MESZ schrieb Aaron Montag <notifications@github.com>:

PR #789, opened 2019-04-26
Op sound amp fix

pdywilson:
small fix for smoother continuous amp changes of masterGain.


montaga:
Thank you. With this, adjusting amp works.
However, it breaks the modification parameters of a damped tone as in
`examples/sound/simplepiano.html`
because the `masterGain.gain.value` also encodes the damping. We do not want to reset the damping if we want, for instance, change the harmonics. Maybe, it helps to replace

```
curline.handleModif(modifs, "amp", 'number', curline.amp);
```
with something like
```
curline.handleModif(modifs, "amp", 'number', curline.masterGain.gain.value);
```
in playsin$0.

montaga:
In order to make travis-ci work, you should merge the current master again into your branch.

pdywilson:
Ah I see. Well, we also want the functionality for playsin$1 really, so how about a dampNode (gainNode) after the masterGainNode so the parameters can be adjusted independently? Or would that be too much overhead?

I am also wondering what 

        this.masterGain.gain.setValueAtTime(this.masterGain.gain.value, this.audioCtx.currentTime);

does? It seems redundant to me, but maybe there is a reason for it i am not aware of.

montaga:
> Ah I see. Well, we also want the functionality for playsin$1 really, so how about a dampNode (gainNode) after the masterGainNode so the parameters can be adjusted independently? Or would that be too much overhead?

Introducing a separate dampingNode after the masterGain sounds reasonable and cleaner alternative to me. Similar as the panning node, it could also be optional and used only if damp!=0.
> 
> I am also wondering what
> 
> ```
>     this.masterGain.gain.setValueAtTime(this.masterGain.gain.value, this.audioCtx.currentTime);
> ```
> does? It seems redundant to me, but maybe there is a reason for it i am not aware of.

Likely it is redundant. In the line before, I wanted to have something as [cancelAndHoldAtTime](https://developer.mozilla.org/en-US/docs/Web/API/AudioParam/cancelAndHoldAtTime), which is not supported by many browsers.

src/js/libcs/OpSound.js
```diff
@@ -132,8 +132,7 @@ var OpSound = {
         let audioCtx = this.getAudioContext();
         let gainNode = audioCtx.createGain();
         gainNode.gain.value = 0;
-        oscNode.connect(gainNode);
-        gainNode.connect(masterGain);
+        oscNode.connect(gainNode).connect(masterGain);
```
montaga: chained `connect` fails on Safari and thus should be avoided :(

src/js/libcs/OpSound.js
```diff
@@ -278,12 +277,29 @@ class OscillatorLine {
...
-            this.masterGain.gain.setTargetAtTime(1, this.audioCtx.currentTime + this.attack, (-this.damp));
+        //insert dampNode in between if necessary
+        if (this.damp !== 0) {
+            if (!this.masterGain.dampNode) {
+                if (this.masterGain.panNode) {
+                    this.masterGain.disconnect(this.masterGain.panNode);
+                    this.masterGain.dampNode = this.audioCtx.createGain();
+                    this.masterGain.connect(this.masterGain.dampNode);
+                    this.masterGain.dampNode.connect(this.masterGain.panNode);
+                    this.masterGain.panNode.pan.value = this.pan;
```
montaga: panning is already done in `panit`

pdywilson:
I thought about it for a while, but i can't come up with a more elegant way, we have to keep panit and dampit separate to cover all use cases, and we also can't incorporate it into the createNewMasterGain, incase it is only later damped and panned. So the way i see it we have to stick with this unelegant way.

montaga:
I feel that the damp-node should also be taken into consideration in  `panit()`. When the `panNode` is inserted between `materGain` and `audioCtx.destination`, then a possible previously generated `dampNode` might be missed in our chain.
I would add a seperate function `updateChain(withDampNode, withPanningNode)`, which fixes the chain as it is currently done in `dampit()`. Then a `updateChain(this.masterGain.panNode || this.pan, this.masterGain.panNode || this.damp)` in `evokeplaysin` could do the job for us.

pdywilson:
I implemented the requested changes together with an example in OpSoundExamples where the transition between different chaining options of pan and damp nodes can be tested.

PR #793, opened 2019-05-03
Text at angle

montaga:
This is based on #729 and travis CI should hopefully be happy with that.


strobelm:
Looks good and works fine. Thanks @DivaStudent and @montaga

PR #795, opened 2019-05-30
TypeScript support

strobelm:
Hi! 

This PR adds support for TypeScript (TS). 

TS is a superset of JavaScript that adds a lot of useful (everything is optional) features, foremost typing, to JS. Then the whole source code is transpiled to plain JS. 

Type annotations are much more concise than using the closure compiler:

```
// TypeScript
function greeter(person: string) {
    return "Hello, " + person;
}

// Closure Compiler
/**
 * @param {string} person
 */
function greeter(person) {
    return "Hello, " + person;
}
``` 

One can benefit from compiler warnings and errors like this:
```
src/ts/example.ts(3,61): error TS2345: Argument of type 'string' is not assignable to parameter of type 'number'.
```

The 5 Minute intro to TS shows some of the features: https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html

I'm not sure if we are willing to adopt this, but at least I wanted to show that there is quite an easy way to get typing support.

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -5,6 +5,7 @@ var glob = require("glob");
 var path = require("path");
 var Q = require("q");
 var qfs = require("q-io/fs");
+var tsc = require("typescript-compiler");
 
 var getversion = require("./getversion");
 var src = require("./sources");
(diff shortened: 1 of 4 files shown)


strobelm:
The node package typescript-compiler seems to be unmaintained now. I'm closing this to prep an updated version with a different package.

PR #804, opened 2019-07-05
Buttontext color

pdywilson:
In http://www.mathe-vital.de/Physik/7-1.html the two ToggleButtons Strahlen and Ränder should have white text specified by the `color: [1.0, 1.0, 1.0]` attribute.

This PR implements this behaviour for Buttons together with the `alpha` attribute.


montaga:
This looks fine. Thank you for this contribution :-)

Issue #668, opened 2018-06-14
The attributes `fillcolor` and `fillalpha` and `alpha` for "Text" are not supported yet


montaga:
A geoobject
`{name: "Text0'", type: "Text", color: [0.0, 0.0, 0.0], fillcolor: [0.012, 0.655, 0.737], fillalpha: 0.20000000298023224, alpha: 0.8999999761581421, text: "Some Text", textsize: 16.0, dock: {corner: "UL", offset: [10.0, -40.0]}},`
will ignore the attrubutes `fillcolor`, `fillalpha` and `alpha`. However, these attributes seem to be supported for elements of type "EditableText" (they are also rendered as HTML object instead of being drawn on the canvas)

PR #805, opened 2019-07-05
Changed geoOps.Text to a html element and refactored code

pdywilson:
#668 

This changes how text is rendered so that it is possible to use and fillcolor, fillalpha for text too. I dont know if this is the ideal way to do it, but it is one possibility. Currently Text can be moved around by the mouse, e.g. in https://cindyjs.org/examples/Buttons2.html this is not possible the way I implemented this now. An example fillcolored text is given in the attached Buttons2.html.

This is also related to #804.


kortenkamp:
Before this is merged somebody has to check whether LaTeX rendering still works and also whether this introduces any incompatibilities with classic Cindy. Sorry, no time for that right now so I cannot do it myself, but I wanted to warn about that..,

montaga:
LaTeX renders fine with this implementation :-)

Could you please add something as `pointer-events: none;` to the CSS? Right now, the text-selection courser appears if the mouse hovers the HTML-texts and mouse-events for the canvas beyond the text are stolen.

kortenkamp:
Another question – what happens if I use a text like "<b>Hello</b>"? Will this be rendered verbatim or will it interpret the HTML?

PR #816, opened 2019-11-12
Fix snap distance

strobelm:
Well this is a bit embarrassing, but nevertheless: the check for snap distance contains a bug:

In order to check if ```port.snapdistance``` exists, the code evaluated ```port.snapdistance``` which is correct of it would be an object, but this is of type number and therefore 0 would evaluate to false, which is unintended.

I really hope that https://github.com/tc39/proposal-optional-chaining is coming soon, to prevent such bugs.

Sorry for that!

### DIFF
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -269,7 +269,7 @@ function createCindyNow() {
                 cstgrid = port.tgrid;
             if (port.snap)
                 cssnap = true;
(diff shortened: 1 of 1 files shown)


montaga:
of course, you are right! That is not the first time that I stumbled into this. Thank you for double checking!

PR #817, opened 2020-01-28
Add new plug-ins CindyPrint, CindyLeap, CindyXR

chrismile:
This PR adds the following three plug-ins to CindyJS:
- CindyPrint: A plug-in for printing 3D objects from Cindy3D and CindyGL using a 3D printer. CindyPrint can export 3D meshes from scenes that can be directly used as an input for 3D printer slicer software.
- CindyLeap: A plug-in that allows CindyJS developers to use the tracking information from Leap Motion controllers for hand and finger tracking.
- CindyXR: A plug-in that allows Cindy3D and CindyGL developers to use the WebXR standard for creating VR and AR applications. WebXR supports a wide variety of devices, including smartphone VR using cardboard boxes and VR headsets with head and controller tracking.

The plug-ins allow users to be more creative when creating Cindy3D and CindyGL apps.
Each plug-in is accompanied by additional examples and a reference documentation.
The integration into CindyJS is minimally invasive and our aim was to export as few symbols as possible into the global scope (thanks @montaga for your feedback!). But still, some changes needed to be done especially in Cindy3D and CindyGL. In core CindyJS, the only change is that in Operators.js scheduleUpdate is exposed now.

I'm looking forward to your feedback! Please tell me if you think something should be changed.


kortenkamp:
This sounds great, and I am eager to have a look at it. 

Quick question: is the interface to the LEAP controller backwards compatible to the old LEAP interface in Cinderella?

Side note: I hope that we can re-introduce regular meetings that avoid being surprised by such additions, or at least some kind of announcements that tell the Potsdam group that there is something in the pipeline...

chrismile:
I can understand that this was surprising ;)

CindyLeap is using a new API. It uses the JSON support added recently in CindyJS to directly give the users the data the Leap Motion JavaScript API hands to the application. This way, all features (finger tracking, gesture support, ...) can be used within CindyScript easily. The examples and the documentation should give a good overview of how to use the data, and the Leap Motion JavaScript documentation is also in most parts valid for CindyScript.

montaga:
Thank you @chrismile for this huge contribution and this diligent work! (Stefan Haas should also be mentioned here for his contributions to CindyPrint!). The plugins work all well, their code looks well and are well documented. All "invasive operations" to the rest of the project look reasonable to me and I can not spot any unwanted side-effects.

I made made a tiny pull-request to your branch that also the images within the documentation are copied during a `make ref`.

The LEAP control of cindyleap certainly is different from the undocumented LEAPPlugin for Cinderella (Which can be found on https://cinderella.de/tiki-index.php?page=Plugins ). I think this should not keep us from merging these plugins here, because the plugins have distinct names and plugins remain exchangeable by the user and demand.

PR #818, opened 2020-03-05
No es7-shim due to bad performance

montaga:
After being unhappy with the performance of a page that used several CindyJS instances in iframes, I did some performance profiling, and I realized that initializing the `es7-shim`, introduced in #672, takes more than 100ms on a modern machine running Chromium v80. On mobile devices, this overhead might be even higher. Dropping the `es7-shim`, the total initialization time for small CindyJS-applets drops to less than 50ms.
I feel we should avoid this dependence as I have the impression that we hardly used any ES7 features.
So far I have replaced
* all occurrences of `Array.prototype.includes`, which is an ES7 feature, with `indexOf(...)!=-1`.

Did I miss something? Did we use ES7 features at some place that we could easily replace with ES6?

@strobelm Please let me know if you do not have time for reviewing anymore.

### DIFF
--- a/editor/export.js
+++ b/editor/export.js
@@ -116,7 +116,7 @@ var Export = {
         </style>
         <link rel="stylesheet" href="https://cindyjs.org/dist/latest/CindyJS.css">
         <script type="text/javascript" src="https://cindyjs.org/dist/latest/Cindy.js"></script>
-        ${(configuration.use.indexOf("CindyGL")!=-1) ? '<script type="text/javascript" src="https://cindyjs.org/dist/latest/CindyGL.js"></script>' : ''}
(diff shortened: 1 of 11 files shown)


strobelm:
Hey! That's of course very unfortunate.

@montaga: you're right. The only feature we are really using is 
Array.prototype.includes. So the changes are fine and can be merged.

As an alternative I suggest to have a look at

https://github.com/zloirock/core-js

montaga:
Thank you for your review. :+1: 
And thank you for pointing out `core-js`. We would need to pipe it through browserify or google-closure in order to include it. I am afraid to further bloat our library by including it unless there is urgent demand. The odds are hight that core-js will induce similar losses of time during initialization as es7-shim.

PR #820, opened 2020-03-28
Fix json precedence

### LINKED ISSUE #819: Initialize dictionary/JSON values by expression; {"a": 1+1} does not work
Right now it seems not possible to initialize a Dictoionary/JSON object with user given expressions. For instance, the script

```
errc({"a": 1+1})
```
gives an error
```
Error: JSON keys have to be strings.
===> {}
```

strobelm:
Fixes #819. New precedence should still be checked by someone.

Best,
Michael

### DIFF
--- a/src/js/libcs/Json.js
+++ b/src/js/libcs/Json.js
@@ -37,8 +37,8 @@ Json.setField = function(where, field, what) {
 
 Json.GenFromUserDataEl = function(el) {
     // key/obj are reversed due to the semantics of the ":" operator in CindyScript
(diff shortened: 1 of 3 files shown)


montaga:
Thank you for going into this and still contributing to the project.

Your PR fixes #819, however it breaks the column-operator to attach user-data to geometric elements:

```
A:"a"=41;
A:"a1"=100;
errc(A:"a"+1); //gives 100, and not 42
```

strobelm:
This is now superseded.

PR #821, opened 2020-03-29
WIP: Code base modernization (ES5 -> ES6)

strobelm:
Hi all,

our code base relies heavily on ES5 which us quite dated nowadays (time passes by quickly). There are quite some handy constructions in ES6 that make the life of JS development much easier (see for example http://es6-features.org/). Therefore, I propose to move even more to ES6 than we did before. Fortunately, the conversion can be done automatically most of the time. This PR shows work of around 2 hours, including bug fixing and testing.

This is explicitly marked as WIP and should be discussed first. I don't see many drawbacks, but a lot of benefits here. Since we are transpiling back to ES5 anyway, we don't even loose compatibility to IE11.


kortenkamp:
I think it is a good idea to do that. Actually - is IE11 compatibility worth the trouble of transpiling? Or, differently stated: Are we losing anything by the transpilation in terms of memory or cpu efficiency? If so, we should reconsider support for IE11.

strobelm:
Well, the answer is, like most of the time, 'jein'. 

Since modern browsers use JIT compilation on JS code, the answer should be given by benchmarks we could conduct. 

There are constructions in ES6 that are powerful, but also expensive. But these features would not be introduced with this PR. There is a nice comparison (although not including closure compiler that we use) here: http://incaseofstairs.com/six-speed/  btw.

When we discussed this topic some years ago, we settled the argument of using native ES6 with 'no', we have to support IE11, because it is still widely used in schools. Later we introduced the transpilation for that.

Sure, the logic kinda weird: convert ES5 to ES6 and then transpile it back to ES5. Currently, I don't believe that transpilation really imposes performance penalties (at least not for the code we are using).

montaga:
If I see this correctly, this PR is more or less syntactic sugar and more a matter of taste. Closure will transcompile it back to ES5. My personal taste: I am in favor of `const` and `let` instead of `var`, but functions can remain to be called `function` as long as they are not anonymous. 

At this point, I do not see reasons why we can use some ES6 features here and there (also outside this PR) unless they have terrible performance.

The code looks good at some random positions. I guess you used a script or regexes to convert these files, and it might be easier to trust the conversion process than check each of the almost 6000 changed lines.

We should keep in mind that merging this PR might force us to rebase other open PRs by hand. If we agree that we now like to see this style more often and this conversation process runs almost automatically, it might save us some workload if we rerun this conversation at a later point once only a few PRs are open (?)

strobelm:
Sure, I did not do all the conversions by hand :). I used a project called `lebab` (https://github.com/lebab/lebab) and the process is mostly automatic. I fixed changed some `const` to `let` here and there. This can be redone at any time.

By the way: arrow functions are in fact syntactic sugar most of the time, except for if `this` is involved. `function` does not bind the `this` pointer of the outer scope automatically, while arrow functions do.

After most of the PR are done I would redo the process (perhaps without arrow functions, I don't have strong preference here, but I am a strong believer in `const` and `let`.)

PR #832, opened 2020-06-19
Use forked version of leapjs

montaga:
The forked version of leapjs is compatible with the recent version of
nodejs. Thank @chrismile for pointing out this.


montaga:
This is superseded by #833

PR #833, opened 2020-06-19
Add event API to CindyXR

chrismile:
In the commits attached to this PR, an event API for CindyXR is added.
The examples have been refactored for the use of the new API and one new example was added.
Furthermore, some code was moved from CindyXR.js to WebXRHelper.js.
src/js/libcs/Operators.js now exports the function 'labelCode' in order to enable plugins like CindyXR to compile plugin-specific event handler scripts. This was necessary, as plugin-specific scripts should of course not be hard-coded in the core CindyJS code.
'leapjs' in package.json is now referencing a fixed fork of leapjs (cf. pull request #832).

@camden-bock has been able to start testing the new event API in a project by @maine-imre, and so far, the new API seems to offer exactly the flexibility necessary for creating interactive CindyXR applications.


camden-bock:
@chrismile Thank you so much for your work on this!

montaga:
Thank you for this contribution!

Could you please transfer the scripts via the API instead of loading them from the DOM via `document.getElementById(scriptName)`?

Then it would be possible to specify the scripts also via
```
CindyJS({
  scripts: {xrdraw: `print("your code....");`, ...}
  ...
})
```

This pattern is sometimes useful, for instance, to embed cindyscript-code from external files.

You could export the object `cscompiled` (see https://github.com/CindyJS/CindyJS/blob/master/src/js/Setup.js#L350 ) via the plugin API ( https://github.com/CindyJS/CindyJS/blob/master/src/js/libcs/Operators.js#L4419 ). The XR scripts can be added to https://github.com/CindyJS/CindyJS/blob/master/src/js/Setup.js#L312.
Furthermore, with this change, there would be no need to export the function `labelCode` via the API.

chrismile:
Thanks @montaga for your suggestions. I've switched to using cscompiled (see #
e2238dc). An disadvantage/advantage (depends on how one wants to view this) is that 'cs' (or similar, depending on scripts pattern) now also needs to be prepended to the xr* event names (see examples). Should we stick with this? While I liked the previous, shorter names more, I guess that this is more in line with how Cindy is handling scripts so far.

montaga:
Thank you :)

Too me it sound more consistent to keep it in that way you have implemented it in e2238dc .

If CindyJS is initialized via
```
CindyJS({scripts: "*",...});
```
instead of
```
CindyJS({scripts: "cs*",...});
```
then the cs-prefix is not needed at all.
If you like, you might change the cindyxr-examples to use scripts without the cs-prefix to increase readability, but it is not that important.

chrismile:
Thanks for your suggestion! I changed it now to just use a single wildcard character (i.e., '\*') in the CindyXR examples. I also fixed a bug in Setup.js. A wildcard character at the beginning of the string was not accepted so far. I tested both the CindyXR examples with script pattern '\*' and other examples with 'cs\*', and both should work now.

montaga:
I was not aware that the pattern `"*"` has not been accepted. Thank you for fixing the issue in Setus.js.

This PR makes CindyXR much more usable :)

PR #836, opened 2020-09-11
Enable setting of Texts

kortenkamp:
A student complained that CindyJS is not compatible with Cinderella when setting .text attributes of texts. This should enable setting this attribute.


strobelm:
Hi! Thanks for implementing this.

PR #837, opened 2020-11-04
Fix download of text files in CindyPrint

chrismile:
`downloadTextFile` in plugins/cindyprint/src/js/DownloadFile.js fails to pass the file name parameter to `downloadBlobAsFile`. This PR fixes the problem.
It seems like I didn't test this well, as CindyPrint by default uses binary meshes instead of textual representations in its final version (due to smaller sizes and more performant parsing of binary meshes).


strobelm:
Hi. Thanks for fixing this.

PR #840, opened 2020-12-27
Code formatter: prettier

strobelm:
Hi, I am currently working on improving our tool chain. The first step is the code formatter.

With this PR we switched to prettier, which handles a lot more file types than js-beautify (what we used before), for example the HTML or Markdown files. Prettier found already a lot of HTML files that contained invalid code.

Furthermore, the code formatting is now done with a git commit hook and only operates on staged files, so it is much faster than before. I disabled the test for beautification, which should not be necessary anymore.


kortenkamp:
Should I check and merge or is the assignment to @montaga on purpose?

strobelm:
> Should I check and merge or is the assignment to @montaga on purpose?

No, this was not on purpose at all (@montaga was just the first one who github suggested). I am super happy if one of the BDFLs has a look :+1: 

I must admit that I have no idea who is currently on the university payroll to work on Cindy.

PR #842, opened 2021-01-09
Eslint

strobelm:
Hi, as promised there here is the merge request that adds `eslint` and removes `jshint` for linting our JavaScript files.

Eslint is much more configurable and comes with a lot of nice plugins that can be adapted to ours needs. Furthermore, it seems to get along with our code base better than jshint: I could remove a lot of comments that were turned off some features of jshint within our code. Additionally, the IDE support for eslint seems to be much better, so you should get linting results out of the box while coding.

Also I added a pre-commit hook, s.t. you cannot commit code that fails the eslint check (can be bypassed with `git commit -m "my awesome commit" --no-verify`).

Since we are not using imports/exports to compile our code base (yet) the checks are done both on the source files and the compiled code. Hence, I also wrote a reporter plugin that backrefs to the original source file if errors were found in the compiled one (like we had for jshint).

This MR is based on #841 and should be merged after imports/exports.


kortenkamp (approved):
I did a quick check and this makes sense to me.

PR #843, opened 2021-02-02
Fix elements at mouse

kortenkamp:
elementsatmouse was not working, as the function to determine whether a segment is hit by the mouse did not work without el.farpoint being set for segments. This fixes it.


kortenkamp:
Obsoleted by #845

PR #848, opened 2021-02-10
fix undefined image size

### LINKED ISSUE #847: initscript is executed before images have been loaded
What is says in the title: The function `imagesize("image name")` returns `[NaN, NaN]`.

strobelm:
Hi, this is untested but might fix #847.


kortenkamp:
I don’t see how this fixes #847 — the original problem is indeed that images are not loaded before csinit. Your fix looks like it replaces the NaN pair with Nada, which is also good, but not fixing the original cause?

kortenkamp:
I merge this as it makes sure that a non-loaded image will not cause a hiccup, but only a nada.

PR #855, opened 2021-07-29
update npm packages

strobelm:
This PR upgrades a lot of npm packages to the latest version. Fixed some breaking changes during the process.


kortenkamp:
Well, thank you for this update.

IMHO, CindyJS depends on too many other libraries, a common problem in the node/JS world.

It would be great to find out which of the packages are

- necessary for the build process
- necessary for automatic testing
- necessary for creating documentation
- necessary for creating the website
- actually used by Cindy.js 
- used by plugins
- …

Is it realistic/easy to find this out?

PR #858, opened 2021-07-31
remove some ununsed files

strobelm:
Mainly to test the migration to travis-ci.com


kortenkamp:
Shall I merge now? Probably…

strobelm:
Looks good from the CI. Let us see if the CD also kicks in.

kortenkamp:
Well. These were files that were created by IDEA, and I was the one who used them, and probably the only one. So it’s fine to remove them.

PR #860, opened 2021-11-16
assign code to user fields of objects and lists

kortenkamp:
This commit should provide compatibility between CindyJS and Cinderella. In Cinderella it is possible to assign code not only to a variable, but also to a user field:
```
A:"draw" := draw(A,A+(1,1));
```
This code can be called using 
```
A:"draw";
````

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -54,3 +54,4 @@ examples/162_sunflower_crlf.html  eol=crlf
 
 .prettierignore text
 .eslintignore text
+pre-commit	text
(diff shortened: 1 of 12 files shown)


strobelm:
Hi! Any reason why we cannot call `evaluateAndVal`? Then we could simplify the code a bit.

strobelm:
And thanks for implementing this!

kortenkamp:
> Hi! Any reason why we cannot call `evaluateAndVal`? Then we could simplify the code a bit.

Yes: This causes the assignment of code to a UserData field to be executed, which is the wrong semantics. That is the reason why I had to create new code.

However, this is still work in progress.

kortenkamp:
I just hope I did not mess up too much code… I will check then main branch again after this.

PR #861, opened 2021-12-06
Userdata lists reference copy

strobelm:


kortenkamp:
Thank you @strobelm for providing this fix!

PR #866, opened 2022-02-12
Remove some circular dependencies of CSNumber

strobelm:
This MR removes some circular dependencies, especially CSNumber does not depend on List and General anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
         "lint-staged": "^11.1.1",
         "marked": ">=0.3.5 <=0.3.12",
         "mocha": "^9.2.0",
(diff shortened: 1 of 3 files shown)


kortenkamp:
Thank you for this - one question though: isn’t it a bit strange to have solveCubic in List instead of CSNumber? Semantically it does not belong there, does it?

strobelm:
That is a fair point, it feels a bit heavy for CSNumber, too. I might move it to a separate file then.

The background of the whole story is to remove all circular dependencies that sneaked in to the code s.t. we can remove the old build system in the long run (circular  dependencies makes it very hard to use an off-the-shelf solution)

PR #869, opened 2022-02-18
do not print stack trace  when if condition is undefined

### LINKED ISSUE #868: `Condition for if is not boolean` should not be thrown
Evaluation of an `if` function should not complain about undefined conditions. In original Cinderella, you can use undefined values in the condition of `if` which causes the result of `if` to be undefined as well (and neither the `true` nor the `false` function is evaluated). This seems to be true in CindyJS as well, but a stack trace is thrown. 

In particular this is difficult as the `isundefined` function seems to be broken as well (will be a separate issue).

Example code for testing could be:
```
if(blabla,print("a"),print("b"));
```

kortenkamp:
This fixes #868 - there should be no stack trace when the if condition is undefined.

### DIFF
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -823,7 +823,7 @@ evaluator.if$3 = function (args, modifs) {
         } else if (args.length === 3) {
             return evaluate(args[2]);
         }
-    } else {
+    } else if (v0.ctype !== "undefined") {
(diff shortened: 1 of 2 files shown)


strobelm:
Thanks a lot for this. I was wondering if CindyScript also knows the concept of weak comparison or would it make sense to introduce? I can see that `"", [], {}` etc could be considered falsy.

PR #871, opened 2022-04-16
(chore) run prettier

strobelm:
Somehow the pre-commit hook of prettier got lost. So we need to do some formatting. Running prettier by hand and fixed the pre-commit hook.


kortenkamp:
Thank you, @strobelm !

PR #884, opened 2022-08-01
introducing "else" in Evaluator.js to make the code more readable.

kortenkamp:


elkins0:
If you really want to make the code more readable perhaps the use of switch statements would be more effective here.
Note that switch cases use strict comparison (===).
For example there are three opportunities to do so here:
```javascript
function evaluate(a) {
    if (a === undefined) {
        return nada;
    }
    switch (a.ctype) {
    case "infix":
        return a.impl(a.args, {}, a);
    case "variable":
        return evaluate(namespace.getvar(a.name));
    case "function":
        callStack.push(a);
        a = eval_helper.evaluate(a.oper, a.args, a.modifs);
        callStack.pop();
        return a;
    case "void":
        return nada;
    case "field": {
        const obj = evaluate(a.obj);
        switch (obj.ctype) {
        case "geo":
            return evaluate(Accessor.getField(obj.value, a.key));
        case "list":
            return List.getField(obj, a.key);
        case "JSON":
            return evaluate(Json.getField(obj, a.key));
        default:
            return nada;
        }
    }
    case "userdata": {
        const obj = evaluate(a.obj);
        let key = General.string(niceprint(evaluate(a.key)));
        if (key.value === "_?_") key = nada;
        switch (obj.ctype) {
        case "geo":
            return evaluate(Accessor.getuserData(obj.value, key));
        case "list":
        case "string":
            return evaluate(Accessor.getuserData(obj, key));
        case "JSON":
            return evaluate(Json.getField(obj, key.value));
        default:
            return nada;
        }
    }
    default:
        return a;
    }
}
```

kortenkamp:
I tried the `switch` before and did not find it more readable… and I still think so.

elkins0:
It is a personal preference I suppose. If I write the code I will always use switch statements in situations like this. See [switch vs if else](https://www.geeksforgeeks.org/switch-vs-else/) for some thoughts on this. It starts with:
"A switch statement is usually more efficient than a set of nested ifs. Deciding whether to use if-then-else statements or a switch statement is based on readability and the expression that the statement is testing."

strobelm:
Well it is definitely more readable than before. I think the best way would be to extract the details of the cases into separate functions in the long run.

PR #886, opened 2022-08-05
 Transition to ECMAscript 2016

### LINKED ISSUE #885: Transition to ECMAscript 2016
With the latest modernisations by @strobelm we should think about moving the codebase entirely to ECMAscript 2016 (ES7), as the `**` operator is ES7. 

This means - check which browsers we want/need to support and use the appropriate ECMAscript version as a target for the build process. If there is no longer a need for transpilation, I suggest to get rid of it altogether. 

Any comments?

strobelm:
Closes #885

### DIFF
--- a/make/build.js
+++ b/make/build.js
@@ -307,7 +307,7 @@ module.exports = function build(settings, task) {
         this.setting("closure_version");
         var opts = {
             language_in: "ECMASCRIPT6_STRICT",
-            language_out: "ECMASCRIPT5_STRICT",
+            language_out: "ECMASCRIPT6_STRICT",
             dependency_mode: "PRUNE_LEGACY",
             create_source_map: "build/js/Cindy3D.js.map",
             compilation_level: this.setting("c3d_closure_level"),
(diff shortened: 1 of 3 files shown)


make/build.js
```diff
@@ -700,7 +700,7 @@ module.exports = function build(settings, task) {
         this.setting("closure_version");
         var opts = {
             language_in: "ECMASCRIPT_2018",
-            language_out: "ECMASCRIPT_2018",
+            language_out: "ECMASCRIPT_2016",
```
kortenkamp, line 703: Does it make sense to use `ECMASCRIPT_2016` here (it was `ECMASCRIPT_2018`) – the CindyXR plugin needs WebXR, and I do not think that any browser capable of doing WebXR is not capable of doing `ECMASCRIPT_2018`…

kortenkamp (changes requested):
Can you check my remark on the WebXR plugin and maybe change the `language_out` accordingly?

kortenkamp (approved):

PR #887, opened 2022-08-22
CindyScript import

BernhardWerner:
Loads CindyScript files (marked by `.cjs` file ending) and adds their code **at the beginning** of the init-script. The file names (and relative paths) have to be listed in the dictionary that's passed to CindyJS/createCindy as an array for the key `import`. I.e., like this:
```
          var cdy = CindyJS({
                ports: [{ id: "CSCanvas", width: 500, height: 500 }],
                scripts: "cs*",
                import: ["cindyscript_libraries/libraryA.cjs", "cindyscript_libraries/libraryB"]

            });
```
Adding the file ending to the path is optional.

Note that the libraries get prepended to the init-script **back to front**. So, in the example above, the code from libraryA comes before the code from libraryB which comes before the custom user code. That way, libraries can reference each other's code.

Since this uses the `fetch` command in JavaScript, it only works on a web server. So, for local testing, start one with `python -m http.server` or however else you are comfortable with.

If the import key doesn't exist or its value is an empty array, opening the file locally works as always.

---

See `examples/cindyscript_import.html` for a very minimalistic example. It uses two libraries created in `examples/cindyscript_libraries`

### DIFF
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -371,6 +371,7 @@ CAUTION!
 Since this uses the 'fetch' command, it only works on a web server. So, for local testing, start one with 'python -m http.server' or however else you are comfortable with.
 If the key 'import' doesn't exists or its value is an empty array, opening the file locally works as always.
 */
+
     if (data.import && Array.isArray(data.import) && data.import.length > 0) {
         let initId = "csinit";
         if (data.initscript) {
@@ -391,15 +392,27 @@ If the key 'import' doesn't exists or its value is an empty array, opening the f
             console.log("Loading " + library + " ...");
 
             let query = library.search(/.+\.cjs$/) == -1 ? library + ".cjs" : library;
-            let response = await fetch(query);
 
-            if (response.status === 200) {
-                let code = await response.text();
-                let safety = code[code.length - 1] == ";" ? "" : ";";
-                fullCode = code + safety + "\n" + fullCode;
-                console.log(library + " loaded!");
-            } else {
-                console.log("CAUTION! Import of " + library + " failed.");
+            try {
+                let response = await fetch(query);
+
+                if (response.status === 200) {
+                    let code = await response.text();
+                    let safety = code[code.length - 1] == ";" ? "" : ";";
+                    fullCode = code + safety + "\n" + fullCode;
+                    console.log(library + " loaded!");
+                } else {
+                    console.log("CAUTION! Import of " + library + " failed.");
+                }
+            } catch (e) {
+                if (e.message === "Failed to fetch") {
+                    console.log("CAUTION! Import of " + library + " failed.");
+                    console.log(
+                        "This website seems to not run on a web server. CindyJS will continue without importing CindyScript libraries."
+                    );
+                } else {
+                    throw e;
+                }
             }
         }
 
--- a/src/js/Setup.js
+++ b/src/js/Setup.js
@@ -244,7 +244,7 @@ function isCinderellaBeforeVersion() {
     return false;
 }
 
-function createCindyNow() {
+async function createCindyNow() {
     startupCalled = true;
     if (waitForPlugins !== 0) return;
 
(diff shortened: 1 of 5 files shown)


kortenkamp:
Apparently the Travis checks fail… can you check out why?

BernhardWerner:
Is there a way to get more details out of Travis than just "it failed"?

When building CindyJS locally, I get 0 errors and 8 warnings (and the new code works perfectly fine). But the warnings are all in places far away from my code changes, as far as I can see. In particular, they were also present when I ran the built process right after forking the main branch, IIRC.
Apart from that, I have no idea what the issue might be.

kortenkamp:
The problem seems to be the new file type .cjs — see https://app.travis-ci.com/github/CindyJS/CindyJS/builds/254654421#L1067 (don’t know whether this is visible for you). It is the textattr task that checks the git attributes.

BernhardWerner:
I see it, thank you! (But I had no idea how to get to this report.)
Since I have very little experience with this: What exactly is the problem? That the file extension .cjs is unknown? That it is perceived as malicious? Something else?

Since we're now at it: I decided to have a custom file name for CindyScript code. This is, however, not necessary. We can easily say that the CindyScript code should be stored in a .txt-file. I just thought it might be easier to keep track of the files in a bigger project when they use a custom file type.
Does anyone have an opinion on whether we such a custom file ending or not?

kortenkamp:
> I see it, thank you! (But I had no idea how to get to this report.) 
In order to find the detailed report you have to go to "Details" and then to "The build"… 
> Since I have very little experience with this: What exactly is the problem? That the file extension .cjs is unknown? That it is perceived as malicious? Something else?
It is just unknown. And I am not the one who introduced all these checks and the automatic builds… I would not have checked for that. However, you can introduce this attribute in `.gitattributes` – just add a line 
```
*.cjs.     text
```
to tell git that this is a text file and not a binary.

> Since we're now at it: I decided to have a custom file name for CindyScript code. This is, however, not necessary. We can easily say that the CindyScript code should be stored in a .txt-file. I just thought it might be easier to keep track of the files in a bigger project when they use a custom file type. Does anyone have an opinion on whether we such a custom file ending or not?

I think it is good to have a `.cjs` ending.

BernhardWerner:
Thank you for your help, @kortenkamp ! 
I added the line to gitattributes. Let's see what happens...

src/js/Setup.js
```diff
@@ -356,6 +356,61 @@ function createCindyNow() {
...
+    /*
+Loads CindyScript files (marked by .cjs file ending) and adds their code AT THE BEGINNING of the init script.
+The file names (and paths) have to be listed in the dictionary that's passed to CindyJS/createCindy as an array for the key 'import'. I.e., like this:
+          var cdy = CindyJS({
+                ports: [{ id: "CSCanvas", width: 500, height: 500 }],
+                scripts: "cs*",
+                import: ["cindyscript_libraries/libraryA", "cindyscript_libraries/libraryB"]
+
+            });
+Note that the libraries get prepended to the init-script BACK TO FRONT. So, in the example above, the code from libraryA comes before the code from libraryB whoch comes before the custom user code. That way, libraries can reference each others code.
```
kortenkamp, line 368: This cautionary message is a bit confusing – they are prepended back-to-front, yes, but actually they just come in the order in which they are (as expected). So you could also write "The libraries are prepended as a whole in the order listed, so libraryA is loaded before libraryB."

src/js/Setup.js
```diff
@@ -356,6 +356,61 @@ function createCindyNow() {
...
+
+CAUTION!
+Since this uses the 'fetch' command, it only works on a web server. So, for local testing, start one with 'python -m http.server' or however else you are comfortable with.
```
kortenkamp, line 371: It would be nice to fall back to regular file loading if this is not on a webserver.

src/js/Setup.js
BernhardWerner: Oh, that's much better, yes!

src/js/Setup.js
kortenkamp: And actually I wonder whether this could be done using a `file://` URL, but I do not have detailed knowledge about this in JavaScript, so this is a wild guess.

kortenkamp:
> Thank you for your help, @kortenkamp !
> I added the line to gitattributes. Let's see what happens...

Works fine now.

src/js/Setup.js
BernhardWerner: Yes, it would!
I don't know how to check for that, yet. But I will look into it now...

kortenkamp (approved):
Looks fine to me. I suggest to add loading plain files (without the need for a web server), but that is not a showstopper.

src/js/Setup.js
kortenkamp: Ok, I will wait a little with merging then.

BernhardWerner:
> Looks fine to me. I suggest to add loading plain files (without the need for a web server), but that is not a showstopper.

I'm not sure what you mean. Do you mean the exact same functionality without the need to be on a web server?
If yes, then it's not possible as far as I know. The only way to load a local file that I'm aware of is to use either XMLHttpRequest (old) or fetch (new). Both only work if you are on a web server for security reasons. But, of course, it could very well be that I haven't found any alternative.

kortenkamp:
Well, there seems to be FileReader: https://w3c.github.io/FileAPI/#dfn-filereader – not sure whether that is an option. But really, this is not very important, it just adds a bit of convenience for local files.

src/js/Setup.js
BernhardWerner: Alright, as far as I can see, CindyJS proceeds as normal if the fetch command fails due to not being on a web server. (Of course, missing code from the not loaded code files might be an issue.) But I added a console output, that makes it clearer that this is the case.

BernhardWerner:
I looked into FileReader a bit, and it might only work with files provided via an input field on the website. I'll do a bit more research and testing. But the current webserver-dependent version works.

PR #889, opened 2022-09-05
JSON object extensions

kortenkamp:
This (important) pull request adds the `self()` function that can be used in code that is stored in user fields or JSON objects to identify the containing object, and the `eval()` function that is needed to call objects while setting local variables. 

I also added unit tests, and two new examples that use that code. If @BernhardWerner or @strobelm can have a quick look I am very happy, as we have to create a release version as soon as possible for the VAM project of DZLM. Thanks!

### DIFF
--- a/examples/164_JSON-self.html
+++ b/examples/164_JSON-self.html
@@ -1,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>JSON Self test</title>
+    <style type="text/css">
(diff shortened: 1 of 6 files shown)


kortenkamp:
After some positive feedback from @strobelm I will just merge this, if necessary, we will do bug fixes later. As all tests pass, this should not be a large risk.

BernhardWerner:
Sounds great! Will try it out this afternoon!

(And sorry for the late response. I'm on holiday right now...)

PR #891, opened 2022-09-06
Comparing two JSON objects works.

### LINKED ISSUE #890: Comparing two JSON objects returns false, always
Two JSON objects should be the same if they reference the same object. Currently, comparison of JSON objects always returns `false`.

kortenkamp:
Fixes #890 by introducing the necessary comparisons in code. Also adds the necessary tests.


kortenkamp:
Fixes #890

PR #895, opened 2022-11-02
adding free points from CindyScript does not invoke the prover

### LINKED ISSUE #894: createpoint is slow
An old oral comment of @montaga about https://cinderella.de/files/ml215/stellenwerttafel.html was that moving a point from the "Einer" to the "Tausendstel" is very slow. The root cause of this is that 999 elements are being created. However, adding elements seems to be $O(n)$ instead of $O(1)$ for points, which leads to quadratic runtime in the number of elements being added. 

It is important to check whether the prover is needed at all when adding an element. It is not if adding a free point from a script. In Cindy Classic it is necessary to use the prover on all elements that have been added by the user, so we have to take care to use the right semantics when we will create "Cindy in Cindy", but that is not the case yet.

kortenkamp:
When adding elements using the `createpoint(…)` function, the prover does not need to be invoked. However, it was, leading to a quadratic runtime when adding _n_ points. I tried to fix this by invoking `addElementNoProof` instead of `addElement`, but this still had quadratic runtime as the conjectures comparing the new element to all other elements are created in the `addElementNoProof` code.

This pull request changes the behaviour of `addElementNoProof` to match its name – conjectures are only created when the prover is invoked, and also uses `addElementNoProof` for the `createpoint(…)` function.  

An example illustrating the now instant creation of 999 elements has been added as well, the Stellenwerttafel. Add a point in the first column and then drag it to the rightmost column to create 999 elements. This took about 5-15 seconds, depending on the computer/tablet you used, and is now immediate.

This closes #894.

### DIFF
--- a/examples/165_stellenwerttafel.html
+++ b/examples/165_stellenwerttafel.html
@@ -1,0 +1,390 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Stellenwerttafel-Tausendstel.cdy</title>
+  <style type="text/css">
+    html,body { margin: 0px; padding: 0px; }
+    html,body,#CSCanvas { width: 100%; height: 100%; }
+  </style>
+     <style type="text/css">
+         * {
+             margin: 0px;
+             padding: 0px;
+         }
+
+         #CSConsole {
+             background-color: #FAFAFA;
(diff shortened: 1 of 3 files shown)


kortenkamp:
Merging this as an emergency fix – Stellenwerttafel is being used today at a lecture in Dresden.

PR #896, opened 2022-11-03
TypeScript

strobelm:
Hi,

this PR adds support for TypeScript to the project. This is something that I wanted to introduce for quite a while. In some plugins we are using Closure Compiler annotations, but those are quite cumbersome for our code style. 

The adoption will be gradually, as a tracer I added typing to CSNumber. There is only little overhead when it comes to the source code. Also the enforcement is very light, you can use as many plain JavaScript as you want (and most of the source files are still JavaScript, maybe add an any type if needed). 

The type system already uncovered that this line was essentially unused: https://github.com/CindyJS/CindyJS/blob/main/src/js/libcs/CSNumber.js#L58

Please let me know what you think.


kortenkamp:
Just to be on the same page: what exactly are the consequences of that change?

strobelm:
From a run-time perspective: nothing. The type information is removed with a compilation step. But you get some nice feedback from the TypeScript compiler in case you made a mistake.

If you by accident want to add a raw number, say 3 to CindyScript Number e.g.
```
function myTestFunction(a: CindyNumber, b: CindyNumber): CindyNumber {
        return {
                ctype: "number",
                value: {
                        real: a.value.real + b.value.real,
                        imag: a.value.imag + b.value.imag,
                },
        };
};

myTestFunction(CSNumber.real(3), 3);
```
then the compiler will tell you that this not possible:
```
src/js/libcs/CSNumber.ts:806:31 - error TS2345: Argument of type 'number' is not assignable to parameter of type 'CindyNumber'.

806 myTestFunction(CSNumber.real(3), 3);
```

Or if you have a typo in the ctype:
```
function myTestFunction(a: CindyNumber, b: CindyNumber): CindyNumber {
  return {
    ctype: "numberr",
    value: {
      real: a.value.real + b.value.real,
      imag: a.value.imag + b.value.imag,
    },
  };
};

myTestFunction(CSNumber.real(3), CSNumber.zero);
```

then the compiler will tell you
```
src/js/libcs/CSNumber.ts:798:5 - error TS2322: Type '"numberr"' is not assignable to type '"number"'.

798     ctype: "numberr",
        ~~~~~

  src/js/types.ts:8:3
    8   ctype: "number";
        ~~~~~
    The expected type comes from property 'ctype' which is declared here on type 'CindyNumber'
```

src/js/types.ts
```diff
@@ -1,99 +1,100 @@
...
-  };
+    usage?: "Angle";
+    ctype: "number";
+    value: {
+        real: number;
+        imag: number;
+    };
 }
 
-export interface Nada {
```
kortenkamp: This is becoming very verbose... I would prefer to keep such names short. Can we keep using Nada?

strobelm: Sure. Are there any preferences for the interface names?

e.g.
CindyNumber, CNumber, CindyNum?
CindyList, CList, CindyLst?

kortenkamp:
It would be nice to get some input from @richter-gebert – on the one hand I am happy with using TypeScript for more type safety, on the other hand it seems to become more and more work just to maintain that. Just as one example: when I saw your changes from List to CindyList I immediately felt like this is raising questions that might be interesting in a way, but do not help when we have to implement some (still missing) features… So, I suggest we try to schedule a meeting to discuss this.

src/js/types.ts
kortenkamp: Prepend with `C` or `CS`. Probably `CS` is better.  

I had a look at the Cinderella code, there we use (thanks to packages) normal names for math:

```
math (beta) % ls
Assignable.java			Int.java			PolarConversion.java
AssignableAdapterB.java		Mat.java			RMat.java
AssignableAdapterI.java		Mat2.java			Tracable.java
AssignableVector.java		MathHelper.java			TransformationHelper.java
Bool.java			Matrix3D.java			TypedObject.java
Complex.java			MovableReference.java		Vec.java
Doub.java			MyMath.java			Vec2.java
FormulaValue.java		Nada.java
GetSetAdapter.java		Point2D.java
```

For geometric elements we usually prepend with `PG`. And CindyScript operators are called `OpSomething`. And Lists are called CEnum.

richter-gebert:
OK, von mir aus. Macht halt nix kaputt!

PR #897, opened 2023-02-14
drawtext will return the coordinates of the bounding box of the text

kortenkamp:
Currently it is very difficult to find out the screen bounds of a text. This small change gives access to the coordinates of the bounding rectangle for a text that has been output through `drawtext`. I plan to add this to Cinderella classic, too, but right now there is a semantic problem when using several ports. I will discuss this with @richter-gebert soon, but until then it would be nice to use this in production code already.

### DIFF
--- a/examples/166_boxsize.html
+++ b/examples/166_boxsize.html
@@ -1,0 +1,60 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Construction.cdy</title>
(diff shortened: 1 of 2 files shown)


strobelm:
I will have a look at it asap. Will also add some unit tests.

strobelm:
The code looks good to me, just did a small refactoring to make to code a bit more concise.  Unfortunately we do not have an easy way right now to unit test drawing functions.

strobelm (approved):

PR #902, opened 2023-08-27
(chore): run prettier

strobelm:
No functional changes, only fixing the code style.


kortenkamp:
Great, thank you, please let me know how I could do this myself when I check in something else someday!

strobelm:
Oh that is actually pretty easy, just run

` npm run-script prettier`

kortenkamp:
thanks a lot!

So it would make more sense to let this run automatically before a commit, instead of checking the prettyness and rejecting commits that are not formatted properly, wouldn't it? Or we should put it somewhere in the coding guidelines that you should always run `npm run-script prettier` before committing…

PR #904, opened 2023-08-30
Make sure that /Cindy.js is matched in base URL

kortenkamp:
If your HTML is using `src="Cindy.js"`, please change this to `src="./Cindy.js"˙.


kortenkamp:
This is a quick fix for the presentation in Kyoto, so I skipped review. I hope this does not break any existing examples (see the comment above to work around)

PR #907, opened 2023-09-08
Box direct assign

strobelm:
There are some concerns regarding performance, changing back to the original version.


kortenkamp (approved):
Thank you

PR #908, opened 2023-09-24
List test suite

strobelm:
Hi, I have created a comprehensive test suite for out List operations. It should cover all of the crucial operations.

Some things I did not directly test related:

- Discovered a bug in the vandermonde matrix code, but it was not used anyway.
- Removed the unused Hessenberg transformation
- Removed List._helper.isAlmostId, since it also was unused 
- Removed some unused constants


strobelm:
While we are at it, any input on this PR? @kortenkamp @richter-gebert @BernhardWerner?

kortenkamp:
As these are just tests, it should be no problem to merge. However, I think that removing `List._helper.isAlmostId` is not a good idea, instead we should investigate why it is not used (should be used for `~==` comparison of lists, should it not?)

strobelm:
Oh and while I look at the code, there might be room for improvement.

Looking at  https://github.com/CindyJS/CindyJS/blob/3716f4086a6d05ea5e095035c8c0660cd921f38e/src/js/libcs/Operators.js#L912C1-L915C6   
```
    for (let i = 0; i < a1.value.length; i++) {
        const av1 = a1.value[i];
        const av2 = a2.value[i];

        if (av1.ctype === "list" && av2.ctype === "list") {
            erg = erg && List.almostequals(av1, av2).value;
        } else {
            erg = erg && comp_almostequals([av1, av2], []).value;
        }
    }
```
could return early when `erg` equals to `false`

 ```
    for (let i = 0; i < a1.value.length; i++) {
        const av1 = a1.value[i];
        const av2 = a2.value[i];
        
        if(!erg) break;

        if (av1.ctype === "list" && av2.ctype === "list") {
            erg = erg && List.almostequals(av1, av2).value;
        } else {
            erg = erg && comp_almostequals([av1, av2], []).value;
        }
    }
```

kortenkamp:
> could return early when `erg` equals to `false`
> 
> ```
>    for (let i = 0; i < a1.value.length; i++) {
>        const av1 = a1.value[i];
>        const av2 = a2.value[i];
>        
>        if(!erg) break;
> 
>        if (av1.ctype === "list" && av2.ctype === "list") {
>            erg = erg && List.almostequals(av1, av2).value;
>        } else {
>            erg = erg && comp_almostequals([av1, av2], []).value;
>        }
>    }
> ```

But if you are at it, then why not write it as a while loop? Or is this too old-school?

strobelm:
Both would work, but this is not really the scope of this PR. I will create a separate one.

strobelm:
Rebased that on current main. If there are no objections by tomorrow I am going to merge it.

PR #909, opened 2023-10-24
Added the `import` function to CindyScript.

BernhardWerner:
It loads the cjs file it is given via its path as an argument, assumes the file contains CindyScript code and executes it on the spot. In particular, it behaves like the `import` function in Cinderella. Note that it uses fetch and therefore only works when the CindyJS applet runs on a web server. (Just like the code import on CindyJS start.)
In the example file `import_cindyscript.html` this is shown by loading the file `cindyscript_libraries/libraryC.cjs` on every mousedown which then changes the colour of the applet. Not the intended main use, but shows the functionality.

### DIFF
--- a/examples/cindyscript_libraries/libraryC.cjs
+++ b/examples/cindyscript_libraries/libraryC.cjs
@@ -1,0 +1,8 @@
+
+
+println("Library C");
+col = 1 - col;
+println("This is executed during runtime.");
+
+
+
--- a/examples/import_cindyscript.html
+++ b/examples/import_cindyscript.html
@@ -9,16 +9,24 @@
             println("Main code");
             test = test^2;
             println("test is " + test);
+            
+            col = 1;
         </script>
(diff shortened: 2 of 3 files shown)


BernhardWerner:
Any way to find out what the problem here is?

BernhardWerner:
@strobelm, do you happen to know how to make Travis cooperate and at least tell me, what exactly failed?

strobelm:
@BernhardWerner the main branch should work now again. Could you rebase your PR?

BernhardWerner:
> @BernhardWerner the main branch should work now again. Could you rebase your PR?

Done. Let's see what Travis says this time around...

strobelm:
The code looks mostly fine to me, I have streamlined it a bit with https://github.com/BernhardWerner/CindyJS/pull/1/commits/61db7f9ce42628cb8284ad3f0a6259ce5ce222c1

What we need to consider is that we do not really check any integrity of the imported file that might open us up to code injection.

PR #910, opened 2024-02-18
Update to node 20

strobelm:
This PR will update node to the latest LTS.


strobelm:
@kortenkamp or @richter-gebert : any objections? I guess this is a no brainer.

PR #913, opened 2024-03-01
Bug #912: handle case where bounding box is undefine

### LINKED ISSUE #912: KaTeX rendering bug in version 0.8.23
When using the currently latest version 0.8.23 of CindyJS, I get the following error when I want to render a maths formula:

chrome_2XdlmfrjkP

In the simple case of, say, 
```
drawtext([0,0], "$123$");
```
the program seems to work fine otherwise, and the text is still rendered. In a more complicated set-up, the whole site crashes. I haven't identified the right combinations of commands yet. But it seems to happened when I draw a maths formula to a texture via the `canvas` command and then try to manipulate the texture via a CindyGL shader.

strobelm:
When we introduced the bounding box, we did not take care of the case when `box` is undefined. E.g. here: 

https://github.com/CindyJS/CindyJS/blob/85add6015766a7037977ba072c5f85684893aca1/src/js/libcs/OpDrawing.js#L982-L989

@BernhardWerner could you test if this fixes your crash? If so, this closes  #912.

### DIFF
--- a/src/js/libcs/OpDrawing.js
+++ b/src/js/libcs/OpDrawing.js
@@ -1027,6 +1027,8 @@ eval_helper.drawtext = function (args, modifs, callback) {
 
 evaluator.drawtext$2 = function (args, modifs) {
     const box = eval_helper.drawtext(args, modifs, null);
+    if (!box) return nada;
+
     const pt1 = General.withUsage(List.realVector(csport.to(box.left, box.bottom)), "Point");
     const pt2 = General.withUsage(List.realVector(csport.to(box.right, box.bottom)), "Point");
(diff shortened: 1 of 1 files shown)


kortenkamp (approved):

kortenkamp:
This should be introduced in `pixelsize()`, too, I guess. But for the moment, I see no reason not to merge.

strobelm:
I briefly checked `pixelsize` and could not really spot where we return something null or undefined, but rather default values. Could you point me to where this could go wrong? Anyway I am going to release a hotfix version now.

kortenkamp:
I did not have a deeper view, but as I wrote `pixelsize` by duplicating code from `drawtext` I had the strong feeling it could suffer from the same problem. But maybe it doesn't. Thanks a lot for releasing #913!

PR #917, opened 2024-05-03
WIP: Restore travis

strobelm:


kortenkamp:
What is your plan here, @strobelm ? I do not think we intend to revert to Travis, do we?

PR #918, opened 2024-05-09
Cindyscript import: packages

BernhardWerner:
An extension to the import system of CindyJS. You can now
- import to all events,
- specify the place where the loaded script should be (before or after the given code in the html script tags) and
- import full packages.

An example for all of them:
```
            var cdy = CindyJS({
                ports: [{ id: "CSCanvas", width: 500, height: 500 }],
                scripts: "cs*",
                import: {
                    "init": ["cindyscript_libraries/libraryA.cjs", "cindyscript_libraries/libraryB"],
                    "mousedown": [{"name": "cindyscript_libraries/libraryC", "place": 3}],
                    "packages": ["cindyscript_libraries/test_package"]
                }
            });
```

Script placement: 
If a script is given by just its file path, it will be added before the given code. If it specified by a dictionary with keys `name` and `place`, the place works similar to z-layers. The given code in the html script tag is considered to be on level 0. Negative script places will add the script before it, positive places will set it afterwards.

Packages:
Packages are directories which contain a file `config.json` and cjs-files. This config file contains the name and placement information for the CindyScript files, basically identical to the dictionary used in the import itself. I.e., the config file might look like this:
```
{
    "init": [
        {"name": "init.cjs", "place": -4}
    ],
    "draw": [
        "draw.js"
    ]
}
```
--- 

Note that currently, packages cannot load other packages directly. Also, if a user loads multiple packages and stand-alone scripts, it is currently their own responsibility to make sure that the placements of the individual script files make sense and that the scripts don't interfere with another.

---

In principle, this import feature works as is and can be merged to the main branch. (Hence the pull request.) Since loading multiple packages is desirable, a namespace system will be added in the future.


kortenkamp:
Thank you for implementing this!

PR #923, opened 2024-05-18
Optional indices in array functions + compatability fix for max/min

BernhardWerner:
Functions that iterate over an array and apply some sort of code to the entries now have a version with four arguments where the third one specifies a reference to the index. E.g. one can now do something like this:
```
points = [(1,2), (3,4), (5,6), (7,8), (9,0)];
forall(points, p, index,
     fillcircle(point, index);
);
```
and get diferently sized circles.

This was already implemented for `apply` in the past, but is now also available for `forall`, `sort`, `select`, `max` and `min`. Moreover, it also works with JSON objects  (when it makes sense) and a reference to the key instead of the index. Again, this was available for `apply`  before and now also works  for `forall` and `select`.

---

Additionally, this pull request is a compatability fix: In the past, `max` and `min` only worked for numbers and lists of number. Now they can take every datatype that `sort` also takes.

### DIFF
--- a/examples/cindyscript_libraries/another_package/config.json
+++ b/examples/cindyscript_libraries/another_package/config.json
@@ -1,0 +1,5 @@
+{
+    "init": [
+        {"name": "init.cjs", "place": 10}
+    ]
+}
--- a/examples/cindyscript_libraries/another_package/init.cjs
+++ b/examples/cindyscript_libraries/another_package/init.cjs
@@ -1,0 +1,1 @@
+println("This is a second package");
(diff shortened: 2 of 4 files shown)


kortenkamp:
Please allow for a few days, I would like to implement this in CindyClassic at the same time to preserve compatibility.

kortenkamp:
Hi, while implementing this in Classic I was thinking about adding unit tests for these as well. Just to make sure: You are aware that the order of `apply` for JSON objects is not determinend, are you?

BernhardWerner:
I am aware. But since I wasn't the one who implemented apply for JSON objects, I didn't think about it much. I will as soon as I find time, but at the moment, I don't think that there's a problem with that.

kortenkamp:
Another thing that we have to discuss at some point: `select` could use 0/1 oder `isEmpty` checks in addition to boolean expressions. In Cinderella Classic this is currently not the case, the expressions must return a boolean result – but maybe it is convenient to break backwards compatibility ;-) ?

PR #926, opened 2024-06-15
Bug fix for text outline

BernhardWerner:
Render2D did not properly reset and handle outline width and outline color when `drawtext` was called without outline modifiers or with alpha. This is now fixed.


kortenkamp:
Thanks for the fix, and what a good thing to have tests for the Renderer

PR #940, opened 2025-01-30
Pointer events

kortenkamp:
This is a major change that will use pointerevents instead of mouse events. It might break multi-touch, however, I do not think this is currently being used by somebody and can be fixed soon. 

The most important feature of this release is that it supports pointer events that occur outside the CindyJS canvas.

Please, give it a spin, in particular also on platforms like iPad or mobile phones.


kortenkamp:
I tested on desktop, iPhone and iPad and also fixed multitouch, so I will just merge this and release. If there is any problem let me know and I will fix on short notice.

Issue #946, opened 2025-02-11
numerical differentiation is not supported in CindyJS


kortenkamp:
Cindy Classic has a `d(f(x),x)` function for numerical differentiation which is not implemented in CindyJS.

PR #947, opened 2025-02-11
946 numerical differentiation is not supported in cindyjs

kortenkamp:
This implements the missing `d(f(#),x)` function.


kortenkamp:
fixes #946

PR #961, opened 2025-06-12
3D Support for CindyGL

benjaminSoelch:
Add support for 3D-geometry to CindyGL

Major changes:

* Update CindyGL to WebGL2
* Add support for depth-information
* Add internal coordinate-system
* Allow restricting rendering to 3D-bounding boxes (sphere+cylinder+cube)
* Add support for triangular meshes
* Quality of life updates for colorplot expressions:
  - Unreachable if-branches in compiled code no longer need to type-check correctly (if condition can be evaluated at compile-time)
  - allow evaluation of length(..) at compile-time
  - add special built-in functions/variables in compiled code, for access to coordinate system access and bounding box
* add support for adding modifiers to plotted objects
* Add plugin-functions for treating expression as CindyScript objects (`cglLazy` -> convert code to object / `cglEval` -> evaluate code previously stored in object )
* Add CindyScript implementation for 3D-rendering on top of 3D colorplot functions (sphere,cylinder,curve,torus,arc,polygon,mesh,surface) -> cgl3d.cjs

high-level documentation of current features:
* `CindyGL3D.md` core renderer
* `Primitives.md` implementation of geometric primitives

Known issues:
* Some CindyGL files do not pass the pre-commit check due to false-positives in the linter, this problem was already present in the original CindyGL version in the main git repository


benjaminSoelch:
@BernhardWerner 


1 .
The plugin distinguishes adding objects to the internal geometry (using `colorplot3d` / the drawing functions in `cgl3d.cjs`) and rendering (using `cglRender3d`).
Every time the render function is called all objects drawn since last call to `cglReset3d` are drawn.

For large scenes the intended way is to call the drawing functions once in the initialization phase or when the geometry changes, while only calling `cglRender3d` in the main render loop.

The reason why drawing and rendering are separated is that executing the (about 50 lines of) CindyScript code (before reaching the main rendering core) needed for drawing a new sphere needs a significant amount of time.

The following code pasted into the `cindygl3d/playground` example will draw a NxNxN cube of spheres
```
N=32;
forall(0..N,x,forall(0..N,y,forall(0..N,z,
  sphere3d(2*(x/N-0.5,y/N-0.5,z/N-0.5),1/N);
)));
```
On my machine *rendering* (even then rotating/zooming) the spheres runs relatively smoothly up to N=50 while the initial drawing leads to a noticeable delay already for N=20.

2. It is possible to update the distance to the view-plane using the `z0` modifier of `cglCoordSystem`, an option for offsetting the camera from the center of the view-plane is on my TODO list, but currently not implemented.

3. The main rendering engine (bounding box computations, list of objects composing the scene, transparency handling, coordinate system) is an extension of CindyGL and lives in the Javascript part of the plugin.

The `cgl3d.cjs` file defines a simple interface for drawing geometric primitive objects, implemented on top of a modified colorplot function and a triangle-renderer (with the help of a few helper functions for allowing to use expressions as data).

BernhardWerner:
@benjaminSoelch 

1. This is exactly the example that I tried! My computer struggled a lot at N=30. Probably not that big of an issue for most use cases.
2. Right. Then the method to do camera movements is to move the objects in the opposite way.
3. Thanks! Didn't look to closely at the individual commits yet.

kortenkamp:
I checked that this does not kill any important examples and decided that it is good to be merged. Thank you for the great work, @benjaminSoelch !

Issue #965, opened 2025-07-22
Modifiers should be evaluated in the call-site context


benjaminSoelch:
When a global variable with the same name as a function parameter is used within a modifier, the value of the parameter will be used when evaluating the modifier.
This behavior can lead to confusing bugs at the call side, as the function will change its behavior depending on the variable names used in the modifiers, without any locally observable change to global the variables.

The expected behavior would be that the variables for the function parameters are only accessible within the function body, while modifiers will only use variables that are visible at the call site.

Example code:
```
// library function
bug(x):=(
  print("local: "+text(x));
  print("modif: "+text(modif));
);
// user code
x = 0;
print("global: "+text(x));
bug(1,modif->x); // modif evaluates to 1
print("global: "+text(x)); // x is still 0

y = 0;
print("global: "+text(y));
bug(1,modif->y); // modif evaluates to 0
print("global: "+text(y));
```

benjaminSoelch:
Looking through the source code, it seems like modifiers being evaluated after function parameters is the currently intended behavior.
Does any code actually depend on this evaluation order?


Additionally I found that the order in which modifiers are declared also has an effect on their values:

```
bug2():=(
    print(a);
    print(b);
);
a = 0;
b = 1;
print((a,b));
bug2(b->a,a->b); // prints 0 twice
print((a,b)); // a and b are unchanged
bug2(a->b,b->a); // prints 1 twice
```

Is this also intended behavior?

kortenkamp:
Closed through #968

PR #966, opened 2025-07-22
Evaluare Modifiers in the call-site context

benjaminSoelch:
Currently Modifiers are evaluated in the scope of the called function, due to this evaluation order using a global variable with the same name as a function parameter can lead to unexpected behavior (see #965 ).

This pull request changes the evaluation order to the more intuitive behavior that modifiers are evaluated at the call-site of the function.


kortenkamp:
Thank you for identifying this issue. May I request that you add some test file as well (just for the future)? It would be of interest whether this bug is present in CindyClassic, too (I don't think so), and it might be the cause of some strange bugs that we  found when working on the VAM framework.

benjaminSoelch:
Continued in pull request #967

PR #967, opened 2025-07-24
Add test-case for modifier error

benjaminSoelch:
Finish pull request #966, add test-case for new behavior and fix typo in comment

### DIFF
--- a/examples/bugs/modifierContext.html
+++ b/examples/bugs/modifierContext.html
@@ -1,0 +1,67 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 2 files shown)


kortenkamp (approved):

benjaminSoelch:
I checked the behavior in Cindy classic: Variable names do not influence how modifiers are evaluated but previously declared modifiers do.

The following code prints `[2,4]` in Cindy classic and `[2,2]` in CindyJS
```
f():=(
  print((x,y));
);
x=1;
y=1;
f(x->2*y,y->2*x);
```

Which behavior is more desirable?

PR #969, opened 2025-07-25
Port support for "lambda-functions" from CindyGL3D to main interpreter

benjaminSoelch:
As far as I can tell CindyScript currently provides no convenient way treat functions as values that can be passed to other functions/stored in variables. (correct me if I am wrong).
  The closest possible solution (used by Cindy in Cindy) seems to be, that code can be stored in dictionary entries and will be evaluated every time that entry is read, but this approach still depends on global variables to pass arguments to that "function".
Being able to treat functions as values is useful in larger projects, for instance for defining callback functions in UI application, or for implementing a simplified version of polymorphism.

In CindyGL3D I use "cglLazy"-values (created wrapping the passed in expression in a CindyScript object at the plugin level using the `cglLazy()` function) that can be used to pass CindyScript expressions through multiple layers of function calls to the CindyGL compiler (or be evaluated as code using `cglEval()`).

This pull-request adds a less hacky version of that feature to the main CindyScript interpreter.
To archive this functionality I would add three new functions to the interpreter (two of them variadic):

    * `lambda(<vars...>,<expr>)`: create a value representing a function with `expr` (an expression) as body and `vars` (zero or more variables) as parameters. Modifiers will be evaluated on creation and stored with the function.
    * `eval(<lambda>,<args..>)`: evaluate the lambda value `lambda` with the given arguments `args`, if both lambda and eval define the same modifier the modifier given to lambda will be used.
    * `islambda(<val>)` check if a value is lambda expression

Currently there is no function ending with `lambda` in the GitRepository, while `eval` already is already a built-in function, so code-collisions should be minimal. I am open to changes to the function names and signatures, as well tweaks the behavior of modifiers.

Optionally it could be usefull to allow to refer to "normal" functions as objects using an expression of the form `<func-name>$<arity>`  (e.g. `sin$1` to get a lambda object for the sin function). This might require changes to how CindyScript code is parsed

  Example Code:
  ```
  plus = lambda(x,y,x+y);
  times = lambda(x,y,x*y);
  reduce(aList,op):=(
    if(len(aList)==0,
      nada
    ,
      res = aList_1;
      forall(2..len(aList),i,
        res = eval(op,res,aList_i);
      )
    );
  );

  reduce([1,2,3,4,5,6,7,8,9,10],plus); // 55
  reduce([1,2,3,4,5,6,7,8,9,10],times); // 3628800
  ```

  ```
  movePoint(name,newPos):=(
    p = points_name;
    p_"pos" = newPos;
    if(!isundefined(p_"onMove"),
      eval(p_"onMove",newPos)
    );
  )

  newPoint((0,0),onMove->lambda(pos,print("moved point to "+text(pos))));
  newPoint((0,1),onMove->lambda(pos,print("|point2|="+text(pos*pos))));
  ```

### DIFF
--- a/src/js/libcs/Essentials.js
+++ b/src/js/libcs/Essentials.js
@@ -172,6 +172,9 @@ function niceprint(a, modifs) {
     if (a.ctype === "image") {
         return "IMAGE";
     }
+    if (a.ctype === "lambda") {
+        return "lambda(" + a.params.map((v) => v.name).join(",") + ",...)";
+    }
 
     return "_?_";
 }
@@ -182,16 +185,11 @@ niceprint.errorTypes = ["_?_", "_??_", "_???_", "___"];
 //this is the container for self-defined functions
 //Distinct form evaluator for code clearness :-)
 //*******************************************************
-function evalmyfunctions(name, args, modifs) {
-    const tt = myfunctions[name];
-    if (tt === undefined) {
-        return nada;
-    }
-
+function evalfunction(params, body, args, modifs) {
     const set = [];
     let i;
 
-    for (i = 0; i < tt.arglist.length; i++) {
+    for (i = 0; i < params.length; i++) {
         set[i] = evaluate(args[i]);
     }
     //  evaluate modifiers in caller-scope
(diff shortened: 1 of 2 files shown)


benjaminSoelch:
closed for now until feature is finalized

PR #971, opened 2025-07-31
proof of concept for early return

benjaminSoelch:
**This is currently only a proof of concept**

It is possible to implement early return without significant changes to the interpreter using exception based control flow:
* `return(...)` is a function that takes the return value as argument, internally this function throws a `CindyScriptReturn` exception
* When evaluating a function body `CindyScriptReturn` exceptions can be caught to obtain the return value, for convenience this pull request adds a `evaluateR` function for this purpose
* The top-level evaluation function for each script should use `evaluateR` to prevent the exception from leaking out of the interpreter.
The same approach can also be using to add the ability to break out of loops early.

Currently unanswered questions:
* Does a return statement fit into CindyScript
* Should `return` be a function or a keyword/operator
* Should return be caught by `eval`
* Should return in userData expression be caught
* How should a return outside a function be handled
* Is using exceptions for control-flow a good idea
* If yes how can we reliably prevent the exception from leaking out of the interpreter

### DIFF
--- a/examples/earlyReturn.html
+++ b/examples/earlyReturn.html
@@ -1,0 +1,94 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    
+    <title>Modifier Evaluation Context</title>
+    <style type="text/css">
+        * {
+            margin: 0px;
+            padding: 0px;
+        }
+        
+        #CSConsole {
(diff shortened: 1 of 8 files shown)


benjaminSoelch:
This a proof of concept not a finished feature, so I will close it for now

PR #972, opened 2025-07-31
Add variadic print functions

benjaminSoelch:
When debugging it is often useful to print multiples values at once, which leads to slightly confusing error messages of the form `print$2` does not exist.
While printing multiple values is possible by putting all elements in an array having a single variadic print function is more convenient.

Variadic printing would also make concatenating multiple values slightly easier:
`print(text(x)+" + "+text(y)+" = "+text(x+y))` -> `print(x,"+",y,"=",x+y)`

### DIFF
--- a/src/js/libcs/Operators.js
+++ b/src/js/libcs/Operators.js
@@ -101,6 +101,17 @@ evaluator.println$1 = function (args, modifs) {
     csconsole.out(niceprint(evaluate(args[0], modifs)));
     return nada;
 };
+// variadic print functions
+evaluator.print = function (args, modifs) {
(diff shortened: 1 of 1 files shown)


BernhardWerner:
First of all, the way I would write this example in the current CindyJS version is 
```print(x + " + " + y + " = " + (x+y))```
which is about the same complexity and readability of the variadic version? To make debugging messages like these easier to create, we might want to think about string formatting? Especially since these kind of strings are also often used in `drawtext` and not just `println`.

Apart from that, a variadic print would be very much welcomed by me!

PR #991, opened 2025-10-24
Fix printing of cyclic JSONs

benjaminSoelch:
The previous code for cycle detection did not work if nested JSONs are stored in lists


benjaminSoelch:
closed for now.

Issue #995, opened 2026-05-29
`make tests` crashes


benjaminSoelch:
When running `make tests` I get the following error message:

```
[nodetest            ] ~/CindyJS/CindyJS/ref/js/runtests.js:5
[nodetest            ] global.navigator = {};
[nodetest            ]                  ^
[nodetest            ] 
[nodetest            ] TypeError: Cannot set property navigator of #<Object> which has only a getter
[nodetest            ]     at Object.<anonymous> (~/CindyJS/CindyJS/ref/js/runtests.js:5:18)
[nodetest            ]     at Module._compile (node:internal/modules/cjs/loader:1734:14)
[nodetest            ]     at Object..js (node:internal/modules/cjs/loader:1899:10)
[nodetest            ]     at Module.load (node:internal/modules/cjs/loader:1469:32)
[nodetest            ]     at Function._load (node:internal/modules/cjs/loader:1286:12)
[nodetest            ]     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
[nodetest            ]     at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)
[nodetest            ]     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:151:5)
[nodetest            ]     at node:internal/main/run_main_module:33:47
[nodetest            ] 
[nodetest            ] Node.js v23.11.0
node ref/js/runtests.js exited with code 1
make: *** [tests] Error 1
```

As far as I can tell the problem is that in newer versions of Node.js (apparently since v21: https://github.com/nodejs/node/issues/39540#issuecomment-2614062593) `global.navigator` is a predefined read-only object.

kortenkamp:
Well… that is the annoying world of node.js. Here I am still running older versions, and with the newer ones several things break.   I recommend (for the moment) to use `nvm` to use v20 or older, or, if you like, fix the run test bug. I have no time for that right now, unfortunately.

benjaminSoelch:
Should be fixed by #997

PR #996, opened 2026-05-29
Minor Features

benjaminSoelch:
changes:
* add `isBool` and `isJSON` checks
* add `sort` modifier to `keys()` function
* add `removeAt(list,index)` function
* add `merge(aJSON,bJSON)` function

------

Do we need test-cases for the new features?

The CindyScript test suite in the tests folder does not seem to work on my device (see #995).


kortenkamp:
Tests are always good… so please add them later when you work around #995.

PR #998, opened 2026-05-29
Custom modifiers

BernhardWerner:
This allows users to define modifiers for their own functions via 
```
foo(x, customColor -> (1,0,0)) := drawtext([0, 0], x, color -> customColor);
```
It is then called as 
```
foo("text", customColor -> (0,1,0));
```


kortenkamp:
I will make this work in CindyClassic, too… until then please wait just a little while.

kortenkamp:
Implemented in the latest CindyClassic version (2132), so I will merge now.

PR #999, opened 2026-06-01
Support `regional` in loops

benjaminSoelch:
Run the expressions passed to built-in  loop / control-flow  functions (`repeat` `forall` `select` `apply` `while` `if` `sum` `product` `min` `max` `sort` `d` `tangent`) in a separate scope, allowing the use of regional variables within loop-bodies

+ add test-cases for general scoping behavior in CindyScript

----

* should `d` and `tangent` be included in this change, they fit the pattern of functions taking an expression as argument, but are not ready control-flow statements

### DIFF
--- a/tests/Scope_tests.js
+++ b/tests/Scope_tests.js
@@ -42,9 +42,23 @@ describe("Scope: global vs regional", function () {
 describe("Scope: regional in loop", function () {
     itCmd("x=0;forall(1..10,regional(t);t=#*#;x=x+t);x", "385");
     itCmd("x=0;forall(1..10,regional(t);t=#*#;x=x+t);t", "___");
+    itCmd("x=0;repeat(10,regional(t);t=#*#;x=x+t);x", "385");
+    itCmd("x=0;repeat(10,regional(t);t=#*#;x=x+t);t", "___");
     itCmd("apply(1..5,regional(t);t=#*#;t)", "[1, 4, 9, 16, 25]");
     itCmd("apply(1..5,regional(t);t=#*#;t);t", "___");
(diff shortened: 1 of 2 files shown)


kortenkamp:
Looks fine to me. I am currently implementing the same behavior in CindyClassic, as soon as this works I’ll merge.

kortenkamp (approved):
Looks fine to me.

tests/Scope_tests.js
```diff
@@ -0,0 +1,50 @@
...
+
+function itCmd(command, expected) {
+    it(command, function () {
+        String(cdy.niceprint(cdy.evalcs(command))).should.equal(expected);
+    });
+}
+
+describe("Scope: argument", function () {
+    itCmd("x=nada;x=0;f(x):=(x);f(3)", "3");
+    itCmd("x=nada;x=0;f(x):=(x);f(3,x->12)", "12"); // TODO do we want modifiers to overwrite arguments?
```
kortenkamp: No. We do not want this, as it is semantically unclear. This test should be removed.

kortenkamp: I think it should be treated as undefined behaviour for now… I know that CindyClassic outputs 3, as I just ran into that problem when I copied your tests for CindyJS to CindyClassic.  But really, I think it is semantically unclear, so let us treat as as undefined.

kortenkamp:
This feature will be available in CindyClassic with builds >2031.

kortenkamp:
Can you also add tests for `sum``product``min``max``sort` ? 

As for `d`and `tangent` – yes, I think they should be included, too, for consistency.

PR #1000, opened 2026-06-04
Optimize List matrix operations

strobelm:
So, in around 2015 I promised @richter-gebert to clean up the linear algebra routines I created. Took a bit longer to fulfill the promise. 

## What changed

- CSNumber: `zero`/`one`/`infinity`/`nan`/`z3a`/`z3b`/`cub*` are pre-allocated constants instead of allocating getters. 
- List: inline the multiply-add in the products; dispatch `mult` on the first element instead of full type scans (ragged/mixed/geo inputs still return `nada`); replace JSON cloning with a structural `copyMatrix` that shares the immutable leaves, and drop two copies that weren't needed (`QRdecomp`, `inverseIteration`); `isUpperTriangular` checks the sub-diagonal directly.
- tests: stub `navigator` once via `.mocharc.cjs`/`tests/setup.cjs` only when absent (fixes Node 21+ read-only global).
- Add `benchmarks/list_matrix.js` + `make benchmark`.

`copyMatrix` also preserves `Infinity`/`NaN`, fixing a latent bug where the JSON clone turned them into `null`.

## Benchmarks (best of 5, isolated processes, vs main)

| operation | main | branch | speedup |
| --- | --- | --- | --- |
| scalproduct vec3 | 72.0 ns | 6.7 ns | 10.8x |
| mult MM 3x3 | 1.08 µs | 250.7 ns | 4.3x |
| inverse 4x4 | 38.3 µs | 2.93 µs | 13.0x |
| LUsolve 4x4 | 18.8 µs | 1.56 µs | 12.0x |
| QRdecomp 4x4 | 74.7 µs | 14.0 µs | 5.3x |
| linearsolve 100x100 cplx (88_LA_Tests) | 20.2 ms | 14.1 ms | 1.4x |
| inverse 100x100 cplx (88_LA_Tests) | 66.5 ms | 48.2 ms | 1.4x |

Largest gains on small/low-dimension matrices where the removed per-call overhead dominates the O(n^3) work.

Pls have a thorough look on the changes, since they touch the core of CindyJS.


kortenkamp:
Cool, thank you. May I ask @richter-gebert or @BernhardWerner to give this a look? I am quite sure that everything is fine, but as linear algebra is kind of crucial for Cindy, a second and third opinion would be great.

PR #1002, opened 2026-06-18
Report line numbers of errors

benjaminSoelch:
Preserves the token locations from the parser into the interpreter allowing to print the line where a specific error occurred.
This makes it easier to determine which line in a large CindyScript file resulted in the warning message being printed.


benjaminSoelch:
The CI failure is due to a change from 0-indexed column numbers to 1-indexed column numbers.

Most Text-editors use 1-indexed columns so it makes sense to report the locations using that numbering scheme. There should not be any code depending on the actual column-numbers used in error reporting, so this should not break anything.

PR #1003, opened 2026-06-18
Add CSV-stype String escaping

benjaminSoelch:
Currently there is no convenient way to create strings containing the character `"`.

This pull request adds minimal string-escaping with similar semantics as the CSV-format, treating `""` within a string as the character `"` instead of ending and restarting a string literal.

Currently strings containing `""` are treated as two consecutive string-literals which is a syntax-error, so this new syntax does not invalidate any existing programs.


kortenkamp:
This is also available in Classic starting with build 2135.

